PR #13, opened 2023-10-11
JetBrains: Cody: Fixing applying autocompletions in html file and for latest EAP

### LINKED ISSUE #57374: JetBrains: autocomplete does not work in latest IntelliJ EAP
To reproduce: install 233.8264.8, trigger autocomplete, press `TAB`, nothing happens. 

I'm unable to reproduce with 233.6745.305. We should either fix this on our side or report an upstream bug.

### LINKED ISSUE #57454: JetBrains: autocomplete does not always work in HTML files
To reproduce:

- Open empty HTML file
- Type pageTitle
- Press alt-backslash to trigger autocomplete
- Verify there is an autocomplete suggestion
- Press tab to accept autocomplete
- Instead of accepting autocomplete, a snippet gets expanded to `<pageTitle></pageTitle>`

Sa1to:
Fixes https://github.com/sourcegraph/sourcegraph/issues/57454 

Fixes https://github.com/sourcegraph/sourcegraph/issues/57374

## Test plan
* Try to accept autocompletions in latest Intellij EAP (233.9102.97)
* Open empty HTML file
* Type any text
* Press alt-backslash to trigger autocomplete
* Verify there is an autocomplete suggestion
* Press tab to accept autocomplete
* Autocompletion is applied

https://github.com/sourcegraph/jetbrains/assets/9321940/459442ef-a72e-4181-a51a-d7197e2fd9ad

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/action/CodyActionPromoter.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/action/CodyActionPromoter.kt
@@ -1,0 +1,29 @@
+package com.sourcegraph.cody.autocomplete.action
+
+import com.intellij.openapi.actionSystem.ActionPromoter
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.editor.actionSystem.EditorAction
+
+class CodyActionPromoter : ActionPromoter {
+  override fun promote(
+      actions: MutableList<out AnAction>,
+      context: DataContext
+  ): MutableList<AnAction>? {
+    return if (actions.stream().noneMatch { action: AnAction? ->
+      action is CodyAction && action is EditorAction
+    }) {
+      null
+    } else {
+      val result: ArrayList<AnAction> = ArrayList(actions)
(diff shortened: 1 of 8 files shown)


src/main/kotlin/com/sourcegraph/cody/autocomplete/action/CodyActionPromoter.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.editor.actionSystem.EditorAction
+
+class CodyActionPromoter : ActionPromoter {
+  override fun promote(
+      actions: MutableList<out AnAction>,
+      context: DataContext
+  ): MutableList<AnAction>? {
+    return if (actions.stream().noneMatch { action: AnAction? ->
```
olafurpg, line 13: nit: it would be nice to extract a variable out here
```suggestion
    val isCodyAction = actions.stream().noneMatch { action: AnAction? -> ...
    return if (isCodyAction) { null } else {
```

olafurpg (approved):
LGTM 👍🏻 Nice job spotting the problem!

PR #18, opened 2023-10-12
Setup automated CI releases

olafurpg:
Previously, I have been releasing the plugin from my local computer using a manual process. This was problematic because it was easy to screw up the release by having a wrong version of Cody cloned locally, or forgetting to publish an accompanying alpha release for every stable release. The 3.4.0 had a critical regression that was caused by this exact root problem.

This PR is the first step towards addressing the problem by adding CI workflows to publish releases. By publishing the releases in CI, we document all the necessary steps to publish a release, and we enable anyone on the team to cut a release at any time.

By default, we only publish alpha releases for now. I want to follow up with a separate PR that automates the process to "promote" an alpha release to stable. My goal is that we use a byte-by-byte identical release to the alpha when promote to stable.

## Test plan

Difficult to test this without running the actual workflows in CI.


src/test/kotlin/com/sourcegraph/cody/config/SourcegraphServerPathTest.kt
```diff
@@ -14,6 +14,8 @@ class SourcegraphServerPathTest {
     // when
     val serverPath = SourcegraphServerPath.from(url, "")
 
+    println()
```
mkondratek: why is it needed?

scripts/verify-release.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -eux
+
+./gradlew clean buildCodeSearchAssets buildPluginAndAssertAgentBinariesExist runPluginVerifier
```
mkondratek, line 8: This script is not used by CI, is it? It is only for us to execute it, right?

.github/workflows/release.yml
```diff
@@ -0,0 +1,16 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: 11
+          cache: 'gradle'
+      - run: echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)-alpha" >> $GITHUB_ENV
```
mkondratek: I see what `scripts/version-from-git-tag.sh` does.
What about the tags? The still need to be pushed manually (via `scripts/push-git-tag-for-next-release.sh`, not by the CI), right?

scripts/next-release.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if ! command -v gdate &>/dev/null; then
+  echo "command not found: gdate"
+  echo "The command gdate is required to compute the next version number"
+  echo "To fix this problem, run:\n  brew install coreutils"
+  exit 1
+fi
+
+
+MAJOR_MINOR=$(gh release list --repo sourcegraph/sourcegraph --limit 1 --exclude-drafts --exclude-pre-releases | awk '{ print $2 }' | tail -n 1 | cut -d. -f1 -f2)
```
mkondratek, line 18: `sourcegraph/sourcegraph`? or `sourcegraph/jetbrains`?

scripts/next-release.sh
```diff
@@ -0,0 +1,24 @@
...
+LAST_RELEASE_TIMESTAMP=$(gh release list --repo sourcegraph/sourcegraph --limit 1 --exclude-drafts --exclude-pre-releases | awk '{ print $5 }')
```
mkondratek, line 19: the same question

CHANGELOG.md
```diff
@@ -1,757 +1,15 @@
 # Sourcegraph Changelog
 
 ## [Unreleased]
-
```
mkondratek, line 4: don't we need the changelog anymore? 🤔

src/test/kotlin/com/sourcegraph/cody/config/SourcegraphServerPathTest.kt
olafurpg: Good catch, removed. Was leftover

scripts/verify-release.sh
olafurpg: This script will run before pushing a tag. CI will only build and publish artifacts. Running `runPluginVerifier` is too slow to run in CI because it requires downloading multiple IDEA versions.

.github/workflows/release.yml
olafurpg: Correct. You have to manually push the git tag, which requires running validation tests locally first, and then CI runs the build+publish.

olafurpg: I will update contributing.md to reflect the new flow once I have validated everything works.

CHANGELOG.md
olafurpg: Nope, we will use GitHub releases instead to document releases.

scripts/next-release.sh
olafurpg: This is correct, we use the same major.minor version as the latest stable Sourcegraph release. The patch version is a time-based formula based on the timestamp of the latest Sourcegraph release.

.github/workflows/release.yml
olafurpg: Updated contributing.md in the latest commit

CHANGELOG.md
mkondratek: ok

Personally, I would prefer to have a changelog in the scope of the repo code.
I guess it is possible to automatically extract the most recent changes with some bash code and create the release with some script (https://cli.github.com/manual/gh_release_create).
I do not insist.

build.gradle.kts
```diff
@@ -77,6 +81,118 @@ java {
 }
 
 tasks {
+  val codeSearchCommit = "9d86a4f7d183e980acfe5d6b6468f06aaa0d8acf"
```
mkondratek, line 84: what is code search? will we need to update this commit sometime?

mkondratek (approved):
lgtm

build.gradle.kts
olafurpg: We have two products: Cody and Code Search. Code Search is implemented in the sourcegraph/sourcegraph repository, and this is a commit in that repo. This commit hash will need to be updated when we update code search features (alt-s).

PR #26, opened 2023-10-13
Fix build for release automation

olafurpg:
I failed to cut a release with my previous attempt. This PR changes the build config to stop using Gradle tasks and uses plain Kotlin functions instead. I was spending a lot of time trying to debug why the tasks were not running, and realized we don't get any benefits from using Gradle tasks anyways (like automatic caching).

## Test plan
Green CI.


CONTRIBUTING.md
```diff
@@ -33,16 +33,16 @@ using the `jetbrains-ide` & `team/integrations` labels.
 | Reformat Java and Kotlin sources                              | `./gradlew spotlessApply`                                                |
 | Debug agent JSON-RPC communication                            | `tail -f build/sourcegraph/cody-agent-trace.json`                        |
 
-## Using Alpha channel releases
+## Using Nightly channel releases
 
-We occasionally publish plugins to the "Alpha" channel instead of the default
-"Stable" channel. The alpha channel is primarily intended to publish
+We occasionally publish plugins to the "Nightly" channel instead of the default
```
mkondratek, line 38: And it will be possible to change it from the `Sourcegraph & Cody` settings too 😄

CONTRIBUTING.md
```diff
@@ -80,9 +80,9 @@ CI. When you have a local cache of IDEA installations then this script can run d
 
 ## Publishing a new stable version
 
-This workflow is not automated yet. Our goal is to publish stable releases by promoting alpha releases that have been
-manually tested. What we're missing is a script to download the zip file of an alpha release, correct the version
-to be non-alpha and then upload it to the JetBrains Marketplace.
+This workflow is not automated yet. Our goal is to publish stable releases by promoting nightly releases that have been
+manually tested. What we're missing is a script to download the zip file of an nightly release, correct the version
```
mkondratek: nit: **a** nightly release 🕵️‍♂️

scripts/verify-release.sh
```diff
@@ -5,4 +5,6 @@ echo "====================================================="
 echo "= Running automated tests before publishing release ="
 echo "====================================================="
 set -x
-./gradlew clean buildCodeSearchAssets buildPluginAndAssertAgentBinariesExist runPluginVerifier
+unset CODY_DIR
+./gradlew clean
```
mkondratek, line 9: was there any trouble without it?

olafurpg: Gradle was failing when I included `clean` with the other tasks. I don't really understand Gradle, it seems like a hot mess compared to other build tools I've used.

CONTRIBUTING.md
olafurpg: Fixed, thanks!

src/main/kotlin/com/sourcegraph/cody/config/ui/UpdateChannel.kt
```diff
@@ -4,12 +4,12 @@ import com.intellij.util.ui.PresentableEnum
 
 enum class UpdateChannel(val channelUrl: String?) : PresentableEnum {
   Stable(null as String?),
-  Alpha("https://plugins.jetbrains.com/plugins/alpha/9682");
+  Nightly("https://plugins.jetbrains.com/plugins/alpha/9682");
```
mkondratek: I think the link can be update now as well 🆙

olafurpg: Done!

olafurpg:
OK, I think this is ready to go. CI is taking 8 minutes though, which is not great so I want to fix that in a separate PR

CONTRIBUTING.md
```diff
@@ -33,16 +33,16 @@ using the `jetbrains-ide` & `team/integrations` labels.
...
-"Stable" channel. The alpha channel is primarily intended to publish
+We occasionally publish plugins to the "Nightly" channel instead of the default
+"Stable" channel. The nightly channel is primarily intended to publish
 pre-releases for internal (within Sourcegraph) testing.
 
 - Open Settings
 - Open "Plugins"
 - Click on cogwheel in the top bar, select "Manage plugin repositories"
-- Add the URL https://plugins.jetbrains.com/plugins/list?channel=alpha&pluginId=9682
+- Add the URL https://plugins.jetbrains.com/plugins/list?channel=nightly&pluginId=9682
```
mkondratek, line 45: nit: We can use:
https://plugins.jetbrains.com/plugins/nightly/9682
here too!

mkondratek (approved):
lgtm

CONTRIBUTING.md
olafurpg: I'll leave this as is because I want to merge the PR. This documentation can be replaced with "Enable nightly channel in settings" once it's live

PR #28, opened 2023-10-13
Persist codebase URL after closing IDE (#6)

### LINKED ISSUE #56948: JetBrains: persist project codebase setting
Currently, `CodyAgentCodebase.{inferredCodebase,explicitCodebase}` are normal string fields that needs to be recomputed every time the IDE restarts. We should persist these values so that they have stable values even after a restart.

https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@63055924b5dff0e458063c5778e3ddd705eaf766/-/blob/client/jetbrains/src/main/java/com/sourcegraph/cody/agent/CodyAgentCodebase.kt?L14-15

exigow:
Persist codebase URL after closing IDE (#6)

Fixes https://github.com/sourcegraph/sourcegraph/issues/56948

Remote URL is now stored in project settings.

Additional note: I've removed the line stating that leaving the address
empty would result in an inferred URL from the Git repository. This statement
is not true: the dialog does not allow you to leave it blank.

## Test plan

* `./gradlew :runIde` and open project.
* Open Context Selection dialog. Change URL and click OK.
* Restart IDE.
* Make sure repository address remains the same as it was before the restart.
* Make sure status bar is properly displayed after 2nd time you open the project.
* Make sure configuration change isn't send on every cursor movement (this should be visible in logs).
* Make sure status bar is properly displayed after opening large project (when indexing takes time).

### DIFF
--- a/src/main/java/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
+++ b/src/main/java/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
@@ -4,32 +4,33 @@ import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.sourcegraph.cody.CodyToolWindowContent
+import com.sourcegraph.cody.config.CodyProjectSettings
 import com.sourcegraph.config.ConfigUtil
 import com.sourcegraph.vcs.RepoUtil
 
-class CodyAgentCodebase(private val underlying: CodyAgentServer, val project: Project) {
+class CodyAgentCodebase(private val underlying: CodyAgentServer, private val project: Project) {
 
-  // TODO: This should ideally be persisted as a project-wide setting. Down the road, it should also
-  // support a list of repository names instead of only a single codebase.
-  private var inferredCodebase: String = ""
-  private var explicitCodebase: String = ""
+  // TODO: Support list of repository names instead of just one.
+  private val application = ApplicationManager.getApplication()
+  private val settings = CodyProjectSettings.getInstance(project)
+  private var inferredUrl: String? = null
 
   init {
-    ApplicationManager.getApplication().executeOnPooledThread {
-      onRepositoryName(RepoUtil.findRepositoryName(project, null))
+    application.executeOnPooledThread {
+      onRepositoryNameChange(RepoUtil.findRepositoryName(project, null))
     }
   }
 
-  fun onNewExplicitCodebase(codebase: String) {
-    explicitCodebase = codebase
+  fun setUrl(url: String) {
+    settings.remoteUrl = url
     onPropagateConfiguration()
   }
 
-  fun currentCodebase(): String? = explicitCodebase.ifEmpty { inferredCodebase }.ifEmpty { null }
+  fun getUrl(): String? = settings.remoteUrl ?: inferredUrl
 
   fun onFileOpened(project: Project, file: VirtualFile) {
-    ApplicationManager.getApplication().executeOnPooledThread {
-      onRepositoryName(RepoUtil.findRepositoryName(project, file))
+    application.executeOnPooledThread {
+      onRepositoryNameChange(RepoUtil.findRepositoryName(project, file))
     }
   }
 
@@ -38,10 +39,10 @@ class CodyAgentCodebase(private val underlying: CodyAgentServer, val project: Pr
     underlying.configurationDidChange(ConfigUtil.getAgentConfiguration(project))
   }
 
-  private fun onRepositoryName(repositoryName: String?) {
-    ApplicationManager.getApplication().invokeLater {
-      if (repositoryName != null && inferredCodebase != repositoryName) {
-        inferredCodebase = repositoryName
+  private fun onRepositoryNameChange(repositoryName: String?) {
+    application.invokeLater {
+      if (repositoryName != null && inferredUrl != repositoryName) {
+        inferredUrl = repositoryName
         onPropagateConfiguration()
       }
     }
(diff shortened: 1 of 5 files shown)


chwarwick (dismissed):
This looks like the same fix I tried the other day.  The problem with it is that it doesn't work the 2nd time you open the project because the codebase is already present in the settings, the `onPropagateConfiguration` doesn't run to update the chat window so it always shows as no repo.

exigow:
@chwarwick, thanks for explaination. I'm going to fix that.

exigow:
Description updated.

exigow:
I've noticed a regression: the status bar still has issues with refreshing during indexing (when importing large projects first time). Marking as draft until I solve this.

exigow:
After extensive testing, I believe all the issues have been fixed. I've slightly modified the concept when `remoteUrl` value is stored.

To sum up:
* The URL is persistent between sessions.
* The status bar reflects reality as-is.
* The remote URL is added to the settings file only when set explicitly. Settings do not contain the URL when the address is inferred from the Git repo, and the user is okay with that.

Test plan takes all issues into account.

chwarwick (commented):
This looks good, the only potential issue I see is that the issue if you open a folder with multiple repos in it, the codebase doesn't adjust as you switch between them.
Screenshot 2023-10-11 at 10 45 21 AM

This is different than the way the VS Code extension behaves, it does dynamic detection at every active file change and it's only persisted when an explicit override is made.  I don't have a sense of how common it is to open multiple repos at once.  In general I think this is a better solution because the settings lives with the code instead of the IDE so it's likely a good trade off.

exigow:
Wow, thanks for pointing that out! I was not aware of that. I will certainly test this multi-repo case and provide a solution.

chwarwick:
> Wow, thanks for pointing that out! I was not aware of that. I will certainly test this multi-repo case and provide a solution.

I don't know if the multi repo case should be a blocker for this, because I don't think it's common, but it is something to be aware of.

exigow:
To be sure I've tested the behavior for multiple repos in a single IDEA project, and it works in such a way that there's a single codebase. Switching between files always points to "global" URL specified in the settings.

By the way, there's a TODO in the code that mentions adding support for multiple repositories in the Context Selection window:

https://github.com/sourcegraph/jetbrains/blob/3d22b013cd64094b314845657ba71e00a2970721/src/main/java/com/sourcegraph/cody/agent/CodyAgentCodebase.kt#L13

To sum up:

1. I'm re-designing Context Selection UI to allow adding multiple repositories.
2. ...OR we merge this persistent URL feature, but we introduce this minor issue with multiple repositories. We're adding support in UI later.

@olafurpg, what do you think?

src/main/java/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
```diff
@@ -39,9 +40,9 @@ class CodyAgentCodebase(private val underlying: CodyAgentServer, val project: Pr
   }
 
   private fun onRepositoryName(repositoryName: String?) {
```
mkondratek: nit: can we rename it to `onRepositoryNameChange`

src/main/kotlin/com/sourcegraph/cody/auth/ui/EditCodebaseContextAction.kt
```diff
@@ -105,10 +104,6 @@ class EditCodebaseContextAction(val project: Project) : AbstractAction("Cody Con
             }
             .rowComment("Example: github.com/sourcegraph/cody")
         row { text("The URL will be used to retrieve the code context from the Cody API.") }
-        row {
-          text(
-              "If left empty, Cody will automatically infer the URL from your project's Git metadata.")
```
mkondratek, line 110: I do not understand 🤔 isn't it true?

mkondratek (approved):
minor comment, lgtm

src/main/kotlin/com/sourcegraph/cody/auth/ui/EditCodebaseContextAction.kt
exigow: Correct. I you open the Context Selection dialog you can't leave it empty. I've explained that in the commit message/PR description.

This statement is also not valid on main branch.

src/main/java/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
exigow: fixed

olafurpg:
@exigow For now, it's OK to only support a single repository in the UI. On the web, Cody supports multiple repositories (https://sourcegraph.com/cody/) and I anticipate we will add the same support in the JetBrains plugin in the next couple months. If we are persisting values then it might make sense to persist a list of repositories so that we avoid doing a data migration when we add this support.

olafurpg:
If it's not a problem to do a data migration in the future then it's totally fine to just handle one repo for now.

src/main/kotlin/com/sourcegraph/cody/auth/ui/EditCodebaseContextAction.kt
mkondratek: ack! thanks for the explanation

chwarwick:
> @exigow For now, it's OK to only support a single repository in the UI. On the web, Cody supports multiple repositories (https://sourcegraph.com/cody/) and I anticipate we will add the same support in the JetBrains plugin in the next couple months. If we are persisting values then it might make sense to persist a list of repositories so that we avoid doing a data migration when we add this support.

@olafurpg  I wasn't referring to multi repo context, but either opening a folder that contained multiple repos or a repo with submodules.  Previously the repo context would switch when you opened a file in the submodule or different repo, and then switch back when the active tab went back to a file from the other repo.  I think this is ok for now as those types of setup are less common and ultimately multi repo context would likely be a better solution than flipping the context back and forth.

olafurpg:
@chwarwick right. I don't have a strong opinion about that case. I think this kind of multi-repo setup is kind of common, and I kind of like the behavior where we updated the active repo based on your open editor tab. When we support multiple repo context then it won't matter which behavior we go with for now, so I leave it to you @exigow to decide

exigow:
@olafurpg, I decide to merge these changes then. 

Addressing your concerns: It's important to note that if the user doesn't specify their own "global" address in Context Selection, the behavior will remain the same as it was before adding this feature. If the repository is not explicitly defined, the codebase will be inferred from currently opened file.

Thanks @chwarwick for your valuable feedback on this.

PR #33, opened 2023-10-16
Support macos-arm64 with CI releases

olafurpg:
Previously, we ran the x86 binary on Apple M1 computers because the arm64 binary required code signing to run correctly on macOS computers. This PR fixes the problem by using `ldid2` to sign the macos-arms64 binary.

## Test plan

Run a nightly build and confirm it works as expected on macOS.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,13 +22,13 @@ jobs:
       - run: |
           echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)" >> $GITHUB_ENV
       - run: echo "Publishing version $RELEASE_VERSION"
-      - run: ./gradlew "-PpluginVersion=$RELEASE_VERSION" publishPlugin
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -17,10 +17,18 @@ jobs:
           cache: "gradle"
       - name: Gradle Wrapper Validation
         uses: gradle/wrapper-validation-action@v1.1.0
+      - run: ./scripts/install-ldid.sh
       - run: yarn global add pnpm@8.6.7
       - run: |
-          echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)-nightly" >> $GITHUB_ENV
+          echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)" >> $GITHUB_ENV
```
olafurpg, line 23: These changes are unrelated to apple/arm64, but they are required to publish a stable+nightly release together.

mkondratek (approved):

PR #34, opened 2023-10-16
Fix bug in release automation

olafurpg:
The release job failed because it didn't recognize the new `MAJOR.MINOR.PATCH-nightly` format.

## Test plan
n/a


olafurpg:
Requested admin rights so I can push straight to main while I iterate on release automation.

PR #35, opened 2023-10-16
Stop using Gradle cache for sourcegraph+cody clones

olafurpg:
Previously, we did shallow git clones of sourcegraph/sourcegraph and sourcegraph/cody to the Gradle cache. This was problematic because it was taking a very long time to create the cache in CI, defeating the point of caching. This PR moves the shallow clones to another directory so that they are not cached, which hopefully speeds up the caching step in CI (which is taking 5-7 minutes before this PR)

## Test plan

Merge the PR and see CI become faster

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -163,7 +163,7 @@ fun unzip(input: File, output: File, excludeMatcher: PathMatcher? = null) {
 }
 
 val githubArchiveCache =
-    Paths.get(System.getProperty("user.home"), ".gradle", "caches", "sourcegraph").toFile()
+    Paths.get(System.getProperty("user.home"), ".sourcegraph", "caches", "jetbrains").toFile()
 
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -163,7 +163,7 @@ fun unzip(input: File, output: File, excludeMatcher: PathMatcher? = null) {
 }
 
 val githubArchiveCache =
-    Paths.get(System.getProperty("user.home"), ".gradle", "caches", "sourcegraph").toFile()
+    Paths.get(System.getProperty("user.home"), ".sourcegraph", "caches", "jetbrain").toFile()
```
exigow: Typo `jetbrain`

olafurpg: Good catch, fixed!

mkondratek (approved):

PR #36, opened 2023-10-16
Handle updated access token format from RFC 837

willdollman:
As part of [RFC 837](https://docs.google.com/document/d/1aC4gHB8Q5lurwVhc8SCxznR0blNJMKo7yIkpP7WShno/edit#heading=h.a31kbqeib9od) we will be updating the access token format.

This PR updates the JetBrains plugin's token validator to handle this new format. I've specified 61 or 62 characters as we haven't yet determined whether we can keep the `sgp_` prefix or will have to switch to `sgph_` - this will depend on Github when we join their secret scanning program. We can clean this up in a future release.

Example tokens:
- `sgp_abcdef0123456789_1234567890123456789012345678901234567890`
- `sgph_abcdef0123456789_1234567890123456789012345678901234567890`

Related to https://github.com/sourcegraph/sourcegraph/pull/56772.

## Questions for reviewers

- Is this only called the first time a token is configured/edited, or every time a Cody request is made?
- Do we have a way of determining when all extension users have upgraded to a version that contains this change?

### DIFF
--- a/src/main/java/com/sourcegraph/common/AuthorizationUtil.java
+++ b/src/main/java/com/sourcegraph/common/AuthorizationUtil.java
@@ -7,6 +7,8 @@ public class AuthorizationUtil {
     return accessToken.isEmpty()
         || accessToken.length() == 40
         || (accessToken.startsWith("sgp_") && accessToken.length() == 44)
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
Thank you for this contribution! 🙏🏻

PR #37, opened 2023-10-17
Add a hint on trigger when no account signed in

### LINKED ISSUE #56050: JetBrains: autocomplete - no visual hint when authentication fails
To reproduce:

- Remove all authentication settings
- Trigger autocomplete
- Nothing happens

Expectation:
- Cody icon in the status bar indicates the user is logged out
- Some kind of notification?

mkondratek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/56050.

When the trigger is explicit but there is no account let's provide a visual error hint.

## Test plan

1. Remove all authentication settings
2. Trigger autocomplete
3. a visible hint

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -160,6 +160,7 @@ class CodyAutocompleteManager {
     if (isTriggeredExplicitly &&
         CodyAuthenticationManager.instance.getActiveAccount(project) == null) {
       HintManager.getInstance().showErrorHint(editor, "Cody: No account signed-in")
+      return
     }
     cancelCurrentJob(project)
     val cancellationToken = CancellationToken()
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -159,7 +159,7 @@ class CodyAutocompleteManager {
     }
     if (isTriggeredExplicitly &&
         CodyAuthenticationManager.instance.getActiveAccount(project) == null) {
-      HintManager.getInstance().showErrorHint(editor, "Cody: No account signed-in")
+      HintManager.getInstance().showErrorHint(editor, "Cody: Sign in to use autocomplete")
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -155,6 +157,11 @@ class CodyAutocompleteManager {
     if (isTriggeredImplicitly && !autoCompleteDocumentContext.isCompletionTriggerValid()) {
       return
     }
+    if (isTriggeredExplicitly &&
+        CodyAuthenticationManager.instance.getActiveAccount(project) == null) {
+      HintManager.getInstance().showErrorHint(editor, "Cody: No account signed-in")
```
mkondratek: It looks like this:

image

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -155,6 +157,11 @@ class CodyAutocompleteManager {
...
+      return
```
mkondratek, line 163: I think it's okay to stop here in this case - there is no point in making any requests. Am I right?

olafurpg: Correct.

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
olafurpg: Hmm, I'm not sure what is the best phrase. How about this?
```suggestion
      HintManager.getInstance().showErrorHint(editor, "Cody: sign in to use autocomplete")
```

cc/ @danielmarquespt any ideas? This error is only displayed on the explicit alt-backslash shortcut

olafurpg (approved):
stamp to unblock merging

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
danielmarquespt: "Cody:**S**ign in to use autocomplete"

Just a nitpick with the case, but yes, agreed

PR #42, opened 2023-10-18
Fix the automatic repository detection for ssh remotes with port numbers

### LINKED ISSUE #56929: JetBrains: Automatic repo detection does not work for ssh remotes with port numbers
When a git remote using ssh contains a port number the repo is not automatically detected properly for use with embeddings.

A repo name is generated that looks like `github.com/{PORT_NUMBER}/sourcegraph/sourcegraph`

I would expect this to return  `github.com/sourcegraph/sourcegraph`

VS Code's logic to handle converting the remote to a repo name is [here](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/utils.ts?L8:17&popover=pinned)

mkondratek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/56929. 

Before merging this one we need changes to the agent:
https://github.com/sourcegraph/cody/pull/1425

- [x] once your cody PR is merged, you will need to bump up the commit hash on the jetbrains side

## Test plan
- [x] Test it manually against the reported url:
```
test('converts GitHub SSH URL with the port number', () => {
        expect(convertGitCloneURLToCodebaseName('ssh://git@gitlab-my-company.net:20022/path/repo.git')).toEqual(
            'gitlab-my-company.net/path/repo'
        )
    })
```

- [x] and do this:
> when bumping up the cody commit hash, it's worth doing manual tests for chat/autocomplete/onboarding

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -205,7 +205,7 @@ tasks {
     return destinationDir
   }
 
-  val codyCommit = "68281675fd0731c0b2d30f61b85bb84dea165242" // todo
+  val codyCommit = "8a61ffc8a55631350a31276b39f38e003007a3d2"
   fun downloadCody(): File {
     val url = "https://github.com/sourcegraph/cody/archive/$codyCommit.zip"
     val destination = githubArchiveCache.resolve("$codyCommit.zip")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -205,7 +205,7 @@ tasks {
     return destinationDir
   }
 
-  val codyCommit = "68281675fd0731c0b2d30f61b85bb84dea165242"
+  val codyCommit = "68281675fd0731c0b2d30f61b85bb84dea165242" // todo
   fun downloadCody(): File {
     val url = "https://github.com/sourcegraph/cody/archive/$codyCommit.zip"
     val destination = githubArchiveCache.resolve("$codyCommit.zip")
--- a/src/main/java/com/sourcegraph/vcs/RepoUtil.java
+++ b/src/main/java/com/sourcegraph/vcs/RepoUtil.java
@@ -147,9 +147,21 @@ public class RepoUtil {
     VCSType vcsType = getVcsType(project, file);
 
     if (vcsType == VCSType.GIT && repository != null) {
+      String cloneURL;
+      try {
+        cloneURL = GitUtil.getRemoteRepoUrl((GitRepository) repository, project);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
       return CodyAgent.withServer(
           project,
-          server -> server.convertGitCloneURLToCodebaseName(repository.getPresentableUrl()));
+          server -> {
+            System.out.println("[debug] cloneURL="+ cloneURL);
+            return server.convertGitCloneURLToCodebaseName(cloneURL).thenApply(result -> {
+              System.out.println("[debug] result="+ result);
+              return result;
+            });
+          });
     }
 
     if (vcsType == VCSType.PERFORCE) {
--- a/src/main/java/com/sourcegraph/vcs/RepoUtil.java
+++ b/src/main/java/com/sourcegraph/vcs/RepoUtil.java
@@ -7,6 +7,7 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.vcsUtil.VcsUtil;
 import com.sourcegraph.cody.agent.CodyAgent;
+import com.sourcegraph.cody.agent.protocol.CloneURL;
 import com.sourcegraph.cody.config.CodyProjectSettings;
 import com.sourcegraph.common.ErrorNotification;
 import git4idea.GitVcs;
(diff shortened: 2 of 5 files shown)


src/main/java/com/sourcegraph/website/OpenRevisionAction.java
```diff
@@ -74,36 +74,37 @@ public void actionPerformed(@NotNull AnActionEvent event) {
...
+                                  + ", version "
+                                  + productVersion,
+                              e);
+                          return;
+                        }
+                        BrowserOpener.openInBrowser(project, url);
+                      })
+                  .exceptionally(
+                      e -> {
+                        throw new RuntimeException(e);
```
Sa1to: Isn't it better to throw here more specific `IllegalArgumentException`?

src/main/java/com/sourcegraph/vcs/RepoUtil.java
```diff
@@ -122,39 +123,63 @@ private static String doReplacements(
...
+        cloneURL = GitUtil.getRemoteRepoUrl((GitRepository) repository, project);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+      return CodyAgent.withServer(
+          project,
+          server -> {
+            System.out.println("[debug] cloneURL="+ cloneURL);
+            return server.convertGitCloneURLToCodebaseName(cloneURL).thenApply(result -> {
+              System.out.println("[debug] result="+ result);
```
exigow: We should use logger here.

(Assuming this is not wip-only/temporary code for debugging and this will end up on main branch - otherwise ignore).

exigow (approved):
Looks good.

src/main/java/com/sourcegraph/vcs/RepoUtil.java
mkondratek: I added it only as a temporary "logging" - so, to be removed. I can use a proper logger if we want.

src/main/java/com/sourcegraph/vcs/RepoUtil.java
```diff
@@ -47,7 +49,7 @@ public static RepoInfo getRepoInfo(@NotNull Project project, @NotNull VirtualFil
         relativePath = relativePath.substring(relativePath.indexOf("/") + 1);
       }
 
-      remoteUrl = getRemoteRepoUrl(project, file);
+      remoteUrl = getRemoteRepoUrl(project, file).join(); // todo
```
mkondratek: is it ok to do it like this?

src/main/java/com/sourcegraph/vcs/RepoUtil.java
```diff
@@ -85,11 +87,10 @@ public static String findRepositoryName(
...
     }
-    try {
-      return RepoUtil.getRemoteRepoUrlWithoutScheme(project, fileFromTheRepository);
-    } catch (Exception e) {
-      return RepoUtil.getSimpleRepositoryName(project, fileFromTheRepository);
-    }
+
+    return RepoUtil.getRemoteRepoUrlWithoutScheme(project, fileFromTheRepository)
+        .exceptionally(e -> RepoUtil.getSimpleRepositoryName(project, fileFromTheRepository))
+        .join(); // todo
```
mkondratek: is it ok to do it like this?

src/main/java/com/sourcegraph/vcs/RepoUtil.java
exigow: I don't have practical experience with Java CompletableFutures, but it seems to me that this code will be blocking no matter how long we postpone using `join` in this case.

IMO that's totally fine, but what would be the difference if we used join immediately after sending a request to the server (for example: `server.convertGitCloneURLToCodebaseName(cloneURL).join()`)? It looks to me that the effect would be the same but code would be simpler. I'm not sure if there's room for writing asynchronous code here since this logic heavily rely on the agent's response.

src/main/java/com/sourcegraph/vcs/RepoUtil.java
```diff
@@ -65,7 +67,7 @@ public static RepoInfo getRepoInfo(@NotNull Project project, @NotNull VirtualFil
         message = "Error determining repository info: " + err.getMessage();
       }
       ErrorNotification.show(project, message);
-      Logger.getInstance(RepoUtil.class).warn(message);
```
mkondratek, line 68: unrelated

src/main/java/com/sourcegraph/vcs/RepoUtil.java
mkondratek: Thanks! Very good point. It looks like this code is run on a pooled thread anyway and it is safe to `.join()` immediately. Everything is simple now 😅

build.gradle.kts
```diff
@@ -205,7 +205,7 @@ tasks {
     return destinationDir
   }
 
-  val codyCommit = "68281675fd0731c0b2d30f61b85bb84dea165242"
+  val codyCommit = "68281675fd0731c0b2d30f61b85bb84dea165242" // todo
```
mkondratek: - [x] to be updated ❗

build.gradle.kts
```diff
@@ -205,7 +205,7 @@ tasks {
     return destinationDir
   }
 
-  val codyCommit = "68281675fd0731c0b2d30f61b85bb84dea165242"
```
mkondratek, line 208: ~https://github.com/sourcegraph/cody/commit/8a61ffc8a55631350a31276b39f38e003007a3d2~

mkondratek: https://github.com/sourcegraph/cody/commit/65cbc12c00aed56c4013ef6c97736a82ac0dc294

pkukielka: Out of curiosity why there is an hardcoded commit hash istead of branch or tag name?

PR #44, opened 2023-10-19
Allow to specify cody directory for use instead of the committed version

mkondratek:
I have two run configurations, for Cody from commit:
image

And for the local clone of Cody:
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,17 +21,18 @@ using the `jetbrains-ide` & `team/integrations` labels.
     - https://plugins.jetbrains.com/plugin/8527-google-java-format
     - https://plugins.jetbrains.com/plugin/14912-ktfmt
 
-| What                                                          | Command                                                                  |
-|---------------------------------------------------------------|--------------------------------------------------------------------------|
-| Run the plugin locally                                        | `./gradlew :runIDE`                                                      |
-| Run the plugin locally with fresh build of Cody               | `./gradlew -PforceAgentBuild=true :runIDE`                               |
-| Run the plugin locally with fresh build of Code Search assets | `./gradlew -PforceCodeSearchBuild=true :runIDE`                          |
-| Build Code Search assets (separate terminal)                  | `pnpm build`                                                             |
-| Continuously re-build Code Search assets (separate terminal)  | `pnpm watch`                                                             |
-| Code Search "Find with Sourcegraph" window                    | `pnpm standalone && open http://localhost:3000/`                         |
-| Build deployable plugin                                       | `./gradlew buildPlugin` (artifact is generated in `build/distributions`) |
-| Reformat Java and Kotlin sources                              | `./gradlew spotlessApply`                                                |
-| Debug agent JSON-RPC communication                            | `tail -f build/sourcegraph/cody-agent-trace.json`                        |
+| What                                                             | Command                                                                  |
+|------------------------------------------------------------------|--------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -21,17 +21,18 @@ using the `jetbrains-ide` & `team/integrations` labels.
...
-| Build deployable plugin                                       | `./gradlew buildPlugin` (artifact is generated in `build/distributions`) |
-| Reformat Java and Kotlin sources                              | `./gradlew spotlessApply`                                                |
-| Debug agent JSON-RPC communication                            | `tail -f build/sourcegraph/cody-agent-trace.json`                        |
+| What                                                             | Command                                                                  |
+|------------------------------------------------------------------|--------------------------------------------------------------------------|
+| Run the plugin locally                                           | `./gradlew :runIDE`                                                      |
+| Run the plugin locally with fresh build of Cody                  | `./gradlew -PforceAgentBuild=true :runIDE`                               |
+| Run the plugin locally with fresh build of a local clone of Cody | `CODY_DIR=<path_to_cody> ./gradlew -PforceAgentBuild=true :runIDE`       |
```
mkondratek, line 28: tbd if we want to have it here

build.gradle.kts
```diff
@@ -228,7 +228,11 @@ tasks {
       println("Cached $buildCodyDir")
       return buildCodyDir
     }
-    val codyDir = unzipCody()
+    val codyDirEnv: String? =
+        System.getenv("CODY_DIR")?.replace("~", System.getProperty("user.home"))
```
mkondratek: for some reason `~` does not work 🙅

build.gradle.kts
```diff
@@ -228,7 +228,11 @@ tasks {
...
+    val codyDir: File =
+        if (codyDirEnv.isNullOrEmpty()) unzipCody() else Paths.get(codyDirEnv).normalize().toFile()
+    println("Using cody from codyDir=$codyDir")
```
mkondratek, line 239: imho some info can be helpful

build.gradle.kts
pkukielka: You could consider reading it from properties file as well.
It would give you more flexibility with overriding:

1) using env var, as in your orginal case (just with special prefix so it will be translated by gradle to property):
`ORG_GRADLE_PROJECT_codyDir=xxx`

2) inline when calling `gradle`command:
`gradle -PcodyDir=xxx`

3) set `codyDir` in `gradle.properies` if you want to make it pernament in a given workspace

Sa1to (approved):
LGTM

build.gradle.kts
```diff
@@ -217,6 +217,16 @@ tasks {
     return destination
   }
   fun unzipCody(): File {
+    val fromEnvironmentVariable = System.getenv("CODY_DIR")
+    if (!fromEnvironmentVariable.isNullOrEmpty()) {
+      val pathString =
+          if (fromEnvironmentVariable.startsWith("~")) {
```
mkondratek, line 220: I know, I could use a regex for that - I think this is more readable 📖

exigow:
Looks good. :+1:

build.gradle.kts
exigow: Not necessarily. The code is very straightforward. IMO if someone sets CODY_DIR, we don't need to additionally inform about that.

CONTRIBUTING.md
exigow: +1

PR #45, opened 2023-10-19
Remove explicit JUnit 5 dependency

exigow:
Additional changes:
* Restore ignored/failing test in URLBuilderTest.
* Migrate some tests from Java to Kotlin.
* Fix packaging in test directories (follow package-local convention).

Motivation:
JUnit is already bundled in org.jetbrains.intellij Gradle plugin. Bundled version enforces a way of implementing tests because all base classes for plugin development (like BasePlatformTestCase) are implemented in older version of JUnit. It is possible to mix both versions at the same time, but there is no benefit from it.

## Test plan

* Run `./gradlew test`
* Verify: all tests passed successfully


mkondratek (approved):
lgtm

olafurpg:
@exigow feel free to merge without my review!

PR #46, opened 2023-10-20
Simplify access token matching and add docs URL

willdollman:
Simplify token matching format, as the previous length-based validation missed an edge case. Also remove `sgph_` prefix as we're going ahead with `sgp_`.

~URL in comment is pending https://github.com/sourcegraph/sourcegraph/pull/57722 merging.~

## Test plan

- CI

### DIFF
--- a/src/main/java/com/sourcegraph/common/AuthorizationUtil.java
+++ b/src/main/java/com/sourcegraph/common/AuthorizationUtil.java
@@ -4,7 +4,7 @@ import org.jetbrains.annotations.NotNull;
 
 public class AuthorizationUtil {
   public static boolean isValidAccessToken(@NotNull String accessToken) {
-    // Sourcegraph access token formats: see https://docs.sourcegraph.com/dev/security/secret-formats
+    // Sourcegraph access token formats: https://docs.sourcegraph.com/dev/security/secret-formats
     return accessToken.isEmpty()
         || accessToken.length() == 40
         || accessToken.startsWith("sgp_");
--- a/src/main/java/com/sourcegraph/common/AuthorizationUtil.java
+++ b/src/main/java/com/sourcegraph/common/AuthorizationUtil.java
@@ -4,11 +4,9 @@ import org.jetbrains.annotations.NotNull;
 
 public class AuthorizationUtil {
   public static boolean isValidAccessToken(@NotNull String accessToken) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/sourcegraph/common/AuthorizationUtil.java
```diff
@@ -4,11 +4,7 @@
 
 public class AuthorizationUtil {
   public static boolean isValidAccessToken(@NotNull String accessToken) {
-    return accessToken.isEmpty()
-        || accessToken.length() == 40
-        || (accessToken.startsWith("sgp_") && accessToken.length() == 44)
-        // See https://docs.google.com/document/d/1aC4gHB8Q5lurwVhc8SCxznR0blNJMKo7yIkpP7WShno
-        || (accessToken.startsWith("sgp_") && accessToken.length() == 61)
-        || (accessToken.startsWith("sgph_") && accessToken.length() == 62);
+    // Sourcegraph access token formats: https://docs.sourcegraph.com/dev/security/secret_formats
```
mkondratek, line 7: image

Should the page be yet available? 🤔

willdollman: Not quite yet I'm afraid!

> URL in comment is pending https://github.com/sourcegraph/sourcegraph/pull/57722 merging.

willdollman: PR has merged and link now works @mkondratek https://docs.sourcegraph.com/dev/security/secret_formats

mkondratek: Thanks! It looks good. 

I am curious... what about `sgd_`, `slk_`, `sgs_` tokens? They are mentioned in the docs. However, they are invalid in the sense of this method. Is it ok? 🤔

willdollman: Yep not a problem. The jetbrains extension only needs to use access tokens (`sgp_`) which are used to talk to a sourcegraph instance. These other tokens are used for various other backend authentication purposes so don't need to be supported here 😄

mkondratek (approved):
ok, thanks for the explanations. makes sense. lgtm

PR #48, opened 2023-10-20
Add a notification about the missing token

### LINKED ISSUE #57373: JetBrains: missing authentication token but logged in
I've hit on this issue twice, in IntelliJ CE and Goland where it looks like I'm logged in but nothing works (chat/autocomplete). When I open settings I see the root problem, the authentication token is missing

Image

We should treat this error as being logged out.
```[tasklist]
### Tasks
- [x] Report notification when the user is missing an access token
- [ ] Display error icon in the status bar whe the user is missing an access token
```

mkondratek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/57373.

Note this is a notification part only! Not a solution to the root cause.

Demo:

https://github.com/sourcegraph/jetbrains/assets/19799111/b279dab5-6d07-4e1f-85fb-215a95691ec0


Sa1to (approved):
LGTM

exigow:
Looks great :+1:

@mkondratek, please note that commit message has `fixes` so we will need to re-open linked issue after merge.

PR #49, opened 2023-10-20
Fix rendering of html tags in the chat

### LINKED ISSUE #30: JetBrains: Pasted code breaks the chat message
Screenshot 2023-10-16 at 10 18 37

Pasting this code bock on chat causes some strange behaviours

```
function InfoGroup ({children}, props) {
return 
   <div className="group border border-neutral-800 p-2 rounded border-dashed mt-4">
      <div className="group-title border-neutral-800 border bg-neutral-900 text-center text-xs uppercase p-0.5 text-neutral-500 rounded font-plex-mono">
         {props.GroupName} <div>
      <div className-"grid-cols-2 grid gap-1.5 mt-1.5 "> {children} <div>
   </div>;
```

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/30.

I guess there is still an area for improvement - the chat could keep whitespaces. 
However, now the html tags are handled properly 🚀 

Demo:

https://github.com/sourcegraph/jetbrains/assets/19799111/3de7ebba-34d0-4b40-8fd9-b3b5c1e2ac2a

### DIFF
--- a/src/main/java/com/sourcegraph/cody/CodyToolWindowContent.java
+++ b/src/main/java/com/sourcegraph/cody/CodyToolWindowContent.java
@@ -23,6 +23,7 @@ import com.intellij.util.concurrency.annotations.RequiresEdt;
 import com.intellij.util.ui.JBUI;
 import com.intellij.util.ui.StatusText;
 import com.intellij.util.ui.UIUtil;
+import com.intellij.xml.util.XmlStringUtil;
 import com.sourcegraph.cody.agent.CodyAgent;
 import com.sourcegraph.cody.agent.CodyAgentManager;
 import com.sourcegraph.cody.agent.CodyAgentServer;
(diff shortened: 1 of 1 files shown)


exigow (approved):
Looks good :+1:

exigow (changes requested):
However, this might introduce significant issue. What if a user wants to paste some Python code (or any other language that relies on indentation)? Will it still be readable? I'm not sure if we are okay with that.

(These are my assumptions - I haven't tested it this way)

mkondratek:
image

The problem seems to be with the rendering only. See my example ☝️ please note that the issue with whitespaces already exists (I am not introducing it here), this PR fixes html tags only.

olafurpg:
Just confirmed that the lack of indentation is not a regression in this PR


CleanShot 2023-10-23 at 11 28 04@2x

exigow (approved):

PR #57, opened 2023-10-25
Add `testIntegration` task with Autocompletion Tests

exigow:
Add `testIntegration` task with Autocompletion Tests

Fixes https://github.com/sourcegraph/sourcegraph/issues/57976

Added `testIntegration` task with dedicated source set. If `SRC_ACCESS_TOKEN` is missing, these tests will fail.

Running integration testing is optional for developers.

### Test plan

* Add `SRC_ACCESS_TOKEN` environment variable with valid token.
* Run `./gradlew testIntegration`.
* Verify: `TriggerAutocompleteActionTest` passes.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,6 +28,7 @@ val isForceCodeSearchBuild = isForceBuild || properties("forceCodeSearchBuild") 
 
 plugins {
   id("java")
+  id("idea")
   // Dependencies are locked at this version to work with JDK 11 on CI.
   id("org.jetbrains.kotlin.jvm") version "1.9.10"
   id("org.jetbrains.intellij") version "1.15.0"
@@ -52,11 +53,27 @@ intellij {
   updateSinceUntilBuild.set(false)
 }
 
+sourceSets {
+  create("testIntegration") {
+    compileClasspath += sourceSets.main.get().output
+    runtimeClasspath += sourceSets.main.get().output
+  }
+}
+
+val testIntegrationImplementation: Configuration by
+    configurations.getting {
+      extendsFrom(configurations.implementation.get())
+      extendsFrom(configurations.testImplementation.get())
+    }
+
+idea { module { testSources.from(sourceSets["testIntegration"].kotlin.srcDirs) } }
+
 dependencies {
   implementation("org.commonmark:commonmark:0.21.0")
   implementation("org.commonmark:commonmark-ext-gfm-tables:0.21.0")
   implementation("org.eclipse.lsp4j:org.eclipse.lsp4j.jsonrpc:0.21.0")
   implementation("com.googlecode.java-diff-utils:diffutils:1.3.0")
+  testIntegrationImplementation(sourceSets.main.get().output)
 }
 
 spotless {
@@ -357,6 +374,18 @@ tasks {
     certificateChain.set(System.getenv("CERTIFICATE_CHAIN"))
     privateKey.set(System.getenv("PRIVATE_KEY"))
     password.set(System.getenv("PRIVATE_KEY_PASSWORD"))
+  }
+
+  task<Test>("testIntegration") {
+    doFirst {
+      systemProperty("cody-agent.directory", buildCodyDir.parent)
+      require(System.getenv("SRC_ACCESS_TOKEN") != null) {
+        "Missing SRC_ACCESS_TOKEN environment variable"
+      }
+    }
+    group = "verification"
+    testClassesDirs = sourceSets["testIntegration"].output.classesDirs
+    classpath = sourceSets["testIntegration"].runtimeClasspath
   }
 
   publishPlugin {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,7 +40,10 @@ group = properties("pluginGroup")
 
 version = properties("pluginVersion")
 
-repositories { mavenCentral() }
+repositories {
+  mavenCentral()
+  maven(url = "https://packages.jetbrains.team/maven/p/ij/intellij-dependencies")
+}
 
 intellij {
   pluginName.set(properties("pluginName"))
@@ -58,6 +61,10 @@ sourceSets {
     compileClasspath += sourceSets.main.get().output
     runtimeClasspath += sourceSets.main.get().output
   }
+  create("testUI") {
+    compileClasspath += sourceSets.main.get().output
+    runtimeClasspath += sourceSets.main.get().output
+  }
 }
 
 val testIntegrationImplementation: Configuration by
@@ -66,14 +73,29 @@ val testIntegrationImplementation: Configuration by
       extendsFrom(configurations.testImplementation.get())
     }
 
-idea { module { testSources.from(sourceSets["testIntegration"].kotlin.srcDirs) } }
+val testUIImplementation: Configuration by
+  configurations.getting {
+    extendsFrom(configurations.implementation.get())
+    extendsFrom(configurations.testImplementation.get())
+  }
+
+idea {
+  module {
+    testSources.from(sourceSets["testIntegration"].kotlin.srcDirs)
+    testSources.from(sourceSets["testUI"].kotlin.srcDirs)
+  }
+}
 
 dependencies {
   implementation("org.commonmark:commonmark:0.21.0")
   implementation("org.commonmark:commonmark-ext-gfm-tables:0.21.0")
   implementation("org.eclipse.lsp4j:org.eclipse.lsp4j.jsonrpc:0.21.0")
   implementation("com.googlecode.java-diff-utils:diffutils:1.3.0")
+  testImplementation("com.intellij.remoterobot:remote-robot:0.11.20")
+  testImplementation("com.intellij.remoterobot:remote-fixtures:0.11.20")
+  testImplementation("com.squareup.okhttp3:logging-interceptor:4.12.0")
   testIntegrationImplementation(sourceSets.main.get().output)
+  testUIImplementation(sourceSets.main.get().output)
 }
 
 spotless {
@@ -364,8 +386,10 @@ tasks {
   // Configure UI tests plugin
   // Read more: https://github.com/JetBrains/intellij-ui-test-robot
   runIdeForUiTests {
+    systemProperty("cody-agent.directory", buildCodyDir.parent)
     systemProperty("robot-server.port", "8082")
     systemProperty("ide.mac.message.dialogs.as.sheets", "false")
+    systemProperty("ide.show.tips.on.startup.default.value", "false")
     systemProperty("jb.privacy.policy.text", "<!--999.999-->")
     systemProperty("jb.consents.confirmation.enabled", "false")
   }
@@ -386,6 +410,12 @@ tasks {
     group = "verification"
     testClassesDirs = sourceSets["testIntegration"].output.classesDirs
     classpath = sourceSets["testIntegration"].runtimeClasspath
+  }
+
+  task<Test>("testUI") {
+    group = "verification"
+    testClassesDirs = sourceSets["testUI"].output.classesDirs
+    classpath = sourceSets["testUI"].runtimeClasspath
   }
 
   publishPlugin {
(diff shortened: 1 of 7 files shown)


exigow:
I managed to code a test that uses an external token to communicate with the real agent. I was able to successfully authorize it, simulate the autocompletion-related trigger, and it generally works. However, to make it fully automatic and non-flaky, I have to fix following problems:

* Completions are generated and applied to the document asynchronously. This makes writing tests challenging because we need to somehow synchronize the trigger action with produced CompletableFuture. I managed to address this by modifying the completion manager code to force a "completable.join()" but it's IMO too hacky. Another option is to, for example, wait few seconds after triggering the completion, but I believe this will significantly extend test execution times (and no one will want to run them becuase of that).
* Autocompletions are rendered with custom renderer (who's calling drawString) on a separate thread designated for AWT rendering. In short, the action triggers asynchronous code, which in turn asynchronously (again!) invokes the renderer. With this async-in-async code, the test becomes super flaky, and I'm working on finding a solution.
* Tests that require a token fail when running all tests. I need to find a solution to either make them ignored when there's no token or define a separate sourceSet and task for these tests (something like "./gradlew integrationTest").

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -168,6 +167,7 @@ class CodyAutocompleteManager {
     val autocompleteRequest =
         triggerAutocompleteAsync(
             project, editor, offset, textDocument, triggerKind, cancellationToken)
+    if (!isAsync) autocompleteRequest.join()
```
exigow, line 172: This code along with `disableAsyncAutocomplete` was added only for testing purposes. 

IMO this looks very bad, but I don't have any other idea on how to ensure that the CompletableFuture completes within this scope.

exigow: Any ideas?

Or we can just merge that and I'll fix that later.

src/test/kotlin/com/sourcegraph/cody/autocomplete/action/TriggerAutocompleteActionTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.intellij.testFramework.fixtures.BasePlatformTestCase
+import com.sourcegraph.cody.autocomplete.CodyAutocompleteManager
+
+class TriggerAutocompleteActionTest : BasePlatformTestCase() {
+
+  fun `test autocomplete`() {
+    // TODO: replace this condition with test framework-level feature (maybe
+    // something similar to SkipException from TestNG). Tests should be "yellow"
+    // (but still passing), not "green".
+    if (missingToken()) return
```
exigow: There is cool SkipException feature in TestNG where you can throw this during test and it will be marked as skipped (AKA "yellow"). I'm looking for something similar in JUnit 3. We can bump JUNit version, but this causes other issues.

Another option is to add this test to different source set and create separate task for integration tests. Something like `./gradlew testIntegration`.

exigow:
Also, I'm really considering using something like https://github.com/virtuslab/ide-probe.

mkondratek:
> Also, I'm really considering using something like https://github.com/virtuslab/ide-probe.

If you are looking for examples here is one 😄 
https://github.com/VirtusLab/git-machete-intellij-plugin

build.gradle.kts
```diff
@@ -354,6 +354,8 @@ tasks {
     password.set(System.getenv("PRIVATE_KEY_PASSWORD"))
   }
 
+  test { systemProperty("cody-agent.directory", buildCodyDir.parent) }
```
mkondratek: Is it needed? We have added `CODY_DIR` recently. Why can't we use it here?

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -45,9 +45,7 @@ import com.sourcegraph.utils.CodyEditorUtil.isImplicitAutocompleteEnabledForEdit
 import difflib.Delta
 import difflib.DiffUtils
 import difflib.Patch
-import java.util.concurrent.CancellationException
-import java.util.concurrent.CompletableFuture
-import java.util.concurrent.CompletionException
+import java.util.concurrent.*
```
mkondratek: btw are we ok with wildcard imports?

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
mkondratek: not great not terrible

src/main/kotlin/com/sourcegraph/cody/auth/SelectOneOfTheAccountsAsActive.kt
```diff
@@ -1,17 +1,29 @@
...
+
+  private val logger = Logger.getInstance(SelectOneOfTheAccountsAsActive::class.java)
+
   override fun runActivity(project: Project) {
-    val codyAuthenticationManager = CodyAuthenticationManager.instance
-    if (codyAuthenticationManager.getActiveAccount(project) == null) {
-      val newActiveAccount = codyAuthenticationManager.getAccounts().getFirstAccountOrNull()
-      codyAuthenticationManager.setActiveAccount(project, newActiveAccount)
+    val authManager = CodyAuthenticationManager.instance
+    val token = System.getenv("SOURCEGRAPH_TEST_TOKEN")
```
mkondratek: I think it is worth to rename it to `testToken` - to strongly reflect it's purpose. Or maybe even extract the test account set up to a separate private method.

src/main/kotlin/com/sourcegraph/cody/auth/SelectOneOfTheAccountsAsActive.kt
```diff
@@ -1,17 +1,29 @@
...
+    if (token != null) {
+      val testAccount = CodyAccount("test", "Test User")
+      authManager.updateAccountToken(testAccount, token)
+      authManager.setActiveAccount(project, testAccount)
+      logger.warn("Test account with external SOURCEGRAPH_TEST_TOKEN selected as active")
+    }
+    if (authManager.getActiveAccount(project) == null) {
```
mkondratek: huh! so you cannot have both test account and "regular" account! I assume this is intended

src/test/kotlin/com/sourcegraph/cody/autocomplete/action/TriggerAutocompleteActionTest.kt
mkondratek: `./gradlew testIntegration` sounds great and it seem to be a common practice

src/test/kotlin/com/sourcegraph/cody/autocomplete/action/TriggerAutocompleteActionTest.kt
```diff
@@ -0,0 +1,25 @@
+package com.sourcegraph.cody.autocomplete.action
+
+import com.intellij.testFramework.fixtures.BasePlatformTestCase
```
mkondratek: wow, I was not aware of these 😮

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
pkukielka: I have one idea but I did not very carefully verified all corner cases.
Looks like `myFixture.testAction(TriggerAutocompleteAction())` under the hood runs `CodyAutocompleteManager.instance.triggerAutocomplete(
        editor, offset, InlineCompletionTriggerKind.INVOKE)`.
All other calls of the `triggerAutocomplete` are with `InlineCompletionTriggerKind.AUTOMATIC`.
So you could simply check if `triggerKind != InlineCompletionTriggerKind.INVOKE` and it's something we are already doing in this class, e.g. to enchance logging verbosity.

src/main/kotlin/com/sourcegraph/cody/auth/SelectOneOfTheAccountsAsActive.kt
exigow: Correct. I believe that if someone explicitly sets external token, ignoring that token and switching to a real account would be undesirable.

src/test/kotlin/com/sourcegraph/cody/autocomplete/action/TriggerAutocompleteActionTest.kt
exigow: You mean BasePlatformTestCase? Honestly, I am very surprised. :smile: These fixtures are the most default way to test plugin functionalities. 

In general, I recommend everything that inherits from UsefulTestCase.

src/main/kotlin/com/sourcegraph/cody/auth/SelectOneOfTheAccountsAsActive.kt
exigow: You're right. I've cleaned this code now.

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
exigow: Thanks for noticing. IMO static imports should be forbidden with Spotless rules. I quickly skimmed through Spotless docs, but I don't see such option.

Fixed manually.

build.gradle.kts
exigow: Moved to `testImplementation` task.

Regards your question: If property `cody-agent.directory` is not defined, the tests are looking for an agent in default `/build/idea-sandbox/` path (default for base platform tests), which is invalid. At the same time, I wouldn't want the CODY_DIR variable to always be required to run the tests.

exigow:
Updated title & description to match PR content.

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
exigow: I'm going to address these issues in next PR.

build.gradle.kts
```diff
@@ -366,6 +383,18 @@ tasks {
     password.set(System.getenv("PRIVATE_KEY_PASSWORD"))
   }
 
+  task<Test>("testIntegration") {
+    doFirst {
+      systemProperty("cody-agent.directory", buildCodyDir.parent)
+      require(System.getenv("SOURCEGRAPH_TEST_TOKEN") != null) {
```
olafurpg: How do I specify the Sourcegraph URL? To use the Sourcegraph cli `src`, you need to set both `SRC_ENDPOINT` and `SRC_ACCESS_TOKEN`. It could make sense to respect those same environment variables.

olafurpg (commented):
I ran `./gradlew testIntegration` with the access token and it seems stuck waiting for some future to complete

```
<===========--> 91% EXECUTING [14m 46s]
> :testIntegration > 28 tests completed
> IDLE
> :testIntegration > Executing test com.sourcegraph...action.TriggerAutocompleteActionTest
```

build.gradle.kts
exigow: Thanks, I was not aware of that! I'm going to change that env var to `SRC_ACCESS_TOKEN`.

I'm still trying to figure out why the tests hangs in your case. They should take few seconds max.

Question: Do we currently need to support `SRC_ENDPOINT`? For now, I don't see any test case that would benefit from that. Of course, there is possibility that such variable will be needed future, but handling this address doesn't seem trivial at first glance. I think adding support for such a variable should be part of future PRs that adds tests that may require to customize this address.

exigow: Using SRC_ACCESS_TOKEN now.

exigow:
I have added a commit to this branch with chat test PoC because both proposals require similar changes in the code.

I have also expanded CONTRIBUTING.MD with simple introduction of how to run such tests because it is starting to get a bit complicated (integration tests require a token, UI tests require an IDE running in the background - both serve different purposes). The test on the recording is performed by intellij-ui-robot and checks the basic functionality of the chat. This is quite intrusive because it essentially hijacks your keyboard and mouse, but it's faster than manual testing.

chat_test

mkondratek:
FYI: I tried it locally. It failed:
image

PR #65, opened 2023-10-25
Add more intelligent agent rebuilding

olafurpg:
Previously, we didn't rebuild the agent if the binaries were available on disk. This was problematic because it meant that we don't automatically rebuild the binaries when bumping up the Cody commit. This PR fixes the problem by generating a new `cody-revision.txt` file alongside the agent binaries. This revision file includes the cody commit hash, and we automatically rebuild the binaries if this commit hash doesn't match the cody commit hash in the build.

## Test plan

- `:runIde`, confirm the agent is rebuild
- `:runIde` again, confirm the agent is cached
- Update `codyCommit` hash to a new commit hash
- `:runIde` again, confirm the agent is rebuilt

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -212,26 +212,56 @@ tasks {
     download(url, destination)
     return destination
   }
+  fun customCodyDir(): File? {
+    val fromEnvironmentVariable = System.getenv("CODY_DIR") ?: return null
+    // "~" works fine from the terminal, however it breaks IntelliJ's run configurations
+    val pathString =
+        if (fromEnvironmentVariable.startsWith("~")) {
+          System.getProperty("user.home") + fromEnvironmentVariable.substring(1)
+        } else {
+          fromEnvironmentVariable
+        }
+    return Paths.get(pathString).toFile()
+  }
+
   fun unzipCody(): File {
-    val fromEnvironmentVariable = System.getenv("CODY_DIR")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -212,26 +212,62 @@ tasks {
...
+            "agent-macos-arm64",
+            "agent-macos-x64",
+            "agent-win-x64.exe",
+            "cody-revision.txt")
+    if (children.sorted() != expectedBinaries.sorted()) {
+      return false
+    }
+
+    val revision = codyRevisionPath.readText()
+    val revisionHasChanged = revision == codyCommit
```
pkukielka, line 261: If `revision == codyCommit` doesn't that mean revision has **not** changed?

build.gradle.kts
```diff
@@ -212,26 +212,62 @@ tasks {
...
+        listOf(
+            "agent-linux-arm64",
+            "agent-linux-x64",
+            "agent-macos-arm64",
+            "agent-macos-x64",
+            "agent-win-x64.exe",
+            "cody-revision.txt")
+    if (children.sorted() != expectedBinaries.sorted()) {
+      return false
+    }
```
exigow, line 258: If we replace list with set, sorting will be unnecessary.

exigow (changes requested):

olafurpg:
Closing this. Feel free to pick it up you want this feature

PR #67, opened 2023-10-25
Do not run onFileOpened on caret position change

### LINKED ISSUE #62: JetBrains: `git/codebaseName` request on every keystroke
to reproduce: move the cursor and notice that we are sending a `git/codebaseName` request to the agent on every keystroke

Image

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/62.

Before this change, the call was performed from here: `CodyCaretListener::caretPositionChanged`.

### Remate change

https://github.com/sourcegraph/jetbrains/assets/19799111/9db7c4fd-d483-4681-9bd0-06dee9417c24

### Agent calls & keystrokes

https://github.com/sourcegraph/jetbrains/assets/19799111/0f834322-8740-4c3d-b877-ce88fdd76e23

## Test plan 
- [x] 1. Manual - the number of calls test
  1. `runIde`
  2. `tail -f build/sourcegraph/cody-agent-trace.json`
  3. verify the number of `git/codebaseName` calls is reasonable (not on every keystroke)
- [x] 2. Manual - the codebase name update
  1. `runIde`
  2. Search for action (cmd + shift + A), `Manage Remotes...`, change the remote URL
  3. verify that the repository name updates 

Note that a change to url from outside - e.g. terminal, vim, .git/config - will not trigger the update. It never worked like this.


src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.java
```diff
@@ -186,7 +186,8 @@ private static Range getSelection(Editor editor) {
   }
 
   // Sends a textDocument/didChange notification to the agent server.
-  public static void informAgentAboutEditorChange(@Nullable Editor editor) {
+  public static void informAgentAboutEditorChange(
+      @Nullable Editor editor, boolean skipCodebaseOnFileOpened) {
```
olafurpg: nit: I generally try to avoid boolean parameters like this where the majority of the call-sites pass the same value (`true` in this case here). One problem with this approach is that the method becomes difficult to reason about when you add a second boolean parameter the next time you have to customize the behavior even more. Another problem is that passing in a boolean `true` at call-site makes it difficult to understand what the `true` value does.

- A cheap workaround is to at add an overload without the boolean parameter.
- A better workaround is to rewrite this code into Kotlin so that you can use default parameter values and named argument.

We should aim to rewrite everything into Kotlin so I vote for the second option.

src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.kt
```diff
@@ -0,0 +1,187 @@
...
+      completions.clearAutocompleteSuggestions(editor)
+      if (isImplicitAutocompleteEnabledForEditor(editor) &&
+          isEditorValidForAutocomplete(editor) &&
+          !CommandProcessor.getInstance().isUndoTransparentActionInProgress) {
+        informAgentAboutEditorChange(editor)
+        val changeOffset = event.offset + event.newLength
+        if (editor.caretModel.offset == changeOffset) {
+          val requestType =
+              if (event.oldLength != event.newLength) InlineAutocompleteTriggerKind.Invoke
+              else InlineAutocompleteTriggerKind.Automatic
```
mkondratek: @olafurpg, can you take a look at this part?

`requestType` is not used. I think it is safe to delete it.

src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.java
mkondratek: moved to Kotlin! 🎉

src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.kt
olafurpg: Fine to remove it if it's not used. We hardcode `AUTOMATIC` below, which makes sense since this event is triggered on document changes, not the alt-backslash shortcut.

src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.kt
```diff
@@ -0,0 +1,183 @@
...
+                Position()
+                    .setLine(selectionStartPosition.line)
+                    .setCharacter(selectionStartPosition.column))
+            .setEnd(
+                Position()
+                    .setLine(selectionEndPosition.line)
+                    .setCharacter(selectionEndPosition.column))
+      }
+      val carets = editor.caretModel.allCarets
+      if (!carets.isEmpty()) {
```
Sa1to: You can use `carets.isNotEmpty()`

src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.kt
```diff
@@ -0,0 +1,183 @@
...
+        val changeOffset = event.offset + event.newLength
+        if (editor.caretModel.offset == changeOffset) {
+          completions.triggerAutocomplete(
+              editor, changeOffset, InlineCompletionTriggerKind.AUTOMATIC)
+        }
+      }
+    }
+  }
+
+  companion object {
```
Sa1to: I'm getting this warning from IDE. Maybe it makes sense to create separate class for this part 😅 
Screenshot 2023-10-26 at 15 02 18

exigow (approved):
Looks good :+1:

Issue seems to be fixed. However, during testing, I noticed that codebase call is executed twice when we change file. From what I can see, this issue is present also on main branch so it's not related to your changes.

src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.kt
```diff
@@ -0,0 +1,183 @@
...
+    fun informAgentAboutEditorChange(editor: Editor?, skipCodebaseOnFileOpened: Boolean = true) {
+      if (editor == null) {
+        return
+      }
+      if (editor.project == null) {
+        return
+      }
+      if (!isConnected(editor.project!!)) {
+        return
+      }
```
Sa1to, line 159: Minor nitpick: isn't it better to just combine those if statements into one condition?

Sa1to (changes requested):
Issue seems to be resolved. I just have few minor concerns regarding the code 😄

src/main/java/com/sourcegraph/cody/autocomplete/CodyEditorFactoryListener.kt
exigow: Nice catch! How do you invoked this analysis? I'm trying to reproduce this locally, just for curiosity.

Sa1to (approved):
LGTM

PR #71, opened 2023-10-26
Update next-release.sh to make it work on Linux

exigow:
Changes in script:
* Choose between `gdate` and `date`
* Verify `gdate` or `date` is installed
* Verify `gh` is installed
* Update `cut` usage (not sure why, but -fX fields as separate arguments are unsupported on Linux)

### Test Plan

* Run `./next.release.sh` on macOS or Linux
* Expect output similar to `5.2.4100`

### DIFF
--- a/scripts/next-release.sh
+++ b/scripts/next-release.sh
@@ -7,21 +7,38 @@
 # https://docs.google.com/document/d/11cw-7dAp93JmasITNSNCtx31xrQsNB1L2OoxVE6zrTc/edit#bookmark=id.ufwe0bqp83z1
 set -eu
 
-if ! command -v gdate &>/dev/null; then
-  echo "command not found: gdate"
-  echo "The command gdate is required to compute the next version number"
-  echo "To fix this problem, run:\n  brew install coreutils"
+if [[ "$(uname)" == "Darwin" ]]; then
+  if ! command -v gdate &>/dev/null; then
+    echo "Command not found: gdate"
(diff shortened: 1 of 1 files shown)


exigow:
As a formality: I confirm that it works on Linux.

olafurpg (approved):
Just ran the script on my macOS to confirm it works correctly 👍🏻

olafurpg:
BTW, this script will get replaced with a new `sg next-version` command. The tool `sg` can be installed in the sourcegraph/sourcegraph repository.

scripts/next-release.sh
```diff
@@ -7,21 +7,38 @@
...
+    echo "Command not found: date"
+    exit 1
+  fi
+  date_program() {
+    date "$@"
+  }
 fi
 
+if ! command -v gh &>/dev/null; then
+  echo "Command not found: gh"
```
mkondratek, line 31: would it make sense to add a similar message like for gdate? I mean smth like
```
    echo "The command gh is required to ..."
    echo "To fix this problem, run:\n  brew install gh"
```

mkondratek:
I do confirm it works for me as expected! 🚀

mkondratek (approved):

scripts/next-release.sh
exigow: Naah, I'm just merging - looks like it is going to be removed soon. But thanks for comment!

PR #76, opened 2023-10-31
Extract components from CodyToolWindowContent

mkondratek:
This is a refactor only. No changes to functionalities.

It will help implement the new send button design (https://github.com/sourcegraph/sourcegraph/issues/56678).
Another good reason is that `CodyToolWindowContent` was becoming a bit messy.

## Test plan
- [x] Chat history
- [x] Send prompts
    - [x] "Send" button
    - [x] Enter keyboard button
- [x] Stop generation
- [x] Prompt overflow

### DIFF
--- a/src/main/java/com/sourcegraph/cody/ControlsPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/ControlsPanel.kt
@@ -1,0 +1,21 @@
+package com.sourcegraph.cody
+
+import com.intellij.util.ui.JBUI
+import java.awt.BorderLayout
+import javax.swing.JButton
+import javax.swing.JPanel
+import javax.swing.border.EmptyBorder
+
+class ControlsPanel(
+    promptPanel: PromptPanel,
+    sendButton: JButton,
+) : JPanel() {
+
+  init {
+    layout = BorderLayout()
+    border = EmptyBorder(JBUI.insets(0, 14, 14, 14))
+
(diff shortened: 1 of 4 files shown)


src/main/java/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -84,72 +73,12 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
-    promptInput.document.addDocumentListener(
-        object : DocumentAdapter() {
-          override fun textChanged(e: DocumentEvent) {
-            sendButton.isEnabled = promptInput.getText().isNotEmpty()
-          }
-        })
-    promptPanel.add(autoGrowingTextArea.scrollPane, BorderLayout.CENTER)
-    promptPanel.border = BorderFactory.createEmptyBorder(0, 0, 10, 0)
+    val promptPanel =
+        PromptPanel(project, chatMessageHistory, ::sendChatMessage) { sendButton.isEnabled = it }
```
exigow: IMO there is small issue with readability - I had to check the implementation of the constructor to make sure what we are doing with this trailing lambda.

Suggestion: we should use named parameter here, like `onTextChanged = { sendButton.isEnabled = it }`

exigow (approved):
Looks good :+1:

Sa1to (approved):
LGTM

PR #77, opened 2023-10-31
Apply new design to prompt window and send button

### LINKED ISSUE #56678: JetBrains: improve design of Chat "Send" button

mkondratek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/56678.

### Demo
image

## Test Plan
todo


danielmarquespt:
@mkondratek 

Looks good, but we need to respect the paddings of the overall sidebar like this:

Screenshot 2023-11-01 at 10 46 40


Can we also display an icon instead of "Send"? And remove the stylings of the button (Border and shadow) so that we display only the icon and have it acting like a button?

mkondratek:
## Demo 

<details>
<summary> Demo video </summary>


https://github.com/sourcegraph/jetbrains/assets/19799111/3a562eee-6c7a-4db3-b0d3-54ad7e36eab7



</details>

exigow (changes requested):
Moving Cody toolbar from right panel to bottom panel introduces some issue with positioning:

image

exigow (commented):
I also see a small difference in the font shape on my PC. Note that the arrow ("►") is not properly centered. Maybe we should use SVG here? We would have a guarantee that it looks the same everywhere.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -258,16 +257,15 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
+    val myButton = TransparentButton("►")
+    myButton.addActionListener { _: ActionEvent? ->
       GraphQlLogger.logCodyEvent(this.project, "recipe:chat-question", "clicked")
       sendChatMessage()
     }
-    sendButton.isEnabled = false
-    return sendButton
+    myButton.isEnabled = false
+    myButton.isVisible = true
+    myButton.setToolTipText("Send message")
```
exigow: We can replace this setter with assignment.

src/main/kotlin/com/sourcegraph/cody/ui/AutoGrowingTextArea.kt
```diff
@@ -23,7 +23,7 @@ import javax.swing.text.PlainDocument
 import kotlin.math.max
 import kotlin.math.min
 
-class AutoGrowingTextArea(private val minRows: Int, maxRows: Int, outerPanel: JPanel) {
+open class AutoGrowingTextArea(private val minRows: Int, maxRows: Int, outerPanel: JComponent) {
```
exigow: I think we can remove `open` keyword. Nothing inherits from it.

src/main/kotlin/com/sourcegraph/cody/ui/TransparentButton.kt
```diff
@@ -1,16 +1,19 @@
...
 class TransparentButton(text: String) : JButton(text) {
-  private val opacity = 0.7f
+  private val enabledOpacity = 0.7f
+  private val disabledOpacity = 0.4f
   private val cornerRadius = 5
   private val horizontalPadding = 10
   private val verticalPadding = 5
-  private val textColor = UIUtil.getLabelForeground()
+  private val enabledTextColor = UIUtil.getLabelForeground()
+  private val disabledTextColor = ColorUtil.darker(UIUtil.getLabelForeground(), 3)
```
exigow: I think most of these variables can be locally defined in function bodies. I think it's unnecessary to place them in the class field if we only use them once. It needlessly clutters the code, unless we're doing some optimizations here - please correct me if yes.

src/main/kotlin/com/sourcegraph/cody/ui/TransparentButton.kt
```diff
@@ -28,14 +31,22 @@ class TransparentButton(text: String) : JButton(text) {
...
-    g2.composite = AlphaComposite.SrcOver.derive(opacity)
+
+    if (isEnabled) {
+      g2.composite = AlphaComposite.SrcOver.derive(enabledOpacity)
+      g.color = enabledTextColor
+    } else {
+      g2.composite = AlphaComposite.SrcOver.derive(disabledOpacity)
+      g.color = disabledTextColor
+    }
+
```
exigow, line 38: Can you please elaborate on this condition? Can't we just hardcode that transparency is always "on"?

mkondratek: Transparency is always on. This condition only changes the styling a bit b/w the two states of the button (enabled/disabled)

mkondratek:
> Moving Cody toolbar from right panel to bottom panel introduces some issue with positioning:
> 
> <image>

fixed!

exigow (changes requested):
Thanks for the fixes! 

However, I noticed that there are still a few issues. Please take a look:

image

1. Vertical padding size != horizontal padding (9px vs 8px). It's just one pixel difference, but it's noticeable.

2. Placeholder text (`Ask (...)`) is not respecting padding the same way the button does. @danielmarquespt please correct me if I'm wrong.

3. The arrow is not correctly centered on my Linux. I suspect this is due to the font being slightly different from the one on your Mac. I think the best solution would be to replace this character with built-in icon to avoid this kind of issues. See:
    * https://jetbrains.design/intellij/resources/icons_list/
    * `package com.intellij.icons`

4. The positioning issue still persists. I've also noticed that the button is repainted only after resizing the window. GIF:

send-button-issues

mkondratek:
@exigow your idea about using an icon instead of text was great - it made the button impl much simpler. Can you pls test if it works well on your machine now? 3️⃣ should be resolved now.

mkondratek:
@exigow about 4️⃣ - are you sure you tested it against the newest version - I cannot reproduce that issue

mkondratek:
Btw. here is the newest demo:

<details>
<summary>video</summary>


https://github.com/sourcegraph/jetbrains/assets/19799111/18b29910-f3f9-45da-bb9c-cfba7741e0c3



</details>

mkondratek:
@exigow I do not understand the number 1️⃣ can you elaborate?

danielmarquespt:
@mkondratek  Overall it looks ok, I would just remove the borders and move it a bit more into the corner so that it does not overlap the text.

If possible an hover state would be cool as well, but that can be added afterwards

Screenshot 2023-11-24 at 09 51 52

exigow (approved):

exigow (approved):
Hey @mkondratek, I noticed that you're back from sick leave. I hope you're feeling better now.

I reviewed your changes again and it looks like all the issues have been fixed or are no longer relevant. I played around with the new Send button and tried to break it, but to no avail. Nice work! :+1:

PR #78, opened 2023-10-31
Add Check for plugin updates option to the settings

### LINKED ISSUE #68: JetBrains: add setting to update the plugin
It's currently quite difficult to stay up to date with nightly builds. I do the following steps to force a plugin update when I know that there is a  nightly build

- "Manage plugin repositories" and press OK
- Update

It's unreasonable to expect internal users to do the same when we cut new releases. To solve this problem, we should add a checkbox setting next to the update channel where users can enable/disable the behavior that the plugin automatically checks for updates and maybe even sends a balloon notification when an update is available.

Image

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/68.

I checked it against IntelliJ 2022.1 - it works 🎉 
I did `runPluginVerifier` - it works too 🟢 

### Demo

https://github.com/sourcegraph/jetbrains/assets/19799111/e99cc100-64b4-4fb9-81d9-36f3cb64a13e

## Test plan
1. `runIde` with these changes
2. play with `Nightly`/`Stable` switch
(in the case when no updates appear you can try to modify `pluginVersion` in `gradle.properties` and re-run)


src/main/kotlin/com/sourcegraph/cody/config/ui/CheckUpdatesTask.kt
```diff
@@ -0,0 +1,81 @@
...
+            UpdateChecker.javaClass.getMethod(
+                "getInternalPluginUpdates", BuildNumber::class.java, ProgressIndicator::class.java)
+        val internalPluginUpdates = getInternalPluginUpdatesMethod.invoke(null, null, indicator)
+        val getPluginUpdatesMethod = internalPluginUpdates.javaClass.getMethod("getPluginUpdates")
+        val pluginUpdates = getPluginUpdatesMethod.invoke(internalPluginUpdates)
+        val getAllEnabledMethod = pluginUpdates.javaClass.getMethod("getAllEnabled")
+        val allEnabled = getAllEnabledMethod.invoke(pluginUpdates)
+        return allEnabled?.let { it as (Collection<PluginDownloader>) } ?: emptyList()
+      } catch (e: Exception) {
+        when (e) {
```
Sa1to, line 47: Shouldnt we also catch `ClassCastException` here?

src/main/kotlin/com/sourcegraph/cody/config/ui/AccountConfigurable.kt
```diff
@@ -71,13 +69,17 @@ class AccountConfigurable(val project: Project) :
                   SimpleListCellRenderer.create("") { it.presentableText })
               .bindItem({ channel }, { channel = it!! })
         }
+        row {
+          checkBox("Check for plugin updates").bindSelected(settingsModel::shouldCheckForUpdates)
```
danielmarquespt: "Automatically check for plugin updates"

Other than this, LGTM

Sa1to:
I dont know if it is important issue but i noticed that when there is new version of nightly and stable release im getting new notification whenever i switch between updating nightly and stable. Should it work this way? Maybe we can remove the previous popup 🤔 

https://github.com/sourcegraph/jetbrains/assets/9321940/6b5fbaa8-c871-4171-81e5-acb83b997854

mkondratek:
> I dont know if it is important issue but i noticed that when there is new version of nightly and stable release im getting new notification whenever i switch between updating nightly and stable. Should it work this way? Maybe we can remove the previous popup 🤔
> 
>  Screen.Recording.2023-11-07.at.13.04.50.mov

I'd say it's expected. It is possible that your currently installed version is older than available stable and nightly - in both cases the notification seems to be justified.

Sa1to (approved):
LGTM

PR #81, opened 2023-11-03
Refactor sourcegraph API request framework

Strum355:
As part of an effort to reduce the complexity and cognitive load of the codebase, this PR tackles refactoring parts of the sourcegraph API request framework to simplify and remove unused code.

Reviewing tip: `SourcegraphApiRequestExecutor.kt`  is best reviewed with whitespace changes hidden, as most of the lines are just a re-indentation from removing of nesting

## Test plan

Tested whether the avatar & current user details are successfully loaded in the Manage Accounts menu upon a fresh sign-in on a clean-slate `gradle :runIde` run


efritz:
LGTM (let's get that money)

src/main/kotlin/com/sourcegraph/cody/api/SourcegraphApiRequestExecutor.kt
```diff
@@ -18,160 +16,122 @@ import java.io.InputStream
...
-          .accept(request.acceptMimeType)
-    }
+  private fun createRequestBuilder(request: SourcegraphApiRequest<*>): RequestBuilder {
+    return when (request) {
+          is SourcegraphApiRequest.Get -> HttpRequests.request(request.url)
+          is SourcegraphApiRequest.Post -> HttpRequests.post(request.url, request.bodyMimeType)
+        }
+        .userAgent("Cody")
+        .throwStatusCodeException(false)
+        .forceHttps(false)
```
mkondratek, line 88: I see the former version had `.accept(request.acceptMimeType)` but it is not present here. Can you pls verify if it is ok? 🙏

Strum355: It was null by default and never set anywhere, so this should be fine. The green path worked fine, do you think there might be a default value that null unsets perhaps?

Strum355: Have confirmed this is fine, we never set it on `main` branch, so we're just passing in null when the default is already null

mkondratek: thanks for verifying!

src/main/kotlin/com/sourcegraph/cody/api/SourcegraphApiRequest.kt
```diff
@@ -1,53 +1,34 @@
...
-
-  abstract class Get<T>
-  @JvmOverloads
-  constructor(url: String, override val acceptMimeType: String? = null) :
-      SourcegraphApiRequest<T>(url) {
-    abstract class Optional<T>
-    @JvmOverloads
-    constructor(url: String, acceptMimeType: String? = null) : Get<T?>(url, acceptMimeType)
-  }
+  fun withOperationName(name: String): SourcegraphApiRequest<T> =
```
mkondratek: nit: this method is used once - it could be inlined

src/main/kotlin/com/sourcegraph/cody/api/SourcegraphApiRequestExecutor.kt
```diff
@@ -18,160 +16,122 @@ import java.io.InputStream
...
+        .useProxy(useProxy)
+        .execute(request, indicator)
+  }
 
   @TestOnly
   @RequiresBackgroundThread
   @Throws(IOException::class, ProcessCanceledException::class)
   fun <T> execute(request: SourcegraphApiRequest<T>): T = execute(EmptyProgressIndicator(), request)
 
-  fun addListener(listener: AuthDataChangeListener, disposable: Disposable) =
```
mkondratek, line 40: wow, it is unused 😮 I wonder why intelliJ does not gray out them

mkondratek (approved):
lgtm

PR #82, opened 2023-11-03
Add scip-java workflow

Strum355:
Yoinked from https://sourcegraph.com/github.com/yoohaemin/decrel/-/blob/.github/workflows/sourcegraph.yml

## Test plan

N/A, non-critical


varungandhi-src (approved):
Stamp

olafurpg (approved):

PR #86, opened 2023-11-06
chore: move kotlin files to src/main/kotlin from src/main/java

Strum355:
Does what it says on the tin

## Test plan

If it compiles :+1:


mkondratek (approved):
lgtm

PR #87, opened 2023-11-06
Update bug report template and add new PRD issue template

kalanchan:
Aligning JetBrains with VS Code templating structure. Product will use these "PRD-Issues" to submit user stories and will be used to track against feature development. 

Also converting issue template to issue forms to leverage new templating capabilities
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

examples:
image
image

## Test plan
n/a


olafurpg (approved):
LGTM 👍🏻 Thank you!

PR #99, opened 2023-11-07
Convert GraphQL event logger to Kotlin

Strum355:
Difference in output is kept to a minimum:
1. `hashedLicenseKey` and `connectedSiteID` are removed, as they were never set to anything besides `null`
2. `anonymousUserId` is also emitted with the same value as `deviceID` and `userCookieID`, the agent seems to reference this field in a few places

Telemetry will likely be reworked down the line to the new system.

## Test plan

Compared the jsonrpc trace output for the graphql/logEvent request


exigow (commented):
@Strum355, could you please describe in more detail how to verify these trace logs? I clicked through this plugin build, and unfortunately, no `graphql/logEvent` logs are persent in `cody-agent-trace.json`.

Maybe I'm looking in the wrong file? :sweat_smile:

Strum355:
> @Strum355, could you please describe in more detail how to verify these trace logs? I clicked through this plugin build, and unfortunately, no `graphql/logEvent` logs are persent in `cody-agent-trace.json`.
> 
> Maybe I'm looking in the wrong file? 😅

This PR can probably be closed, as a completely different direction is being taken (moving most of this to Agent instead of in-plugin). Sorry if you spent any time on this! 🙂

exigow:
Sure, no problem!

PR #110, opened 2023-11-09
Handle race between querying for VCS repositories and them being loaded by IntelliJ

Strum355:
See: https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000667650-Ensure-initialization-of-GitRepositoryManager-before-doing-plugin-operation and https://intellij-support.jetbrains.com/hc/en-us/community/posts/206105769-Get-project-git-repositories-in-a-project-component

## Test plan

TODO

### DIFF
--- a/src/main/java/com/sourcegraph/vcs/RepoUtil.java
+++ b/src/main/java/com/sourcegraph/vcs/RepoUtil.java
@@ -4,7 +4,7 @@ import com.intellij.dvcs.repo.Repository;
 import com.intellij.dvcs.repo.VcsRepositoryManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.vcs.impl.ProjectLevelVcsManagerImpl;
+import com.intellij.openapi.vcs.ProjectLevelVcsManager;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.vcsUtil.VcsUtil;
 import com.sourcegraph.cody.agent.CodyAgent;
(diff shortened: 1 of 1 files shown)


src/main/java/com/sourcegraph/vcs/RepoUtil.java
```diff
@@ -223,11 +224,22 @@ public static VCSType getVcsType(@NotNull Project project, @NotNull VirtualFile
...
+              synchronized (lock) {
+                lock.notify();
+              }
+            });
+    synchronized (lock) {
+      try {
+        lock.wait();
+      } catch (InterruptedException ignored) {
+      }
+    }
```
exigow, line 241: I did some research to see if there was a better solution and didn't find any other official way similar to this "lock & wait". As I understand it, the correct approach would be to hook into the appropriate event, but this would require rewriting part of this repo identification code to be asynchronous. I think we can accept the proposed solution as a temporary one.

BTW is there difference between `ProjectLevelVcsManager.getInstance(project)` and `ProjectLevelVcsManagerImpl.getInstanceImpl(project)`?

Strum355: I can switch it back to `ProjectLevelVcsManager`, originally was testing with some methods only exposed on the class but not needed anymore

exigow (changes requested):
@Strum355, could you please add the missing Test Plan required for reproducing this race condition?

Strum355:
> @Strum355, could you please add the missing Test Plan required for reproducing this race condition?

I dont believe I was able to reproduce this myself, I passed on a local build containing this fix but I havent heard back yet https://sourcegraph.slack.com/archives/C059N5FRYG3/p1699545417832259?thread_ts=1699461891.240879&cid=C059N5FRYG3

exigow (approved):

PR #125, opened 2023-11-14
Improve Send button & Chat user experience

mkondratek:
- Fix the send button enable/disable state - now it is always disabled when there is no text in the prompt 
- Add placeholder text to the prompt 
- Fix chat history after sending the prompt with "Send" button

## Test plan
- play with prompt text area, send some messages, play with chat history
- all should work as expected in the case of sending with enter keyboard button and "Send" ui button

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,3 @@ If you have any questions, feedback, or bug report, we appreciate if you [open a
   - Mac: Go to `IntelliJ IDEA | Preferences` (or use <kbd>⌘,</kbd>)
 - Click `Plugins` in the left-hand pane, then the `Installed` tab at the top
 - Find `Sourcegraph` → Right click → `Uninstall` (or uncheck to disable)
-
-## Version History
-
-See [`CHANGELOG.md`](https://github.com/sourcegraph/jetbrains/blob/main/CHANGELOG.md).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -208,7 +208,3 @@ If you have any questions, feedback, or bug report, we appreciate if you [open a
   - Mac: Go to `IntelliJ IDEA | Preferences` (or use <kbd>⌘,</kbd>)
 - Click `Plugins` in the left-hand pane, then the `Installed` tab at the top
 - Find `Sourcegraph` → Right click → `Uninstall` (or uncheck to disable)
-
-## Version History
```
mkondratek, line 212: unrel: deprecated

mkondratek:
## Demo


https://github.com/sourcegraph/jetbrains/assets/19799111/24a1e337-812e-40ac-8bbb-665545b7dbbf

Sa1to (approved):
LGTM. I just have a side note:
Maybe we should disable send button until chat response stopped generating text. When you send another message while chat response is being generated, text start to glitch and you won't get answer on a second message anyway:

https://github.com/sourcegraph/jetbrains/assets/9321940/d53f20c1-f0e5-444f-83b4-426167e699c2

mkondratek:
> LGTM. I just have a side note: Maybe we should disable send button until chat response stopped generating text. When you send another message while chat response is being generated, text start to glitch and you won't get answer on a second message anyway:
> 
>  <video>

Yeah. It's here for a long time already. I will check if we have an issue for this one and create one one (or maybe I will manage to just fix faster 😅 )

PR #88, opened 2023-11-06
Convert & consolidate protocol types in Kotlin

Strum355:
As described in the title. Reviewing on Github might be less optimal due to the order of changed files, I've found that the following is very convenient:
1. checkout the branch locally (`git fetch && git checkout nsc/protocol-kt`)
2. load the changes into the staging area (`git reset --soft HEAD~1`)
3. in the vscode "Source control" view, sort the changes by (file) name (see screenshot at the very bottom[1])

Some notes:
1. The two ChatMessage types were collapsed into one, the [issue described regarding the enum serialization](https://github.com/sourcegraph/jetbrains/compare/nsc/protocol-kt?expand=1#diff-9c533c092b1847a0028f4f0b8d02ceab7a330c8805b5b4df79fe45a012b33d3bL6-L7) appears to be [handled by `@SerializedName`](https://github.com/sourcegraph/jetbrains/compare/nsc/protocol-kt?expand=1#diff-c514334703915e2ca2b644739ef330a3993bd9745c2fc27c0b444b22d38f3c5aR6-R7)
2. Some field types were updated to reflect the current state of the types in the https://github.com/sourcegraph/cody repo, which saves us some null checking in some areas (e.g. [title and id on `RecipeInfo`](https://github.com/sourcegraph/jetbrains/compare/nsc/protocol-kt?expand=1#diff-335702a83e5c6967876175457e9f02c4c87ccd4b2b6bde508476c6b3f197092fL249-L251)
3. There are some TODOs that are best done when further classes are converted to Kotlin e.g. [the `Event` type (which is in need of updating)](https://github.com/sourcegraph/jetbrains/compare/nsc/protocol-kt?expand=1#diff-135b4c57ef38eece3f8fb3ee1ada17a0fd807a1382f1789d311b0308cb2faa37R8), [a `@JvmOverloads` on `TextDocument`](https://github.com/sourcegraph/jetbrains/compare/nsc/protocol-kt?expand=1#diff-15b5c585ca94aa06d1bca657c19df89aa1e10f6e186dd12f9f3c5c36eb6fbc46R4-R5)
[...]

PR #129, opened 2023-11-15
Increasing autocompletion trigger sensitivity

### LINKED ISSUE #57366: JetBrains: automatically trigger autocomplete in specific location
When typing `prompt` in the example from this tweet here, autocomplete doesn't trigger automatically. I have to alt-backslash to get autocomplete loading

 https://x.com/_PabloDev/status/1709725732270751956?s=20

We should fix it so that autocomplete is triggered automatically here. Minimal example to reproduce

```ts
interface DoInteractionRequest {
    chatId: string
    prompt: string
    systemPrompt: string
}

interface Fuzz {
    string: () => string
}

const z: Fuzz = {} as any

const interactionRequest: DoInteractionRequest = {
    chatId_CURSOR_
    prompt: z.string(),
    systemPrompt: z.string(),
}

```

Sa1to:
It fixes [JetBrains: automatically trigger autocomplete in specific location#57366](https://github.com/sourcegraph/sourcegraph/issues/57366) and fixes problem with incorrect serialization of `triggerKind` parameter, sent to the agent + fixing range calculation for lookup manager autocompletion.

## Test plan

* Autocompletion is automatically triggered for this example: 
```
interface DoInteractionRequest {
    chatId: string
    prompt: string
    systemPrompt: string
}

interface Fuzz {
    string: () => string
}

const z: Fuzz = {} as any

const interactionRequest: DoInteractionRequest = {
    chatId_CURSOR_
    prompt: z.string(),
    systemPrompt: z.string(),
}
```

https://github.com/sourcegraph/jetbrains/assets/9321940/9f475127-ada6-4e0b-811d-69cc178f12c5

* Lookup manager autocompletions are correctly triggered

https://github.com/sourcegraph/jetbrains/assets/9321940/2bec91b3-cac5-470e-abee-9b2dc335c282

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/AutocompleteParams.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/AutocompleteParams.kt
@@ -1,15 +1,13 @@
 package com.sourcegraph.cody.agent.protocol
 
-import com.google.gson.annotations.SerializedName
-
-enum class AutocompleteTriggerKind {
-  @SerializedName("Automatic") AUTOMATIC,
-  @SerializedName("Invoke") INVOKE,
+enum class AutocompleteTriggerKind(val value: String) {
+  AUTOMATIC("Automatic"),
+  INVOKE("Invoke"),
 }
 
 data class AutocompleteParams(
     val filePath: String,
     val position: Position,
-    val triggerKind: AutocompleteTriggerKind? = AutocompleteTriggerKind.AUTOMATIC,
+    val triggerKind: String? = AutocompleteTriggerKind.AUTOMATIC.value,
     val selectedCompletionInfo: SelectedCompletionInfo? = null
 )
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -12,10 +12,12 @@ import com.intellij.openapi.editor.InlayModel
 import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Disposer
+import com.intellij.openapi.util.TextRange
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.sourcegraph.cody.agent.CodyAgent.Companion.getServer
 import com.sourcegraph.cody.agent.CodyAgentManager.tryRestartingAgentIfNotRunning
 import com.sourcegraph.cody.agent.protocol.AutocompleteParams
+import com.sourcegraph.cody.agent.protocol.AutocompleteTriggerKind
 import com.sourcegraph.cody.agent.protocol.Position
 import com.sourcegraph.cody.agent.protocol.SelectedCompletionInfo
 import com.sourcegraph.cody.autocomplete.render.AutocompleteRendererType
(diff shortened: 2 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/AutocompleteParams.kt
```diff
@@ -1,15 +1,13 @@
...
+  INVOKE("Invoke"),
 }
 
 data class AutocompleteParams(
     val filePath: String,
     val position: Position,
-    val triggerKind: AutocompleteTriggerKind? = AutocompleteTriggerKind.AUTOMATIC,
+    val triggerKind: String? = AutocompleteTriggerKind.AUTOMATIC.value,
     val selectedCompletionInfo: SelectedCompletionInfo? = null
 )
```
Strum355, line 13: Are you sure `@SerializedName` doesnt work? I had tested it originally as part of https://github.com/sourcegraph/jetbrains/pull/88/ and it appeared to work fine :thinking: https://stackoverflow.com/questions/16740078/serialize-and-deserialize-enum-with-gson

Strum355: I created a sample project with the following code:
```kotlin
import com.google.gson.Gson
import com.google.gson.annotations.SerializedName

enum class AutocompleteTriggerKind {
    @SerializedName("Automatic") AUTOMATIC,
    @SerializedName("Invoke") INVOKE,
  }

data class User(val firstName: String, val lastName: String, val kind: AutocompleteTriggerKind) {
    val banana: String = ""
}

fun main() {
    val gson = Gson().toJson(User("Tom", "Broody", AutocompleteTriggerKind.AUTOMATIC))
    println(gson)
}
```
which yields the following output
`{"firstName":"Tom","lastName":"Broody","kind":"Automatic","banana":""}`

Sa1to: It isn't mapped correctly on the agent side. It was always converted to value `0` idk why

Sa1to: Screenshot 2023-11-15 at 15 56 45

https://github.com/sourcegraph/jetbrains/assets/9321940/c72c79c4-7988-49fb-b1a6-38a5f0b5f16f

Sa1to: This is on main branch

mkondratek:
It is clearly improved now.

I managed to find an issue with the "actions" that appear in the lookup.

image

Idk, what we should do about it. Do not autocomplete for them (the actions). If it is possible to find out and fix it easily in this PR. But probably we can ignore it for now.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/AutocompleteParams.kt
Strum355: Looks like this is the issue :facepalm: https://sourcegraph.com/github.com/eclipse-lsp4j/lsp4j/-/blob/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EnumTypeAdapter.java?L53-77

mkondratek: Hm, but the serialization on the client side seems to work fine.

I tested it with `SourcegraphApiContentHelper.toJson` (I believe it uses the object mapper that is actually used).

It still can be either an issue with the deserialization on the agent side or some issue that is specific to the logs
 only.

Strum355: lsp4j.jsonrpc uses a different Gson builder with different typadapters set afaik

Strum355 (approved):

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -184,6 +201,7 @@ class CodyAutocompleteManager {
       triggerKind: InlineCompletionTriggerKind,
       cancellationToken: CancellationToken,
       lookupString: String?,
```
mkondratek, line 203: I looked quickly into the code. 

It seems that it would be possible to disable autocompletion for some specific lookup elements. Now, we are depending directly on the list's selected values (strings). Instead, we could try to check for a precise `com.intellij.codeInsight.lookup.LookupElement` implementation and decide.

If think that would be a nice improvement, b/c the autocompletions for the actions look a bit dumb 😅 I guess we can provide it later. 

What we have here, LGTM.

mkondratek (approved):

PR #140, opened 2023-11-22
Migrate to autocomplete telemetry in-agent vs in-client

Strum355:
**TODO BEFORE MERGE:** bump cody version to a commit containing the merged https://github.com/sourcegraph/cody/pull/1911

---

To maximize the consistency in the type of events as well as the format of events sent between agent consumers (jetbrains, neovim etc) and vscode, we are migrating to emitting as many events as technically possible to the agent. 
This PR is the jetbrains side of things, the rest of the work has been happening over at https://github.com/sourcegraph/cody over the past week. 

What is changing:
- Autocomplete acceptance agent notifications are being sent AFTER document change events. This is due to Agent being architected in a vscode-centric fashion, so we have to try mirror the expectations of how vscode does things. This is [documented in-code here](https://github.com/sourcegraph/jetbrains/pull/140/files#diff-528abfae0db6381bdc68be35bb051d35e01139a40fc106d95b61da3e37d7764bR14-R16) and [here](https://github.com/sourcegraph/jetbrains/pull/140/files#diff-eeecc5eca46edd88b12020a0cad5f310c333d05c8dbf70ab674ca89c164881f6R100).
- A bunch of code was either translated to Kotlin, or else removed if no longer needed :broom: Most notably, completion accepting and event logging is significantly cut down

## Test plan

Much manual testing over the past week


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
```diff
@@ -0,0 +1,70 @@
...
+
+  @JsonNotification("debug/message") fun debugMessage(message: DebugMessage)
+
+  @JsonNotification("autocomplete/clearLastCandidate") fun autocompleteClearLastCandidate()
+
+  @JsonNotification("autocomplete/completionsSuggested")
+  fun completionsSuggested(items: CompletionItemID)
+
+  @JsonNotification("autocomplete/completionsAccepted")
+  fun completionsAccepted(logID: CompletionItemID)
```
Strum355: These two are the new ones, amidst the Kotlin move

src/main/kotlin/com/sourcegraph/cody/autocomplete/action/AcceptAutocompleteActionHandler.kt
```diff
@@ -57,8 +57,14 @@ class AcceptAutocompleteActionHandler : AutocompleteActionHandler() {
...
+
+      val id = telemetry.logID
+      if (id != null) {
+        server.completionsAccepted(id)
+      }
+
+      // GraphQlLogger.logAutocompleteAcceptedEvent(project, telemetry?.params())
       server.autocompleteClearLastCandidate()
       acceptAgentAutocomplete(editor, maybeCaret)
     } else {
```
Strum355: @sourcegraph/jetbrains-plugin does anyone have any insight into why this `else` case might ever be hit? Im not sure when the agent server would ever be null if we're receiving completions

olafurpg: The server may become null after receiving a completion, but I think it's an unlikely situation

src/main/kotlin/com/sourcegraph/cody/vscode/InlineAutocompleteItem.kt
```diff
@@ -1,8 +0,0 @@
-package com.sourcegraph.cody.vscode
-
-class InlineAutocompleteItem(
-    var insertText: String,
-    var filterText: String,
-    var range: Range,
-    var command: Command
```
mkondratek, line 7: Can we remove `com.sourcegraph.cody.vscode.Command` btw?

src/main/kotlin/com/sourcegraph/cody/autocomplete/action/AcceptAutocompleteActionHandler.kt
```diff
@@ -57,33 +20,17 @@ class AcceptAutocompleteActionHandler : AutocompleteActionHandler() {
...
-      val telemetry = CodyAutocompleteManager.instance.currentAutocompleteTelemetry
-      GraphQlLogger.logAutocompleteAcceptedEvent(project, telemetry?.params())
-      server.autocompleteClearLastCandidate()
-      acceptAgentAutocomplete(editor, maybeCaret)
-    } else {
       val caret = maybeCaret ?: getSingleCaret(editor) ?: return
-      AutocompleteText.atCaret(caret)?.let {
-        /* Log the event */
-        GraphQlLogger.logCodyEvent(project, "completion", "accepted")
-        WriteAction.run<RuntimeException> { applyAutocomplete(editor.document, it) }
```
mkondratek, line 69: we remove `applyAutocomplete` here, but this is just fine since IIUC we `applyInsertText` now (which does the the same thing - applies the autocompletion)

mkondratek (approved):
lgtm

src/main/kotlin/com/sourcegraph/cody/autocomplete/action/AcceptAutocompleteActionHandler.kt
Strum355: Correct, this case only exists in the scenario that the server shutsdown between the completion being returned and it being accepted. But we dont need the server to accept it except for telemetry purposes, so the case is imo exceptional enough that we dont need the extra code complexity of both cases

src/main/kotlin/com/sourcegraph/cody/vscode/InlineAutocompleteItem.kt
Strum355: If its no longer being used anywhere, it can definitely be removed (now or later) i think 🙂

exigow:
Green light.  The latest Cody bump contains https://github.com/sourcegraph/cody/pull/1911.

exigow:
Looks like notifications `completionAccepted` and `completionSuggested` are functioning as expected. 

Notification `didChange` is fired **after** `completionAccepted`.

PR #145, opened 2023-11-29
Migrate from filePaths to URIs with latest Cody

Strum355:
See https://github.com/sourcegraph/cody/commit/2b3c3b0a708d11f1f828130c37116315ed140468

## Test plan

Locally tested as part of daily development

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/AutocompleteParams.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/AutocompleteParams.kt
@@ -1,6 +1,7 @@
 package com.sourcegraph.cody.agent.protocol
 
 import java.net.URI
+import java.nio.file.Paths
 
 enum class AutocompleteTriggerKind(val value: String) {
   AUTOMATIC("Automatic"),
@@ -18,5 +19,5 @@ data class AutocompleteParams(
       position: Position,
       triggerKind: String? = AutocompleteTriggerKind.AUTOMATIC.value,
       selectedCompletionInfo: SelectedCompletionInfo? = null
-  ) : this(URI("file://$filePath"), position, triggerKind, selectedCompletionInfo)
+  ) : this(Paths.get(filePath).toUri(), position, triggerKind, selectedCompletionInfo)
 }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/AutocompleteParams.kt
```diff
@@ -1,13 +1,22 @@
...
     val triggerKind: String? = AutocompleteTriggerKind.AUTOMATIC.value,
     val selectedCompletionInfo: SelectedCompletionInfo? = null
-)
+) {
+  constructor(
+      filePath: String,
+      position: Position,
+      triggerKind: String? = AutocompleteTriggerKind.AUTOMATIC.value,
+      selectedCompletionInfo: SelectedCompletionInfo? = null
+  ) : this(URI("file://$filePath"), position, triggerKind, selectedCompletionInfo)
```
olafurpg: This URI constructor is going to crash in windows. It's better to go via nio.file.Path

Strum355: Do you have a source on this? The only historical info I can find on windows path issues was on the Cody side of things and I couldnt find anything on google

olafurpg: We already had the same issue before, which got fixed here https://sourcegraph.com/github.com/sourcegraph/jetbrains/-/commit/ac233c5959d5aff51ca4148a8d7cc5f883ac3581?visible=2

olafurpg: The problem was space characters based on that commit

Strum355: Ah yes I remember now! Couldn't find it in slack or gh issue/PR search 🤔

olafurpg: `type:diff` in Sourcegraph found 😜

Strum355: Done https://github.com/sourcegraph/jetbrains/pull/145/commits/9e2540ca0c88d35768e508a319d7f11f0432dfee

olafurpg (approved):

PR #163, opened 2023-12-05
Fix text wrapping and messages 

mkondratek:
An important cody commit update included!
(https://github.com/sourcegraph/cody/pull/2014)

## Test plan
- `runIde`
- verify autocomplete works
- verify chat works
- verify commands work; the error text should be wrapped


mkondratek:
image

olafurpg (approved):

PR #166, opened 2023-12-06
Add "Upgrade" option to RLE notifications

mkondratek:
## Overview
1. This PR adds `systemProperty("cody.isGa", "true")` to `build.gradle`. We may need to set it to false before the next stable release.
2. Messages related to the RateLimitError have been adjusted in the chat, notifications & status warning dialogs.
3. `Upgrade` option added when applicable 

## Test plan
- play with rate limit exceeded accounts

## Enable `cody.isGa=true` for testing
1. Install the plugin with these changes 
2. Go to `Help` -> `Edit Custom VM options....`
3. Add `-Dcody.isGa=true`
4. Restart IntelliJ (you can `File` -> `Invalidate Caches...` -> `Invalidate and restart`)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -337,6 +337,7 @@ tasks {
     systemProperty(
         "cody.autocomplete.enableFormatting",
         project.property("cody.autocomplete.enableFormatting") ?: "true")
+    systemProperty("cody.isGa", "true")
   }
 
   runPluginVerifier {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -205,7 +205,7 @@ tasks {
     return destinationDir
   }
 
-  val codyCommit = "29e84d968a5eec6b87bedb1658a80a6578bfe86e"
+  val codyCommit = "edd8e27d2eec2e415d730cf3de96db4dad1cf38d"
   fun downloadCody(): File {
     val url = "https://github.com/sourcegraph/cody/archive/$codyCommit.zip"
     val destination = githubArchiveCache.resolve("$codyCommit.zip")
--- a/src/main/kotlin/com/sourcegraph/cody/config/CodyAccountListModel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/CodyAccountListModel.kt
@@ -9,6 +9,8 @@ import com.intellij.util.containers.orNull
 import com.sourcegraph.cody.auth.ui.AccountsListModel
 import com.sourcegraph.cody.auth.ui.AccountsListModelBase
 import com.sourcegraph.cody.localapp.LocalAppManager
+import com.sourcegraph.cody.statusbar.CodyAutocompleteStatusService
+import com.sourcegraph.common.UpgradeToCodyProNotification
 import javax.swing.JComponent
 
 class CodyAccountListModel(private val project: Project) :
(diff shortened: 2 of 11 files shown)


mkondratek:
| Chat & Free account | Chat & Pro account |  
|---|---|
| image | image | 

| Autocompletion & Free account | Autocompletion & Pro account |  
|---|---|
| image | image |

| Status Warning & Free account | Status Warning & Pro account |  
|---|---|
| image | image |
| image | image |
| image | image |

src/main/kotlin/com/sourcegraph/cody/statusbar/CodyStatusBarActionGroup.kt
```diff
@@ -34,27 +34,52 @@ class CodyStatusBarActionGroup : DefaultActionGroup() {
...
+  private fun deriveWarningAction(): RateLimitErrorWarningAction? {
+    val autocompleteRLE = UpgradeToCodyProNotification.autocompleteRateLimitError.get()
+    val chatRLE = UpgradeToCodyProNotification.chatRateLimitError.get()
+
+    // TODO(mikolaj):
+    // RFC 872 mentions `feature flag cody-pro: true`
+    // the flag should be a factor in whether to show the upgrade option
+    val isGa = java.lang.Boolean.getBoolean("cody.isGa")
+    val shouldShowUpgradeOption =
+        isGa && autocompleteRLE?.upgradeIsAvailable ?: chatRLE?.upgradeIsAvailable ?: false
```
mkondratek, line 46: We may need to add a dependency on the mentioned flag.

src/main/kotlin/com/sourcegraph/cody/config/CodyAccountListModel.kt
```diff
@@ -78,6 +80,9 @@ class CodyAccountListModel(private val project: Project) :
     accountsListModel.add(account)
     newCredentials[account] = token
     notifyCredentialsChanged(account)
+    UpgradeToCodyProNotification.autocompleteRateLimitError.set(null)
+    UpgradeToCodyProNotification.chatRateLimitError.set(null)
+    CodyAutocompleteStatusService.resetApplication(project)
```
mkondratek, line 85: btw the status is more up to date now

build.gradle.kts
```diff
@@ -337,6 +337,7 @@ tasks {
     systemProperty(
         "cody.autocomplete.enableFormatting",
         project.property("cody.autocomplete.enableFormatting") ?: "true")
+    systemProperty("cody.isGa", "true")
```
mkondratek, line 340: before the next stable we may need to turn it off

danielmarquespt:
Looks good. Can you change the order of the buttons so that its [Upgrade] [Close]?

Sa1to (approved):
LGTM

exigow (commented):
> This PR adds systemProperty("cody.isGa", "true") to build.gradle. We may need to set it to false before the next stable release.

AFAIK we have to remember about this flag also for nightly builds, not only stable.

exigow (commented):
image

I've received same notification during autocompletion two times in short period of time. However, this seems to be minor issue.

exigow (changes requested):
After clicking on a notification in the notification list, the links are deactivated. Is this correct behavior? For example, in this GIF, I click 'Learn more' after which I would like to click 'Upgrade' but I am unable to.

freeze

Just letting you know. This is a minor issue because notifications shown as toast are automatically hidden (after clicking chosen action), though there might be a case where someone enters the notification list. We can ignore this or fix that later after merge.

mkondratek:
@exigow thanks for testing. I fixed the link availability. Now, I just hide the balloon instead of expiring the notification.

You should not be able to get more than one autocomplete RLE notifications from automatically invoked autocompletion. I never experienced that. I guess we can ignore it for now.

exigow (approved):
Thanks for the quick fix! 

I've played around with it a bit and it now looks fine. 

What I miss the most are notification tests, especially since the notifications seems very isolated from the rest, and such tests could be written based on input and expecting the right HTML content. I assume we don't have time for this now, but I think we can look into it later. Maybe I'll try to write something like this, especially since we want to add notification support in the 'Subscription' panel which will share the same state.

Issue #149, opened 2023-11-30
PLG: Add "Upgrade to Cody Pro" and "Current Usage" settings buttons


exigow:
We want to enable users to see the current number of available Chat/Code queries **without** exceeding the rate limit.

This is scenario when user downloads the plugin from the marketplace and once is signed in, he should be able to check current usage without interacting with chat (theoretically).

* Usage button should navigate to [sourcegraph.com/cody/manage](http://sourcegraph.com/manage).
* Upgrade button should navigate to [sourcegraph.com/cody/subscription](http://sourcegraph.com/cody/subscription).

Important note: PLG features cannot be published before GA, so it must be controlled by the cody-pro feature flag.

exigow:
~@danielmarquespt, do you have some ideas on how we should show this buttons? I suggested placing these buttons in the settings, but I'm not sure if that's a good solution. Can you please verify this?~

EDIT: Design: https://github.com/sourcegraph/jetbrains/issues/156

exigow:
This issue is blocked by https://github.com/sourcegraph/jetbrains/issues/147.

exigow:
Clarification: Pages are blank because only @sourcegraph.com email addresses have access to the `/cody/manage` and `/cody/subscription`.

mkondratek:
1. Does it mean the upgrade and current usage options SHOULD NOT be available to everyone (but to the users with `@sourcegraph.com` only)?
2. How can I tell whether I am `@sourcegraph.com` or not?

PR #171, opened 2023-12-07
Add subscription tab

Sa1to:
It fixes [PLG: Add "Upgrade to Cody Pro" and "Current Usage" settings buttons#149]
(https://github.com/sourcegraph/jetbrains/issues/149)
We need to merge https://github.com/sourcegraph/cody/pull/2274 first
## Test plan
* Log in as Cody Free user
* Check if subscription tab available
* Check if you see correct tier
* Check if you see `Upgrade` and `Check Usage` buttons that are redirecting to `https://sourcegraph.com/cody/subscription` and `https://sourcegraph.com/cody/manage`
* Log in as Cody Pro user
* Check if subscription tab available
* Check if you see correct tier
* Check if you see only `Check Usage` buttons that are redirecting to `https://sourcegraph.com/cody/subscription` and `https://sourcegraph.com/cody/manage`

https://github.com/sourcegraph/jetbrains/assets/9321940/79ccf53d-0685-4b41-815e-511a0beb5f22


src/main/kotlin/com/sourcegraph/cody/config/test.form
mkondratek, line 1: I guess we can drop this file

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -52,8 +53,9 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
   private val sendButton: JButton
   private var inProgressChat = CancellationToken()
   private val stopGeneratingButton =
-      JButton("Stop generating", IconUtil.desaturate(AllIcons.Actions.Suspend))
+    JButton("Stop generating", IconUtil.desaturate(AllIcons.Actions.Suspend))
   private val recipesPanel: JBPanelWithEmptyText
+  private val subscriptionPanel: JPanel
```
mkondratek: This file is already a mess. If there is a chance to add anything in a separate file in the first place we should go for it.

vdavid:
Behavior-wise, 👍  from me as PLG team member. (I'll let Cody JetBrains teammates check the code.)

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -111,6 +113,38 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
     addWelcomeMessage()
   }
 
+  @RequiresEdt
+  private fun addSubscriptionTab() {
+    tryRestartingAgentIfNotRunning(project)
+    val server = getServer(project)
+    if (server != null) {
+      val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyPro"))
```
mkondratek: tbd if we should limit to sourcegraph.com instance only

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -226,7 +226,11 @@ class CodyAutocompleteManager {
                             com.sourcegraph.cody.vscode.Position(lineNumber, startPosition),
                             position)))
     notifyApplication(CodyAutocompleteStatus.AutocompleteInProgress)
-    val completions = server.autocompleteExecute(params)
+    //    runBlocking {
```
mkondratek: nit: I guess the comment can be dropped

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -111,6 +113,41 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
   }
 
+  @RequiresEdt
+  private fun addSubscriptionTab() {
+    val activeAccountType = CodyAuthenticationManager.instance.getActiveAccount(project)
+    if (activeAccountType != null && activeAccountType.isDotcomAccount()) {
+      tryRestartingAgentIfNotRunning(project)
+      val server = getServer(project)
+      if (server != null) {
+        val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyPro"))
```
mkondratek: we should depend on our flag now `cody-pro-jetbrains`

exigow: @mkondratek, can you elaborate and/or suggest a change? Szymon finished early today, so he won't be able to update this till tomorrow. 

Does this boil down to changing from `val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyPro"))` to `val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("cody-pro-jetbrains"))`?

Sa1to: So should we rely **only** on our internal feature flag or should we have both (one form agent and our internal)?

exigow (changes requested):
During tests, my IDE freezes and the last log is: `2023-12-11 18:27:13,812 [  37588]   WARN - #c.s.c.a.CodyAgent - No handler for request with method featureFlags/getFeatureFlag`

Sa1to:
@exigow I have to rebase on master, it requires latest changes in agent 😅

exigow (approved):
Thanks for the fixes! :+1: 

Issue seems to be fixed with latest bump, but there is still an problem with the short IDE freezing (for less than a second but it's visible). Apparently, we have some code that happens in the AWT thread and it blocks the entire UI. 

We can ignore this problem for now.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -111,6 +113,43 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
+  private fun addSubscriptionTab() {
+    val activeAccountType = CodyAuthenticationManager.instance.getActiveAccount(project)
+    if (activeAccountType != null && activeAccountType.isDotcomAccount()) {
+      tryRestartingAgentIfNotRunning(project)
+      val server = getServer(project)
+      if (server != null) {
+        val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyProJetBrains"))
+        if (codyProFeatureFlag.get() != null &&
+            codyProFeatureFlag.get()!! &&
+            java.lang.Boolean.getBoolean("cody.isGa")) {
```
mkondratek: As we decided during the meetings. Let's drop `cody.isGa` condition here. We will request ff exceptions for ones that want to test these features.

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
```diff
@@ -0,0 +1,22 @@
...
+  row {
+    if (!isCurrentUserPro) {
+      val upgradeButton =
+          button("Upgrade") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/subscription") }
+      upgradeButton.component.putClientProperty(DarculaButtonUI.DEFAULT_STYLE_KEY, true)
+    }
+    button("Check Usage") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/manage") }
+  }
+  if (!isCurrentUserPro) {
+    row { label("Please refresh IDE after upgrading to Cody Pro") }
```
mkondratek: pls use the copy from: https://github.com/sourcegraph/jetbrains/issues/178

mkondratek (changes requested):
pls apply the two fixes I mentioned. 

There is one more nice to have - but we can deliver it later:
We could show the sub tab depending on the account type when it changes. For now when I change to the account that is eligible to see sub tab the tab does not appear. I assume you can add some logic there:
https://sourcegraph.com/github.com/sourcegraph/jetbrains/-/blob/src/main/kotlin/com/sourcegraph/cody/config/CodyPersistentAccountsHost.kt?L29%3A1-30%3A1

mkondratek:
I created this: https://github.com/sourcegraph/jetbrains/issues/180

mkondratek (approved):
image


this is good 🚀

PR #172, opened 2023-12-08
Increase getInitializedServer timeout in CodyFileEditorListener

### LINKED ISSUE #169: Investigate and fix the issue with chat hanging

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/169.

## Test plan
- see the related issue for full context
- `runIde` and check if the chat works properly when there is a file opened in the editor on startup
- check sourcegraph links too

### DIFF
--- a/src/main/java/com/sourcegraph/cody/CodyFileEditorListener.java
+++ b/src/main/java/com/sourcegraph/cody/CodyFileEditorListener.java
@@ -25,6 +25,9 @@ public class CodyFileEditorListener implements FileEditorManagerListener {
     }
 
     CodyAgent.getInitializedServer(source.getProject())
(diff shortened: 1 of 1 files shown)


mkondratek:
We should have tried it earlier but anyway...

We were trying to understand what was going on and how com.sourcegraph.cody.agent.CodyAgent#initialize and its field initialized work. TBH I still don't get it. It is likely that the class has grown some tech debt over time making it harder to grasp.

My point is that my PR is not a real fix (from development POV) but it should be enough for now for users. A real fix would be to rewrite the logic and make it cleaner.

exigow (approved):
I approve this workaround.

PR #179, opened 2023-12-11
Update the upsell copies

### LINKED ISSUE #178: Add copy to all Cody Free tier upsells about refreshing their IDE
### Description:

Requirements:
- Append the following text to "Status completion" and "Autocomplete" Free tier upsells (see https://github.com/sourcegraph/jetbrains/pull/166):
  - "(Already upgraded to Pro? Restart your IDE for changes to take effect)"
- For "Chat" upsells, append the above copy _after_ the "Upgrade or learn more" text. 

### Design Tasks:

```[tasklist]
### Design Tasks
- [ ] Add a draft title or issue reference here
```

### Engineering Tasks:

```[tasklist]
### Engineering Tasks
- [ ] Add a draft title or issue reference here
```

### Related Docs:

- [PRD: Your-Title]()

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/178.

## Test plan
- play with runIde and the accounts with the rate limit errors

### DIFF
--- a/src/main/kotlin/com/sourcegraph/common/CodyBundle.kt
+++ b/src/main/kotlin/com/sourcegraph/common/CodyBundle.kt
@@ -1,0 +1,31 @@
+package com.sourcegraph.common
+
+import java.text.MessageFormat
+import java.util.*
+import org.jetbrains.annotations.PropertyKey
+
+object CodyBundle {
+  private const val BUNDLE: String = "CodyBundle"
+  private val instance: ResourceBundle = ResourceBundle.getBundle(BUNDLE)
+
+  /**
+   * A more restrictive version of [MessageFormat.format]. Since each parameter must be a non-null
+   * [String], we can capture the unintended parameter types (like `io.vavr.control.Option`) more
+   * easily during the build (this is realized with ArchUnit; see the test against `Option#toString`
+   * in the top-level project). Note that we consciously use the name "fmt" instead of "format" to
+   * avoid an accidental use of [String.format] and emphasize the need to use the
+   * [lombok.experimental.ExtensionMethod] annotation.
+   *
+   * @param this@fmt as in [MessageFormat.format]
+   * @param args as in [MessageFormat.format], but each parameter must be a non-null [String] and
+   *   not just a nullable [Object]
+   * @return the formatted string
+   */
+  fun String.fmt(vararg args: String): String {
+    return MessageFormat.format(this, *args as Array<Any>)
+  }
+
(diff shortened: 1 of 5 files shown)


mkondratek:
## Error in the chat:

image

## Error from autocompletion:

image


## Warning dialog (from the status widget)

image

chenkc805 (approved):

src/main/kotlin/com/sourcegraph/common/CodyBundle.kt
```diff
@@ -0,0 +1,31 @@
...
+   * @return the formatted string
+   */
+  fun String.fmt(vararg args: String): String {
+    return MessageFormat.format(this, *args as Array<Any>)
+  }
+
+  fun getString(@PropertyKey(resourceBundle = BUNDLE) key: String): String {
+    return instance.getString(key)
+  }
+}
```
exigow, line 31: What's the motivation for adding this `CodyBundle`? 

I think the previous version was clearer. This multiline comment describes situations that are not related to our stack (Vavr and Lombok which are for Java, and ArchUnit which we don't use). I think there's no point in adding such additional abstraction. In the repository, 95% of the text is still hardcoded so we are introducing a second way of text management. If we want to do such bundles, it shouldn't be part of these small fixes and should be a topic for a larger discussion. 

The initiative of introducing this is very good :+1: , but we gonna have two different solutions, and IMO it's more important to have consistency.

Nevertheless, I **approve** these changes because we need to include them in today's release.

exigow (approved):
Looks good :+1: , however I've added one comment regarding `CodyBundle`.

src/main/kotlin/com/sourcegraph/common/CodyBundle.kt
olafurpg: +1 to Jakub's comment

mkondratek: My motivation was that the messages look much cleaner in the bundle esp when using html tags. But I won't insist at this point.

mkondratek: I vote for consistency too. At the moment I have no time to move all messages to the bundle 😅

olafurpg: Fine to merge as is and make everything consistent later (either via bundle or plain strings)

chenkc805:
Thank you JetBrains team! Looks great

PR #183, opened 2023-12-12
Add usage limit notification in subscription tab

### LINKED ISSUE #147: PLG: Add ability to check account Pro status and current quota / rate limits
We need the following queries to start working on PLG features: 
* Is Cody Pro enabled?
* What's the current Chat usage and what's the limit?
* What's the current Code usage and what's the limit?

More details: https://docs.google.com/document/d/1qYDsO6GOxGHEB2ZRSO9BWgjTra2adw0y3BDTNkjZqZw/edit

Sa1to:
It fixes https://github.com/sourcegraph/jetbrains/issues/147

## Test plan
* Login as user with exceeded rate limit
* Type a chat message
* Check if chat notification is present in subscription tab
* Trigger autocomplete
* Check if notification for chat and autocomplete is present in subscription tab
* Change user
* Again login as user with exceeded rate limit
* Trigger autocomplete 
* Check if autocomplete notification is present in subscription tab

https://github.com/sourcegraph/jetbrains/assets/9321940/78c0afed-8bf2-4ce7-b181-6fdf1ab24bf8

https://github.com/sourcegraph/jetbrains/assets/9321940/632ca94f-32aa-4f3c-98a9-01f682998362


mkondratek:
could you please rebase this PR & add new demos? 🙏 for this one screenshots work too (and they are faster to produce 😅). Additionally, the rate-limited accounts are back so you can test against both the pro and free tier.

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
```diff
@@ -2,10 +2,33 @@ package com.sourcegraph.cody
...
+import com.sourcegraph.common.UpgradeToCodyProNotification
 import com.sourcegraph.config.ConfigUtil
 
 fun createSubscriptionTab(isCurrentUserPro: Boolean) = panel {
+  val chatLimitError = UpgradeToCodyProNotification.chatRateLimitError.get()
+  val autocompleteLimitError = UpgradeToCodyProNotification.autocompleteRateLimitError.get()
+  if (chatLimitError != null && autocompleteLimitError != null) {
+    row {
+      label(
+          "<html>\u26A1 You used all you autocompletions and chats for this month. Upgrade to Cody Pro to get unlimited interactions.</html>")
```
mkondratek: we must distinguish pro vs free - it is silly to ask someone who already has pro to upgrade 😅

mkondratek (changes requested):

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
mkondratek: For a pro user you may want to use the copy mentioned there: https://github.com/sourcegraph/jetbrains/issues/194.

pkukielka: Agree. And since Mikolaj already added those messages in the text bundle please use it instead of pasting the text directly.

pkukielka: On the second look, I do not think the same message will fit into subscription tab. 
Still, we should put those messages in the bundle.

pkukielka: And last one: is it intentional that in the first case we do not provide exact info about the limits but we do if only one of the limits is reached?

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -131,9 +131,21 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
                   }
                   .get()
           if (isCurrentUserPro != null) {
-            val subscriptionPanel = createSubscriptionTab(isCurrentUserPro)
             tabbedPane.insertTab(
-                "Subscription", null, subscriptionPanel, null, SUBSCRIPTION_TAB_INDEX)
+                "Subscription",
```
pkukielka: You will get conflict with your other PR but you surely knows that...

pkukielka (approved):
LGTM

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
```diff
@@ -2,10 +2,38 @@ package com.sourcegraph.cody
...
+import com.intellij.ui.dsl.builder.BottomGap
+import com.intellij.ui.dsl.builder.TopGap
 import com.intellij.ui.dsl.builder.panel
+import com.sourcegraph.common.UpgradeToCodyProNotification
 import com.sourcegraph.config.ConfigUtil
 
 fun createSubscriptionTab(isCurrentUserPro: Boolean) = panel {
+  val chatLimitError = UpgradeToCodyProNotification.chatRateLimitError.get()
+  val autocompleteLimitError = UpgradeToCodyProNotification.autocompleteRateLimitError.get()
+  if (!isCurrentUserPro) {
```
pkukielka: So the layout is the same in all 3 options and only difference is text?
If that is the case could you please refactor it so only changing element (label) is determined by ifs chain, and the layout is always the same and just uses proper label text?

pkukielka: Actually I just noticed that cell separator component in one of the cases.
What is that needed for?

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -161,9 +161,13 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
           if (isCurrentUserPro != null) {
-            val subscriptionPanel = createSubscriptionTab(isCurrentUserPro)
-            tabbedPane.insertTab(
-                "Subscription", null, subscriptionPanel, null, SUBSCRIPTION_TAB_INDEX)
+            try {
+              val subscriptionPanel = createSubscriptionTab(isCurrentUserPro)
+              tabbedPane.insertTab(
+                  "Subscription", null, subscriptionPanel, null, SUBSCRIPTION_TAB_INDEX)
+            } catch (e: Exception) {
+              println(e.message)
```
pkukielka: It should not be a `println`, please use a logger.

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
Sa1to: Those labels are implemented according to figma design https://www.figma.com/file/m9Yvgv6W3aPRLfXnxgwnEz/JetBrains?type=design&node-id=215-2737&mode=design&t=ZUILhBlencqhC8Fk-0

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
```diff
@@ -4,11 +4,52 @@ import com.intellij.ide.BrowserUtil
...
 import com.sourcegraph.common.CodyBundle
+import com.sourcegraph.common.UpgradeToCodyProNotification
 import com.sourcegraph.config.ConfigUtil
+import com.sourcegraph.config.ThemeUtil
 
 fun createSubscriptionTab(isCurrentUserPro: Boolean) = panel {
+  val chatLimitError = UpgradeToCodyProNotification.chatRateLimitError.get()
+  val autocompleteLimitError = UpgradeToCodyProNotification.autocompleteRateLimitError.get()
+  val chatLimitLabelValue =
+      "You used all ${chatLimitError?.limit} chat messages for this month. Upgrade to Cody Pro to get unlimited chats."
```
mkondratek: are we sure we want to display that value here? I am asking in the context of https://github.com/sourcegraph/jetbrains/pull/274

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
```diff
@@ -4,11 +4,52 @@ import com.intellij.ide.BrowserUtil
...
+  val autocompleteLimitLabelValue =
+      "You used all ${autocompleteLimitError?.limit} autocompletions for this month. Upgrade to Cody Pro to get unlimited autocompletions."
```
mkondratek: btw, pls move strings to `CodyBundle.properties`

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
```diff
@@ -4,11 +4,52 @@ import com.intellij.ide.BrowserUtil
...
+                    autocompleteLimitLabelValue
+                  }
+                }
+                }</p></td>" +
+              "</tr>" +
+              "</table>" +
+              "</html>")
+    }
+    row {
+      if (ThemeUtil.isDarkTheme()) {
```
pkukielka: What is you switch theme when this tab is visible? 
Will it switch colours properly?

src/main/kotlin/com/sourcegraph/cody/config/RateLimitStateManager.kt
```diff
@@ -18,6 +19,7 @@ object RateLimitStateManager {
     if (UpgradeToCodyProNotification.chatRateLimitError.get() == null) {
       UpgradeToCodyProNotification.chatRateLimitError.set(rateLimitError)
       CodyAutocompleteStatusService.resetApplication(project)
+      CodyToolWindowContent.getInstance(project).refreshSubscriptionTab()
```
mkondratek: adding this method here makes `reportForChat` `@RequiresBackgroundThread`. pls annotate it and ensure that it's called in the proper thread.

mkondratek (changes requested):

src/main/kotlin/com/sourcegraph/cody/config/RateLimitStateManager.kt
mkondratek: you can add `Thread.sleep(10000)` and see if the UI remains functional

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
Sa1to: Valid issue, i tested it before and it doesnt change the colour properly but i am hesitating to do the rerendering of the whole tab...

pkukielka: In fact I would double check if it's really needed. I thought html rendering css is adjusted automatically based on theme.

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
Sa1to: I guess we need to wait for reply from @taylorsperry

mkondratek (changes requested):
- [x] Subscription tab disappears on **RLE** on a **pro account** and **free account** (autocomplete):
1. `:runIde` from these changes
2. switch to a PRO/FREE, rate-limited account 
3. trigger autocompletion 
Actual:
Sub tab disappears and reappears (it looks flaky, unstable)
Expected: 
Sub tab just updates its content 

- [ ] No notification for a **pro account**:
In all other places we provide some handling even for the pro rate-limited accounts (however, this scenario seems unreal). 

 - [x] Subscription tab disappears on **RLE** on a **pro account** and **free account** (chat):
1. `:runIde` from these changes
2. switch to a PRO/FREE, rate-limited account 
3. trigger autocompletion 
Actual:
Sub tab disappears and reappears (it looks flaky, unstable)
Expected: 
Sub tab just updates its content

src/main/java/com/sourcegraph/cody/SubscriptionPanel.kt
Sa1to: I checked and it doesnt work automatically

src/main/resources/CodyBundle.properties
```diff
@@ -52,6 +52,10 @@ chat.rate-limit-error.explain=\
     <a href=\"https://sourcegraph.com/docs/cody/core-concepts/cody-gateway#rate-limits-and-quotas\">Learn more.</a>\
   </html>
 
+subscription-tab.chat-rate-limit-error=You've used all your chat messages and commands for the month. Upgrade to Pro for unlimited usage.
```
mkondratek, line 55: The overall experience is much better now without the flickering effect 🎉 

I have two points:
- [ ] No notification for a **pro account**:
In all other places we provide some handling even for the pro rate-limited accounts (however, this scenario seems unreal). 

- [ ] When I run a recipe the message does not appear.

src/main/kotlin/com/sourcegraph/cody/statusbar/FreeTrialEndNotification.kt
```diff
@@ -0,0 +1,43 @@
+package com.sourcegraph.cody.statusbar
+
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.project.DumbAwareAction
+import com.intellij.openapi.ui.Messages
+import com.sourcegraph.cody.Icons
+import com.sourcegraph.common.BrowserOpener
+import com.sourcegraph.config.ConfigUtil
+
+class FreeTrialEndNotification(
```
mkondratek: whoops, I guess we do not want to include this one here

src/main/resources/CodyBundle.properties
Sa1to: I think we won't show the notification for pro user accounts. Doesn't make sense since pro has unlimited access...

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -120,22 +122,39 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
       loadNewChatId()
     }
+    ApplicationManager.getApplication()
+        .messageBus
+        .connect()
+        .subscribe(
+            LafManagerListener.TOPIC,
+            LafManagerListener {
+              ApplicationManager.getApplication().executeOnPooledThread { refreshSubscriptionTab() }
+            })
```
mkondratek: is there any reason we subscribe from CodyToolWindowContent? can we move it directly to the SubscriptionTabPanel? Let's try to remove code from this class instead of adding it 😅

src/main/resources/CodyBundle.properties
mkondratek: but we have a testing account that is pro & unlimited so it's easy to test + we handle that case everywhere else

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -120,22 +122,39 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
-        addNewSubscriptionTab(agent.server)
       }
     }
   }
 
+  @RequiresEdt
+  private fun refreshSubscriptionTab(data: SubscriptionTabPanelData) {
+    if (data.isDotcomAccount && data.codyProFeatureFlag) {
+      if (tabbedPane.tabCount < SUBSCRIPTION_TAB_INDEX + 1) {
+        tabbedPane.insertTab("Subscription", null, subscriptionPanel, null, SUBSCRIPTION_TAB_INDEX)
```
mkondratek: pls name the `null` params

src/main/resources/CodyBundle.properties
mkondratek, line 55: One more thing:
- [ ] the following scenario:
1. chat message
2. rate limit error
3. notification shows in the subscription panel
4. in the meantime the limit resets (or the quota is increased) - so the account is no longer rate limited 
5. chat message
6. the error notification is invalidated in the widget but not in the subscription panel 

😕

mkondratek: likely all you need is to add 
```
      ApplicationManager.getApplication().executeOnPooledThread {
        CodyToolWindowContent.getInstance(project).refreshSubscriptionTab()
      }
```
to com.sourcegraph.cody.config.RateLimitStateManager#invalidateForChat

mkondratek: I am not sure about the autocompletions - pls verify the similar scenario for them

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
Sa1to: `[NAMED_ARGUMENTS_NOT_ALLOWED] Named arguments are not allowed for non-Kotlin functions`

mkondratek: you can `/* my_name= */ null`

src/main/java/com/sourcegraph/cody/SubscriptionTabPanel.kt
```diff
@@ -0,0 +1,106 @@
...
+      add(createRateLimitPanel(), BorderLayout.PAGE_START)
+    }
+    add(createCenterPanel())
+    ApplicationManager.getApplication()
+        .messageBus
+        .connect()
+        .subscribe(
+            LafManagerListener.TOPIC,
+            LafManagerListener {
+              ApplicationManager.getApplication().executeOnPooledThread { update(isCurrentUserPro) }
```
mkondratek: there is no need to call it on the pooled thread + annotate `update` method as `@RequiresEdt`

src/main/java/com/sourcegraph/cody/SubscriptionTabPanel.kt
```diff
@@ -0,0 +1,102 @@
...
+        if (ThemeUtil.isDarkTheme()) {
+          text("<div style=\"height: 1px; background-color: #404245;\"></div>")
+        } else {
+          text("<div style=\"height: 1px; background-color: #ECEDF1;\"></div>")
+        }
+      }
+    }
+  }
+
+  // todo: move strings to the bundle
```
pkukielka: TODO not needed?

src/main/java/com/sourcegraph/cody/SubscriptionTabUtils.kt
```diff
@@ -0,0 +1,81 @@
...
+@RequiresBackgroundThread
+fun fetchSubscriptionPanelData(
+    project: Project,
+    server: CodyAgentServer
+): CompletableFuture<SubscriptionTabPanelData?> {
+  val activeAccountType = CodyAuthenticationManager.instance.getActiveAccount(project)
+  if (activeAccountType != null) {
+    ensureUserIdMatchInAgent(activeAccountType.id, server)
+    return if (activeAccountType.isDotcomAccount()) {
+      val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyProJetBrains"))
```
pkukielka, line 25: Is that doing something else than `  @JsonRequest("graphql/currentUserIsPro") fun isCurrentUserPro(): CompletableFuture<Boolean?>` method we are using across the codebase?

src/main/java/com/sourcegraph/cody/SubscriptionTabPanel.kt
```diff
@@ -0,0 +1,101 @@
...
+      row {
+        if (ThemeUtil.isDarkTheme()) {
+          text("<div style=\"height: 1px; background-color: #404245;\"></div>")
+        } else {
+          text("<div style=\"height: 1px; background-color: #ECEDF1;\"></div>")
+        }
+      }
+    }
+  }
+
```
mkondratek, line 69: there are still a few strings that we could bundle:
```
val tier =
        if (getIsCurrentUserPro == null) "Loading..."
        else if (getIsCurrentUserPro) "Cody Pro" else "Cody Free"
    row { label("<html>Current tier: <b>$tier</b><html/>") }
```

pls either bundle them or leave the todo comment

src/main/java/com/sourcegraph/cody/SubscriptionTabPanel.kt
```diff
@@ -0,0 +1,101 @@
...
+  private fun createCenterPanel() = panel {
+    val getIsCurrentUserPro = isCurrentUserPro
```
pkukielka, line 71: Why you need this assigment?

mkondratek (commented):
This scenario:

1. rate limit account, no error in the sub tab
2. trigger autocomplete
3. rate limit error in the sub tab ✅ 
4. run any command
5. the rate limit error does not updates (does not reference chats)

we may not want to fix it since we want to move to commands (from recipes api) anyway

src/main/java/com/sourcegraph/cody/SubscriptionTabPanel.kt
mkondratek: so the values does not change in a meantime I guess

Sa1to: I am using it in `createCenterPanel` function and in `LafManagerListener`. It changes whenever we are fetching new `SubscriptionTabPanelData`

src/main/java/com/sourcegraph/cody/SubscriptionTabUtils.kt
Sa1to: Yes, this is feature flag that was used to disable/enable whole PLG (subscription tab) feature.

exigow (approved):
Looks good :+1:

mkondratek (approved):
lgtm

PR #187, opened 2023-12-12
Add testing steps for PLG features

Gedochao:
## Test plan

Go through the testing steps and make sure they cover all the scenarios we need testing. We do not need to cover all the edge cases here, just give a general idea of the happy path.


exigow (approved):
:+1:

I clicked through these scenarios and simultaneously made a few changes (I moved it to a separate PLG section, fixed the button names, changed some sentences to the imperative mode, and made a few smaller fixes).

PR #190, opened 2023-12-12
Refresh subscription tab on accounts changes

### LINKED ISSUE #180: Refresh the subscription tab on the account change
> There is one more nice to have - but we can deliver it later:
> We could show the sub tab depending on the account type when it changes. For now when I change to the account that is eligible to see sub tab the tab does not appear. I assume you can add some logic there:
> https://sourcegraph.com/github.com/sourcegraph/jetbrains/-/blob/src/main/kotlin/com/sourcegraph/cody/config/CodyPersistentAccountsHost.kt?L29%3A1-30%3A1

_Originally posted by @mkondratek in https://github.com/sourcegraph/jetbrains/pull/171#pullrequestreview-1777050489_

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/180.

## Test plan 
- play with pro/free accounts and see if the changes are reflected in the subscription tab
-

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/config/ui/AccountConfigurable.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/ui/AccountConfigurable.kt
@@ -2,6 +2,7 @@ package com.sourcegraph.cody.config.ui
 
 import com.intellij.collaboration.util.ProgressIndicatorsProvider
 import com.intellij.ide.DataManager
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.service
 import com.intellij.openapi.options.BoundConfigurable
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/config/ui/AccountConfigurable.kt
```diff
@@ -120,9 +122,14 @@ class AccountConfigurable(val project: Project) :
 
     applyChannelConfiguration()
 
+    // todo: create som pub & sub mechanism for these
```
mkondratek, line 125: important

pkukielka: `s/som/some/g`

pkukielka: While changing it could you add a bit more context there?
As I understand right now account type changes only if you reopen the dialog, but not if you change account type in the meantime, right?
Is it even realistic that we can change it? E.g. is there any external mechanism we could use for pulling such kind of info in an efficient way?

Sa1to:
https://github.com/sourcegraph/jetbrains/pull/208 should cover this

PR #195, opened 2023-12-14
update docs url

kalanchan:
update existing docs to new docs URL. 

New URL query params are `sourcegraph.com/docs/<your-docs>`
## Test plan
N/A


toolmantim (approved):
Checked all those URLs and they seem valid 👍🏼

exigow (approved):
LGTM :+1:

PR #200, opened 2023-12-15
Fix the issue with commands not loading 

### LINKED ISSUE #160: Cody commands panel won't load
### Installation Information:

I'm on the nightly build (5.2.11078-nightly) and the command panel doesn't load.

### Steps to reproduce:

Open the command panel, see that it doesn't load, click retry and see that it still doesn't load. [Loom.](https://www.loom.com/share/4a1dc1fe60354458b60cef84e39f42f2) The error also still refers to "Recipes," but these are called "Commands" now and the error message should be updated. I can only read the error message by expanding the width of the pane. Can we wrap it? 

### Expected Behaviour:

I should see a list of Commands when I click the Command tab.

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/160.

## Test plan
- runIde
- open commands tab 
- verify the commands load


Sa1to (approved):
LGTM

Issue #60, opened 2023-10-25
JetBrains: slow operation error


olafurpg:
I'm seeing this warning in my idea.log
```
2023-10-25 14:00:39,249 [ 123055] SEVERE - #c.i.u.SlowOperations - Slow operations are prohibited on EDT. See SlowOperations.assertSlowOperationsAreAllowed javadoc.
java.lang.Throwable: Slow operations are prohibited on EDT. See SlowOperations.assertSlowOperationsAreAllowed javadoc.
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:376)
	at com.intellij.util.SlowOperations.assertSlowOperationsAreAllowed(SlowOperations.java:95)
	at com.intellij.workspaceModel.core.fileIndex.impl.WorkspaceFileIndexDataImpl.ensureIsUpToDate(WorkspaceFileIndexDataImpl.kt:130)
	at com.intellij.workspaceModel.core.fileIndex.impl.WorkspaceFileIndexDataImpl.getFileInfo(WorkspaceFileIndexDataImpl.kt:75)
	at com.intellij.workspaceModel.core.fileIndex.impl.WorkspaceFileIndexImpl.getFileInfo(WorkspaceFileIndexImpl.kt:247)
	at com.intellij.workspaceModel.core.fileIndex.impl.WorkspaceFileIndexImpl.findFileSetWithCustomData(WorkspaceFileIndexImpl.kt:230)
	at com.intellij.openapi.roots.impl.ProjectFileIndexImpl.getContentRootForFile(ProjectFileIndexImpl.java:147)
	at com.intellij.openapi.roots.impl.ProjectFileIndexImpl.getContentRootForFile(ProjectFileIndexImpl.java:141)
	at com.sourcegraph.common.ProjectFileUtils.getRelativePathToProjectRoot(ProjectFileUtils.java:14)
	at com.sourcegraph.cody.context.CurrentlyOpenFileListener.selectionChanged(CurrentlyOpenFileListener.java:28)
```

olafurpg:
Another error
```
2023-10-25 14:12:23,024 [  50647] SEVERE - #c.i.u.SlowOperations - Slow operations are prohibited on EDT. See SlowOperations.assertSlowOperationsAreAllowed javadoc.
java.lang.Throwable: Slow operations are prohibited on EDT. See SlowOperations.assertSlowOperationsAreAllowed javadoc.
[...]

PR #201, opened 2023-12-15
Eliminate long operations on EDT

mkondratek:
## Test plan
- runIde
- test ide start up time (should be faster now)
- verify components (commands, sub tab) load

You can mock the initialization issue with:
```kotlin
@JvmStatic
    fun getInitializedServer(project: Project): CompletableFuture<CodyAgentServer> {
      return project.service<CodyAgent>().initialized.apply {
        println("initialized start")
        Thread.sleep(5000)
        println("initialized stop")
        this
      }
    }
```

Before some calls were definitely blocking the UI.

### DIFF
--- a/src/main/java/com/sourcegraph/cody/CodyFileEditorListener.java
+++ b/src/main/java/com/sourcegraph/cody/CodyFileEditorListener.java
@@ -1,5 +1,6 @@
 package com.sourcegraph.cody;
 
+import com.intellij.openapi.application.ApplicationManager;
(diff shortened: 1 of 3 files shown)


mkondratek:
Related to https://github.com/sourcegraph/jetbrains/issues/60.

exigow (approved):
I've tested this both before and after the fix, and indeed, there are no more UI warnings related to freezes. However, I had to introduce an artificial `sleep` to reproduce the issue. During normal usage, the difference is hard to notice.

PR #206, opened 2023-12-18
Update selected file label on the ide start up

mkondratek:
## Test plan
- runIde
- check if the label reflects the currently opened file on start-up (it can take a few seconds)

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
@@ -5,6 +5,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.sourcegraph.cody.CodyToolWindowContent
 import com.sourcegraph.cody.config.CodyProjectSettings
+import com.sourcegraph.common.ProjectFileUtils
 import com.sourcegraph.config.ConfigUtil
 import com.sourcegraph.vcs.RepoUtil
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/context/EmbeddingStatusView.kt
```diff
@@ -103,7 +103,7 @@ class EmbeddingStatusView(private val project: Project) : JPanel() {
   }
 
   fun setOpenedFileName(fileName: String, filePath: String?) {
-    openedFileContent.setText(fileName)
+    openedFileContent.text = fileName
     openedFileContent.setToolTipText(filePath)
```
pkukielka: Why you only changed one setter and not the other? :)

mkondratek: fixed

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
```diff
@@ -31,6 +32,12 @@ class CodyAgentCodebase(private val underlying: CodyAgentServer, private val pro
   fun onFileOpened(project: Project, file: VirtualFile) {
     application.executeOnPooledThread {
       onRepositoryNameChange(RepoUtil.findRepositoryName(project, file))
+
+      val openedFileName = file.name
+      val relativeFilePath: String? = ProjectFileUtils.getRelativePathToProjectRoot(project, file)
```
pkukielka, line 37: What will be the result if someone open file from outside the workspace?

mkondratek: I do not understand. what do you mean by "from outside the workspace"?

pkukielka: File which is not part of the given workspace, e.g. random file from the disk which you drag to the IDE.

mkondratek: image

`my_utils.sh` is a file in `home` that does not belong to the workspace. the tooltip does not appear for it.

pkukielka (approved):
LGTM

PR #208, opened 2023-12-19
Fix disappearing subscription tab when switching accounts

### LINKED ISSUE #186: bug: Switching between an enterprise and dotcom accounts breaks the `Subscription` tab
### Installation Information

#IU-233.11799.241 (IDEA IntelliJ Ultimate 2023.3)
nightly Cody plugin: 5.2.12527-nightly

### Describe the bug

1. Log in to an enterprise account in the plugin
2. Go to the Cody toolbar, the `Subscription` tab should not be present.
1. Separately log in to a dotcom account (with the `cody-pro-jetbrains` feature flag enabled)
3. Go to the Cody toolbar, the `Subscription` tab should be there
4. Now go back to account settings and switch back to the Enterprise account.
5. The `Subscription` tab will be gone (as it should).
6. Go to the account settings once more and switch back to the dotcom account.
7. The `Subscription` tab will still be gone, even though it should be back.

### Expected behavior

Switching back and forth between an enterprise account and a dotcom account with the feature flag enabled should not break PLG features (like showing the Subscription tab)

https://github.com/sourcegraph/jetbrains/assets/18601388/2ef1b4a9-ccfb-4f5b-b93b-679c2827433e

Sa1to:
It fixes #186 
## Test plan
* Add user with enterprise account
* Add another user with dotcom account
* Switch between enterprise and dotcom accounts
* Subscription tab should appear for dotcom account and disappear for enterprise account

https://github.com/sourcegraph/jetbrains/assets/9321940/309974c6-fc1a-45a6-912f-ce3cacfab9d7

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -136,6 +136,7 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
       var agentUserId = getUserId(server)
       var retryCount = 3
       while (jetbrainsUserId != agentUserId && retryCount > 0) {
+        Thread.sleep(200)
         retryCount--
         logger.warn("Retrying call for userId from agent")
         agentUserId = getUserId(server)
@@ -166,12 +167,12 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
     }
   }
 
-  private fun getUserId(server: CodyAgentServer): String {
+  private fun getUserId(server: CodyAgentServer): String? {
     return server
         .currentUserId()
         .exceptionally {
           logger.warn("Unable to fetch user id from agent")
-          ""
+          null
         }
         .get()
   }
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -131,24 +131,49 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
 
   private fun addNewSubscriptionTab(server: CodyAgentServer) {
     val activeAccountType = CodyAuthenticationManager.instance.getActiveAccount(project)
-    if (activeAccountType != null && activeAccountType.isDotcomAccount()) {
-      val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyProJetBrains"))
-      if (codyProFeatureFlag.get() != null && codyProFeatureFlag.get()!!) {
-        val isCurrentUserPro =
-            server
-                .isCurrentUserPro()
-                .exceptionally { e ->
-                  logger.warn("Error getting user pro status", e)
-                  null
-                }
-                .get()
-        if (isCurrentUserPro != null) {
-          val subscriptionPanel = createSubscriptionTab(isCurrentUserPro)
-          tabbedPane.insertTab(
-              "Subscription", null, subscriptionPanel, null, SUBSCRIPTION_TAB_INDEX)
+    if (activeAccountType != null) {
+      val jetbrainsUserId = activeAccountType.id
+      var agentUserId = getUserId(server)
+      var retryCount = 3
+      while (jetbrainsUserId != agentUserId && retryCount > 0) {
+        retryCount--
+        logger.warn("Retrying call for userId from agent")
+        agentUserId = getUserId(server)
+      }
+      if (jetbrainsUserId != agentUserId) {
+        logger.error("User id in JetBrains is different from agent")
+        return
+      }
+
+      if (activeAccountType.isDotcomAccount()) {
+        val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyProJetBrains"))
+        if (codyProFeatureFlag.get() != null && codyProFeatureFlag.get()!!) {
+          val isCurrentUserPro =
+              server
+                  .isCurrentUserPro()
+                  .exceptionally { e ->
+                    logger.warn("Error getting user pro status", e)
+                    null
+                  }
+                  .get()
+          if (isCurrentUserPro != null) {
+            val subscriptionPanel = createSubscriptionTab(isCurrentUserPro)
+            tabbedPane.insertTab(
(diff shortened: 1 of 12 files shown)


Sa1to:
RESOLVED ~~TODO: There is an issue with accounts added before this fix~~

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -38,6 +36,27 @@ class SettingsMigration : Activity {
     RunOnceUtil.runOnceForApp("ToggleCodyToolWindowAfterMigration") {
       ApplicationManager.getApplication().invokeLater { toggleCodyToolbarWindow(project) }
     }
+    RunOnceUtil.runOnceForApp("CodyAccountsIdsRefresh") {
```
mkondratek: wow, what's that? can you tell me a few words about `runOnceForApp`? 😮

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -131,26 +131,51 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
-          tabbedPane.insertTab(
-              "Subscription", null, subscriptionPanel, null, SUBSCRIPTION_TAB_INDEX)
+    if (activeAccountType != null) {
+      val jetbrainsUserId = activeAccountType.id
+      var agentUserId = getUserId(server)
+      var retryCount = 3
+      while (jetbrainsUserId != agentUserId && retryCount > 0) {
+        retryCount--
+        logger.warn("Retrying call for userId from agent")
+        agentUserId = getUserId(server)
```
pkukielka, line 142: How quick is that call?
If it is fast shouldn't there be some sleep in between to make sure we give it enough time to get into proper state?

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -131,26 +131,51 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
+      }
+      if (jetbrainsUserId != agentUserId) {
+        logger.error("User id in JetBrains is different from agent")
+        return
+      }
+
+      if (activeAccountType.isDotcomAccount()) {
+        val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyProJetBrains"))
```
pkukielka, line 150: `"CodyProJetBrains"` - that should be constant somewhere, we should not paste strings like that everywhere. In case of some change or typo it will be extremely hard to catch a bug there.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -131,26 +131,51 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
+        if (codyProFeatureFlag.get() != null && codyProFeatureFlag.get()!!) {
```
pkukielka, line 151: What that second check `codyProFeatureFlag.get()!!` is supposed to add there?

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
pkukielka: BTW. I know it was like that before. But since you are touching it... :)

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
Sa1to: Like IJ documentation says `Perform the task if it was not performed before for this application.`  From the code i can tell you that it stores id and boolean value in `BasePropertyService`

mkondratek:
Tested locally. I can confirm that it works 🚀

mkondratek (approved):
lgtm but can you pls earn another stamp? 😅

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -131,26 +131,51 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
       }
     }
   }
 
+  private fun getUserId(server: CodyAgentServer): String {
+    return server
+        .currentUserId()
+        .exceptionally {
+          logger.warn("Unable to fetch user id from agent")
+          ""
```
pkukielka: I'm not a fan of returning fake "success" result like empty string, ideally that should be encoded in a type (`Option` or at least `String?`). But it's not critical to change

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -70,8 +89,7 @@ class SettingsMigration : Activity {
             dotcomAccessToken,
             server,
             requestExecutorFactory,
-            EmptyProgressIndicator(ModalityState.NON_MODAL),
-            customRequestHeaders)
+            EmptyProgressIndicator(ModalityState.NON_MODAL))
```
pkukielka, line 100: What that change does in practice?

pkukielka (commented):
I left some comments

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
Sa1to: I removed `id` parameter from `addAsDefaultAccountIfUnique`. I don't know why here `customRequestHeaders` were passed as id  but it should be always fetched from `UserDetails`.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
Sa1to: `codyProFeatureFlag.get()!!` can be `true` or `false` and i want to continue here if its `true`

pkukielka: I somehow missed the fact that operator returns the value.
So in this case it's kind of redundant as you already did a null check, but it does not do any harm 👌

pkukielka (approved):
LGTM

PR #214, opened 2023-12-19
Add action for cody agent rastart

pkukielka:
## Test plan

This is small feature for developers and as such does not require testing.

To use it hit `Shift Shift` or `Shift-Cmd-A`, type `Restart Cody Agent` and hit Enter.

It will cause Cody agent to be restarted in the background. If you started your IJ using `./gradlew :runIde` you will also notice some reboot logs in the console.


olafurpg (approved):
Let's gooo!

PR #215, opened 2023-12-19
Revert "Update selected file label on the ide start up (#206)"

exigow:
This reverts commit 3c9f9a0c5345ae636604613087a45f66fda26849.

Reverted commit causes https://github.com/sourcegraph/jetbrains/issues/213.

PR #216, opened 2023-12-19
Update selected file label on the ide start up & fix slow operation on selectionChanged

mkondratek:
The first attempt has been reverted here: https://github.com/sourcegraph/jetbrains/pull/215.

`runIde` was fine - the issue appeared in a newer version of intellij. now it should be fixed. 

now, now slow operation error should appear on start up

## Test plan 
- runIde
- play with files (open a file, have a file opened at start up)
- test against 2023.3


pkukielka (approved):
LGTM (assuming we do not see the same error in the logs anymore)

PR #217, opened 2023-12-19
Agent: move from recipes to the new chat UI

olafurpg:
Previously, the JetBrains plugin used the `chat-question` recipe to power chat. This was problematic because this recipe was not used by VS Code meaning that users had a different experience in JetBrains and VS Code, usually a worse experience in JetBrains because it didn't keep up with the rapid context improvements in VS Code.

Now, with this PR, JetBrains uses new `chat/*` and `webview/*` related endpoints in the agent to power chat. This PR requires the changes from https://github.com/sourcegraph/cody/pull/2396 to be merged. The new chat UI in VS Code is entirely designed around webviews that have a low-level notification-based protocol (no requests). This low-lever protocol is a bit more complicated to work with but it supports richer functionality, including triggering indexing of local embeddings, manually adding individual files to the context, selecting the language model, and more. This PR only implements the basics: submitting a message, stopping a reply, and handling rate limit errors.

Importantly, this PR alone won't close the full gap against VS Code for chat or commands. This is just an important milestone towards that goal.

## Test plan

Only the following two cases work right now:

- Send message "Hello", it replies with "Hello!" without complaining about missing context
- Send message from Cody Free rate limited account and it displays an upsell

The following cases don't work yet, but are P0 issues that need to be fixed, ideally before merging this PR. Both of these issues are on the JetBrains side:

- Cancelation: asking "Give 5 examples comparing Scala 2 and Scala 3", click on "Stop"  button and it stops generating the reply. Currently, the stop button disappears immediately
- Transcript memory: currently, we call `chat/new` on every message, we don't preserve the transcript between messages. To test this feature, send "My name is Olafur", and send another message "What is my name?". It should reply something like "You told me your name is Olafur.".

The following cases need further validation, but are not P0 issues:

- Select some code and ask a question that requires the full open file. This doesn't work in VS Code either unless you have embeddings enabled. I didn't confirm with embeddings in JetBrains.
- Ask a question about that requires the README file. I implemented several changes in the agent to make the automatic README detection work correctly but I didn't get around to confirming this works (neither in agent or JetBrains).

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
@@ -102,9 +102,9 @@ class CodyAgent(private val project: Project) : Disposable {
     }
   }
 
-  fun shutdown() {
-    val server = getServer(project) ?: return
-    executorService.submit<CompletableFuture<Void>> {
+  fun shutdown(): Future<out CompletableFuture<Void>?>? {
+    val server = getServer(project) ?: return null
+    return executorService.submit<CompletableFuture<Void>> {
       server.shutdown().thenAccept {
         server.exit()
         agentNotRunningExplanation = "Cody Agent shut down"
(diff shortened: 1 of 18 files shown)


olafurpg:
One more case that is not working correctly yet, selecting text and asking to translate it to another language

CleanShot 2023-12-19 at 21 26 50@2x

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -102,9 +102,9 @@ class CodyAgent(private val project: Project) : Disposable {
     }
   }
 
-  fun shutdown() {
-    val server = getServer(project) ?: return
-    executorService.submit<CompletableFuture<Void>> {
+  fun shutdown(): Future<out CompletableFuture<Void>?>? {
```
pkukielka, line 105: `server.shutdown` returns `CompletableFuture`, what is the reason to additionally run it on a separate thread using `executorService`?
Wouldn't that be sufficient?
 
```kotlin
  fun shutdown(): CompletableFuture<Void>? {
    val server = getServer(project)
    return server?.shutdown()?.thenAccept {
      server.exit()
      agentNotRunningExplanation = "Cody Agent shut down"
      listeningToJsonRpc.cancel(true)
    }
  }
```

olafurpg:
Closed in favor of https://github.com/sourcegraph/jetbrains/pull/220

PR #220, opened 2023-12-20
Agent: move from recipes to the new webview-based chat API

mkondratek:
I take over Olaf's: https://github.com/sourcegraph/jetbrains/pull/217.
See the original PR for more context.

The related PR in Cody: https://github.com/sourcegraph/cody/pull/2457.

## Test plan 

- [x] Restart Chat Session
- [x] Rate limit error is displayed
- [x] User can cancel chat generation
- [x] Gracefully handle the case where we submit a message when the agent process is killed
- [x] Streaming reply

# The original description

Previously, the JetBrains plugin used the `chat-question` recipe to power chat. This was problematic because this recipe was not used by VS Code meaning that users had a different experience in JetBrains and VS Code, usually a worse experience in JetBrains because it didn't keep up with the rapid context improvements in VS Code.

Now, with this PR, JetBrains uses new `chat/*` and `webview/*` related endpoints in the agent to power chat. This PR requires the changes from https://github.com/sourcegraph/cody/pull/2396 to be merged. The new chat UI in VS Code is entirely designed around webviews that have a low-level notification-based protocol (no requests). This low-lever protocol is a bit more complicated to work with but it supports richer functionality, including triggering indexing of local embeddings, manually adding individual files to the context, selecting the language model, and more. This PR only implements the basics: submitting a message, stopping a reply, and handling rate limit errors.

Importantly, this PR alone won't close the full gap against VS Code for chat or commands. This is just an important milestone towards that goal.

## Test plan

Only the following two cases work right now:

- Send message "Hello", it replies with "Hello!" without complaining about missing context
- Send message from Cody Free rate limited account and it displays an upsell

The following cases don't work yet, but are P0 issues that need to be fixed, ideally before merging this PR. Both of these issues are on the JetBrains side:

- Cancelation: asking "Give 5 examples comparing Scala 2 and Scala 3", click on "Stop"  button and it stops generating the reply. Currently, the stop button disappears immediately
- Transcript memory: currently, we call `chat/new` on every message, we don't preserve the transcript between messages. To test this feature, send "My name is Olafur", and send another message "What is my name?". It should reply something like "You told me your name is Olafur.".

The following cases need further validation, but are not P0 issues:

- Select some code and ask a question that requires the full open file. This doesn't work in VS Code either unless you have embeddings enabled. I didn't confirm with embeddings in JetBrains.
- Ask a question about that requires the README file. I implemented several changes in the agent to make the automatic README detection work correctly but I didn't get around to confirming this works (neither in agent or JetBrains).

### DIFF
--- a/src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
+++ b/src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
@@ -3,12 +3,23 @@ package com.sourcegraph.cody.agent;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.util.Disposer;
+import com.intellij.ui.jcef.JBCefApp;
 import com.sourcegraph.cody.agent.protocol.ChatMessage;
 import com.sourcegraph.cody.agent.protocol.DebugMessage;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+
+import com.sourcegraph.find.FindPopupDialog;
+import com.sourcegraph.find.browser.JSToJavaBridgeRequestHandler;
+import com.sourcegraph.find.browser.SourcegraphJBCefBrowser;
 import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -61,5 +72,35 @@ public class CodyAgentClient {
   @JsonNotification("debug/message")
   public void debugMessage(DebugMessage msg) {
     logger.warn(String.format("%s: %s", msg.getChannel(), msg.getMessage()));
+  }
+
+  // Webviews
+  @JsonRequest("webview/create")
+  public CompletableFuture<Void> webviewCreate(WebviewCreateParams params) {
+    logger.error("webview/create This request should not happen if you are using chat/new.");
+    return CompletableFuture.completedFuture(null);
+  }
+
+  @JsonNotification("webview/postMessage")
+  public void webviewPostMessage(WebviewPostMessageParams params) {
+    if (onChatUpdateMessageInProgress != null
+        && params.getMessage().getType().equals("transcript")) {
+      if (Boolean.FALSE.equals(params.getMessage().isMessageInProgress())) {
+        onFinishedProcessing.run();
+      } else if (params.getMessage().getMessages() != null
+          && !params.getMessage().getMessages().isEmpty()) {
+        ApplicationManager.getApplication()
+            .invokeLater(
+                () ->
+                    onChatUpdateMessageInProgress.accept(
+                        Objects.requireNonNull(params.getMessage().getMessages())
+                            .get(params.getMessage().getMessages().size() - 1)));
+
+      } else {
+        logger.warn("webview/postMessage: no messages in transcript");
+      }
+    } else {
+      logger.warn(String.format("webview/postMessage %s: %s", params.getId(), params.getMessage()));
+    }
   }
 }
--- a/src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
+++ b/src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
@@ -3,21 +3,12 @@ package com.sourcegraph.cody.agent;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
-import com.intellij.openapi.util.Disposer;
-import com.intellij.ui.jcef.JBCefApp;
 import com.sourcegraph.cody.agent.protocol.ChatMessage;
 import com.sourcegraph.cody.agent.protocol.DebugMessage;
-
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-
-import com.sourcegraph.find.FindPopupDialog;
-import com.sourcegraph.find.browser.JSToJavaBridgeRequestHandler;
-import com.sourcegraph.find.browser.SourcegraphJBCefBrowser;
 import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
 import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
 import org.jetbrains.annotations.NotNull;
--- a/src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
+++ b/src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
@@ -75,7 +75,7 @@ public class CodyAgentClient {
   @JsonNotification("webview/postMessage")
   public void webviewPostMessage(WebviewPostMessageParams params) {
     if (onChatUpdateMessageInProgress != null
-        && params.getMessage().getType().equals("transcript")) {
+        && params.getMessage().getType().equals(ExtensionMessage.Type.TRANSCRIPT)) {
       if (Boolean.FALSE.equals(params.getMessage().isMessageInProgress())) {
         onFinishedProcessing.run();
       } else if (params.getMessage().getMessages() != null
@@ -91,6 +91,7 @@ public class CodyAgentClient {
         logger.warn("webview/postMessage: no messages in transcript");
       }
     } else {
+      logger.warn("onChatUpdateMessageInProgress is null or message type is not transcript");
       logger.warn(String.format("webview/postMessage %s: %s", params.getId(), params.getMessage()));
     }
   }
--- a/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
+++ b/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
@@ -35,6 +35,11 @@ data class ExtensionMessage(
     val errors: String?
 ) {
 
+  object Type {
+    const val TRANSCRIPT = "transcript"
+    const val ERRORS = "errors"
+  }
+
   fun toPanelNotFoundError(): PanelNotFoundError? {
     // e.g.: "No panel with id 414f6f9c-ed62-4d7b-8ebd-023ded81e9da found"
     if (this.errors?.matches(Regex("^No panel with id .* found$")) == true) {
--- a/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
+++ b/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
@@ -3,6 +3,7 @@ package com.sourcegraph.cody.agent
 import com.sourcegraph.cody.agent.protocol.ChatError
 import com.sourcegraph.cody.agent.protocol.ChatMessage
 import com.sourcegraph.cody.agent.protocol.ContextFile
+import com.sourcegraph.cody.agent.protocol.PanelNotFoundError
 
 /**
  * A message sent from the webview to the extension host. See vscode/src/chat/protocol.ts for the
@@ -30,7 +31,17 @@ data class ExtensionMessage(
     val chatID: String? = null,
     val isTranscriptError: Boolean? = null,
     val customPrompts: List<List<Any>>? = null,
-    val context: Any? = null
-)
+    val context: Any? = null,
+    val errors: String?
+) {
+
+  fun toPanelNotFoundError(): PanelNotFoundError? {
+    // e.g.: "No panel with id 414f6f9c-ed62-4d7b-8ebd-023ded81e9da found"
+    if (this.errors?.matches(Regex("^No panel with id .* found$")) == true) {
+      return PanelNotFoundError(this.errors)
+    }
+    return null
+  }
+}
 
 data class WebviewPostMessageParams(val id: String, val message: ExtensionMessage)
--- a/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
+++ b/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
@@ -1,0 +1,36 @@
+package com.sourcegraph.cody.agent
+
+import com.sourcegraph.cody.agent.protocol.ChatError
+import com.sourcegraph.cody.agent.protocol.ChatMessage
+import com.sourcegraph.cody.agent.protocol.ContextFile
+
+/**
+ * A message sent from the webview to the extension host. See vscode/src/chat/protocol.ts for the
+ * protocol.
+ */
+data class WebviewMessage(
+    val command: String,
+    val text: String,
+    val submitType: String, // One of: "user", "suggestion", "example"
+    val addEnhancedContext: Boolean? = null,
+    val contextFiles: List<ContextFile>? = null,
+    val error: ChatError? = null,
+)
+
+data class WebviewReceiveMessageParams(val id: String, val message: WebviewMessage)
+
+/**
+ * A message sent from the extension host to the webview. See vscode/src/chat/protocol.ts for the
+ * protocol.
+ */
+data class ExtensionMessage(
+    val type: String,
+    val messages: List<ChatMessage>? = null,
+    val isMessageInProgress: Boolean? = null,
+    val chatID: String? = null,
+    val isTranscriptError: Boolean? = null,
+    val customPrompts: List<List<Any>>? = null,
+    val context: Any? = null
+)
+
+data class WebviewPostMessageParams(val id: String, val message: ExtensionMessage)
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -23,11 +23,7 @@ import com.sourcegraph.cody.agent.CodyAgent.Companion.getInitializedServer
 import com.sourcegraph.cody.agent.CodyAgent.Companion.isConnected
 import com.sourcegraph.cody.agent.CodyAgentManager.tryRestartingAgentIfNotRunning
 import com.sourcegraph.cody.agent.CodyAgentServer
-import com.sourcegraph.cody.agent.protocol.ChatMessage
-import com.sourcegraph.cody.agent.protocol.ContextMessage
-import com.sourcegraph.cody.agent.protocol.GetFeatureFlag
-import com.sourcegraph.cody.agent.protocol.RecipeInfo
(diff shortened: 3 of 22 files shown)


mkondratek:
Transcript memory - fixed ✅ 

image

src/main/kotlin/com/sourcegraph/cody/webview/CodyWebviewPanel.kt
```diff
@@ -0,0 +1,3 @@
+package com.sourcegraph.cody.webview
+
+class CodyWebviewPanel(id: String) {}
```
exigow: It seems that this class is not used anywhere.

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * A message sent from the extension host to the webview. See vscode/src/chat/protocol.ts for the
+ * protocol.
+ */
+data class ExtensionMessage(
+    val type: String,
+    val messages: List<ChatMessage>? = null,
+    val isMessageInProgress: Boolean? = null,
+    val chatID: String? = null,
```
exigow, line 31: Why `chatID` is a date? (eg. `Tue, 02 Jan 2024 13:40:02 GMT`)

Maybe we should use UUID instead?

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
```diff
@@ -50,52 +54,114 @@ class Chat {
...
+                          null
+                        }
+                    val panelNotFoundError =
+                        if (lastReply.type == "errors" && lastReply.errors != null) {
+                          lastReply.toPanelNotFoundError()
+                        } else {
+                          null
+                        }
+                    logger.warn("rateLimitError: $rateLimitError")
+                    logger.warn("panelNotFoundError: $panelNotFoundError")
```
exigow: This looks like logs for debugging purposes during development. IMO they seem obsolete at this point. Alternatively, there definitely shouldn't be a `warn`.

Note: there is also a funny clutter in the logs like "STAAAAAAAAAAAAAAAAAART" and "END" that seems unrelated to these changes. I suspect someone may have forgotten to remove it from the agent.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -62,4 +64,34 @@ public void chatUpdateMessageInProgress(ChatMessage params) {
...
+  @JsonRequest("webview/create")
+  public CompletableFuture<Void> webviewCreate(WebviewCreateParams params) {
+    logger.error("webview/create This request should not happen if you are using chat/new.");
+    return CompletableFuture.completedFuture(null);
+  }
+
+  @JsonNotification("webview/postMessage")
+  public void webviewPostMessage(WebviewPostMessageParams params) {
+    if (onChatUpdateMessageInProgress != null
+        && params.getMessage().getType().equals("transcript")) {
```
pkukielka: Can we replace hardcoded string `"transcript"` with some const?

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -62,4 +64,34 @@ public void chatUpdateMessageInProgress(ChatMessage params) {
...
+                () ->
+                    onChatUpdateMessageInProgress.accept(
+                        Objects.requireNonNull(params.getMessage().getMessages())
+                            .get(params.getMessage().getMessages().size() - 1)));
+
+      } else {
+        logger.warn("webview/postMessage: no messages in transcript");
+      }
+    } else {
+      logger.warn(String.format("webview/postMessage %s: %s", params.getId(), params.getMessage()));
```
pkukielka, line 95: Can we make this warning more informative? Now we do not print out what is the reason of the warning.

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
```diff
@@ -50,52 +54,114 @@ class Chat {
...
+                    if (rateLimitError != null) {
+                      handleRateLimitError(project, chat, rateLimitError)
+                    } else if (panelNotFoundError != null) {
+                      chat.loadNewChatId {
+                        this.sendMessageViaAgent(project, humanMessage, recipeId, chat, token)
```
exigow: `this` can be removed

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -62,4 +64,34 @@ public void chatUpdateMessageInProgress(ChatMessage params) {
...
+    if (onChatUpdateMessageInProgress != null
+        && params.getMessage().getType().equals("transcript")) {
+      if (Boolean.FALSE.equals(params.getMessage().isMessageInProgress())) {
+        onFinishedProcessing.run();
+      } else if (params.getMessage().getMessages() != null
+          && !params.getMessage().getMessages().isEmpty()) {
+        ApplicationManager.getApplication()
+            .invokeLater(
+                () ->
+                    onChatUpdateMessageInProgress.accept(
```
pkukielka, line 86: So if there is no messages exception will be thrown and most likely pop put on the UI.
In theory that situation should never happen?

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.sourcegraph.cody.agent.protocol.PanelNotFoundError
+
+/**
+ * A message sent from the webview to the extension host. See vscode/src/chat/protocol.ts for the
+ * protocol.
+ */
+data class WebviewMessage(
+    val command: String,
+    val text: String,
+    val submitType: String, // One of: "user", "suggestion", "example"
```
pkukielka, line 15: Is there anything which prevents us from making this proper enum?

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
```diff
@@ -0,0 +1,47 @@
...
+)
+
+data class WebviewReceiveMessageParams(val id: String, val message: WebviewMessage)
+
+/**
+ * A message sent from the extension host to the webview. See vscode/src/chat/protocol.ts for the
+ * protocol.
+ */
+data class ExtensionMessage(
+    val type: String,
```
pkukielka, line 28: Similarly, can we make it enum?

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
exigow: BTW this logging affects this code block. It harms in a way that the implementation focuses on assigning objects to variables just to be able to print their values (see: panelNotFoundError). If we remove logging, it's immediately clear that we only want to check the condition.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
```diff
@@ -72,4 +72,18 @@ interface CodyAgentServer {
...
   @JsonNotification("$/cancelRequest") fun cancelRequest(cancelParams: CancelParams)
+
+  // Webviews
+  @JsonRequest("webview/didDispose") fun webviewDidDispose(): CompletableFuture<Void?>
+
+  @JsonNotification("webview/receiveMessage")
+  fun webviewReceiveMessage(params: WebviewReceiveMessageParams)
+
+  @JsonRequest("command/execute")
+  fun commandExecute(params: CommandExecuteParams): CompletableFuture<Any?>
```
exigow, line 83: Some of these methods are not used (IMO we should avoid introducing them in advance to main branch but you can ignore my comment).

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -129,6 +132,20 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
   }
 
+  override fun loadNewChatId(callback: () -> Unit) {
+    id = null
+    promptPanel.textArea.isEnabled = false
+    promptPanel.textArea.emptyText.text = "Connecting to agent..."
+    ApplicationManager.getApplication().executeOnPooledThread {
+      getInitializedServer(project).thenAccept { server ->
+        id = server.chatNew().get()
+        promptPanel.textArea.isEnabled = true
```
pkukielka: Isn't that UI update on non-ui thread?

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -129,6 +132,20 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
+        promptPanel.textArea.emptyText.text = "Ask a question about this code..."
+        callback.invoke()
```
pkukielka, line 147: Also, it makes assumption on which thread callback should be called.
Since you cannot control there what will be in the callback code (e.g. long operation vs UI update) running it in proper theread should be responsibility of the callback code, not the caller.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ChatMessage.kt
```diff
@@ -1,10 +1,44 @@
...
+  fun toRateLimitError(): RateLimitError? {
+    if (this.retryAfter == null ||
+        this.limit == null ||
+        this.userMessage == null ||
+        this.retryMessage == null ||
+        this.feature == null ||
+        this.upgradeIsAvailable == null) {
+      return null
+    }
+    return RateLimitError(
```
pkukielka, line 21: Do we not copy all the parameters, e.g. `feature`?
And if it's not needed then why we check if it's not null?

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
```diff
@@ -50,52 +54,114 @@ class Chat {
...
+                  val reply =
+                      server.chatSubmitMessage(
+                          ChatSubmitMessageParams(
+                              chat.id!!,
+                              WebviewMessage(
+                                  command = "submit",
+                                  text = humanMessage.actualMessage(),
+                                  submitType = "user",
+                                  addEnhancedContext = true,
+                                  // TODO: allow to manually add files to the context via `@`
```
pkukielka: Please convert to ticket if not done already.

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
```diff
@@ -50,52 +54,114 @@ class Chat {
...
+                                  contextFiles = listOf())))
+                  token.onCancellationRequested { reply.cancel(true) }
+                  reply.handle { lastReply, error ->
+                    val rateLimitError =
+                        if (lastReply.type == "transcript" &&
```
pkukielka: I think this can be done simpler and more idiomatic:

```kotlin
val error = lastReply.getError() // Return proper error class, e.g. PanelNotFoundError, all should inherit from common parent 
when (error) {
    is PanelNotFoundError -> ...
    is RateLimitError -> ...
    else -> ...
}
```

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
```diff
@@ -50,52 +54,114 @@ class Chat {
...
+                      }
+                    } else if (error != null) {
+                      handleError(project, error, chat)
+                      null
+                    } else {
+                      RateLimitStateManager.invalidateForChat(project)
+                    }
+                  }
+                } catch (ignored: Exception) {
+                  // Ignore bugs in the agent when executing recipes
```
pkukielka, line 104: We don't want to at least log them?

exigow (commented):
I added a few comments. I clicked through the features, and there is definitely an improvement.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
mkondratek: yes, IIUC, if we receive a transcript type it should contain a message

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
mkondratek: for now, it's always `"user"` and there is only one place where we instantiate it

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ChatMessage.kt
mkondratek: afaik: `feature` is either "autocomplete" or "chat" (or smth else?) 

we do not need them. I simplified the API in this scope

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
mkondratek: I tried this:

```kt
 enum class Type {
    @SerializedName("transcript") TRANSCRIPT,
    @SerializedName("errors") ERRORS
  }
  ```
  
  but it fails on deserialization 😕 switched to const vals eventually

File: src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java (hunk 5 above)
Comment (mkondratek) [reply src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java:95]: I added one more message here to make it more precise

File: src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java (hunk 4 above)
Comment (mkondratek) [reply src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java]: done

File: src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt (hunk 14 above)
Comment (mkondratek) [reply src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt]: https://github.com/sourcegraph/jetbrains/issues/242 created

File: src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt (hunk 15 above)
Comment (mkondratek) [reply src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt]: makes sense but it would require a change to the agent. pls take a look:

```kt
val rateLimitError =
    if (lastReply.type == ExtensionMessage.Type.TRANSCRIPT &&
        lastReply.messages?.lastOrNull()?.error != null) {
      lastReply.messages.lastOrNull()?.error?.toRateLimitError()
    } else {
      null
    }
val panelNotFoundError =
    if (lastReply.type == ExtensionMessage.Type.ERRORS &&
        lastReply.errors != null) {
      lastReply.toPanelNotFoundError()
    } else {
      null
    }
```

For now, here we use two DIFFERENT ways to handle errors. The rate limit error comes from `ExtensionMessage/ChatMessage/ChatError` (I think we can call it **a message error**). However, the panel not found error comes from `ExtensionMessage` directly (I would call it **a reply error**). 

This distinguishment between the message error and the reply error can be reasonable (and probably is in this case).

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
mkondratek: We do not create initialise `chatID` on the JetBrains side. pls take a look at the server - this `chatID` comes from `sessionID` which is somewhat internal in agent. I agree that it looks a bit weird but I would rather avoid adding any extensive changes that are not critical.

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
mkondratek: yup, to be removed before merging

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
mkondratek: I'd rather keep them. They may look redundant for now but the file remains clean and readable. I assume that we will need them soon for commands (migrated from recipes) and for multiple chat handling (`webview/didDispose`).

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
mkondratek: fixed

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
mkondratek: The callback is to be called having the new `id` assigned. We need to do things in the following order:
1.  Reset `id` (set `id` to `null`)
2. Update UI with the connecting state (`text = "Connecting to agent..."`)
3. With an initialized server send a request for the new chat (that cannot be called on UI thread) 
4. Having the new `id`, update UI with the ready state (`emptyText.text = "Ask a question about this code..."`)
5. Run callback

mkondratek: I reviewed the threads run here. The UI actions are called from AWT. The other actions are on pooled threads.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -417,9 +436,8 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
       messagesPanel.revalidate()
       messagesPanel.repaint()
       chatMessageHistory.clearHistory()
-      ApplicationManager.getApplication().executeOnPooledThread {
-        getInitializedServer(project).thenAccept { it?.transcriptReset() }
-      }
+      // todo (#260): call agent to reset the transcript instead of unsetting the chat id
```
mkondratek, line 440: this is more a workaround but it's quick (https://github.com/sourcegraph/jetbrains/issues/260)

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -91,6 +88,8 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
     stopGeneratingButton.addActionListener {
       inProgressChat.abort()
       stopGeneratingButton.isVisible = false
+      sendButton.isEnabled = promptPanel.textArea.text.isNotBlank()
+      ensureBlinkingCursorIsNotDisplayed()
```
mkondratek, line 92: unrel fixes

mkondratek:
I tested all listed scenarios
```
 Restart Chat Session
 Rate limit error is displayed
 User can cancel chat generation
 Gracefully handle the case where we submit a message when the agent process is killed
 Streaming reply
 ``` 
 
 It works fine for now (overall experience is not worse than before, the context issue does not appear).
 
 I did not check them yet as I expect some review fixes.

mkondratek:
I built the plugin, installed it on my actual IDE and run two instances of the IDE (to test it against https://github.com/sourcegraph/jetbrains/issues/243) - I was able to message both chats 🚀

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
pkukielka: @mkondratek Looks like logging is still there and perhaps structure of the code could be improved without it?

pkukielka (approved):
LGTM

Sa1to (approved):
LGTM

PR #223, opened 2023-12-21
Improve CONTRIBUTING.md docs and add option to use :runIde with custom IJ version

pkukielka:
## Test plan

Not needed as this is only impacting dev docs and workflow.

Changes to `build.gradle.kts` I verified manually and for the standard workflow are also verified each time by CI run.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,15 +1,8 @@
 import com.jetbrains.plugin.structure.base.utils.isDirectory
-import java.lang.IllegalArgumentException
 import java.net.URL
-import java.nio.file.FileSystems
-import java.nio.file.FileVisitResult
-import java.nio.file.Files
-import java.nio.file.PathMatcher
-import java.nio.file.Paths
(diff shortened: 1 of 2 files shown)


Gedochao (approved):
LGTM

build.gradle.kts
```diff
@@ -337,6 +337,15 @@ tasks {
     systemProperty(
         "cody.autocomplete.enableFormatting",
         project.property("cody.autocomplete.enableFormatting") ?: "true")
+
+    val platformRuntimeVersion = project.findProperty("platformRuntimeVersion")
```
mkondratek, line 341: I assume these changes allow us to specify the ide version as a runIde flag, is that right?

mkondratek (approved):
lgtm

build.gradle.kts
pkukielka: Correct :)

Sa1to (approved):
LGTM

PR #230, opened 2023-12-22
Fix issues with chat colours after switching UI theme

### LINKED ISSUE #229: bug: Chat window colours are not repainted properly after the theme change
### Installation Information

Happens in all branches and versions.

### Describe the bug

To reproduce:

1. Ask Cody Chat question to which it will reply with various layout elements (list, code snippets, etc)
2. Change theme in settings or using `Themes` action.

You will notice that some sections of background and foreground will have wrong colours and will be unreadable.

### Expected behavior

Changing theme should lead to full repaint of the colours according to the current theme.

_No response_

pkukielka:
It fixes [#229](https://github.com/sourcegraph/jetbrains/issues/229)

## Test plan

1. Ask Cody Chat question to which it will reply with various layout elements (list, code snippets, etc)
2. Change theme in settings or using `Themes` action.
3. Verify that chat text is readable, as well as that there is clear colour distinction between user and assistant sections. Both sections colors should also correspond to the chosen theme.

Repeat the above starting from different themes. 

## PR notes

We should try to avoid hardcoding color computations and instead always try to inherit them from the parent elements.
This way recompilations after theme change etc. are straightforward.
If we have to hardcode anything we must make sure it's in a place where it will be recomputed on either theme change or a repaint.

Colors fix
Colors fix
Colors fix


src/main/kotlin/com/sourcegraph/cody/ui/HtmlViewer.kt
```diff
@@ -32,14 +25,4 @@ object HtmlViewer {
     jEditorPane.addHyperlinkListener(BrowserHyperlinkListener.INSTANCE)
     return jEditorPane
   }
-
```
pkukielka, line 35: @mkondratek I think you added it, do you maybe remember if it was needed for any particular case?
According to my testing it rather seems to be unused (which might be good thing, because it would probably not recompute on theme switch anyway).
Screenshots from the PR are without that code.

src/main/java/com/sourcegraph/cody/chat/AssistantMessageWithSettingsButton.java
pkukielka, line 1: Unused.

exigow (changes requested):
Nice fix! :+1: I didn't expect that the problem was with the hardcoded colors.

Two minor issues:
* You need to run `./gradlew spotlessApply`
* You might consider adding a simple verification step to `TESTING.md` that expects all components to render in the IDE's native colors when color scheme is changed. It seems to me that it would be worthwhile to add a point to pay attention to this (from manual testing perspective), but it's up to you. IMO, this is one of those things that we will never be able to test automatically and we have to use *human eyes*.

src/main/kotlin/com/sourcegraph/cody/ui/HtmlViewer.kt
mkondratek: Something strange happened to the repository. I do not recall changing this file. For some reason, git(-hub?) assigns me as the author of many changes.

mkondratek (approved):

pkukielka:
> Nice fix! 👍 I didn't expect that the problem was with the hardcoded colors.
> 
> Two minor issues:
> 
> * You need to run `./gradlew spotlessApply`
> * You might consider adding a simple verification step to `TESTING.md` that expects all components to render in the IDE's native colors when color scheme is changed. It seems to me that it would be worthwhile to add a point to pay attention to this (from manual testing perspective), but it's up to you. IMO, this is one of those things that we will never be able to test automatically and we have to use _human eyes_.

Good idea, I have added it :)
Looks like IntelliJ also reformatted some of the existing docs, but I guess code standards are the topic for other time ^^
(I see spotless takes care of some cases, but looks like not for all).

exigow (approved):

PR #238, opened 2023-12-29
Fix the subscription panel presentation on `Sign in with an Enterprise Instance`

mkondratek:
We had no issue for that. I just noticed it. The subscription tab is not appearing after login from Github etc. 

This PR fixes the issue + it moves the logic related to account changes to [AccountSettingChangeListener.kt](https://github.com/sourcegraph/jetbrains/pull/238/files#diff-4a4dd118349f1f574c51f4eccff0ced92175b76217627c7c21c5a936d841cc9a).

## Test plan
- test log in with GitHub/GitLab/Google
- test sign in with an enterprise

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/config/CodyPersistentAccountsHost.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/CodyPersistentAccountsHost.kt
@@ -1,10 +1,9 @@
 package com.sourcegraph.cody.config
 
 import com.intellij.openapi.project.Project
-import com.sourcegraph.cody.CodyToolWindowContent
 import com.sourcegraph.cody.agent.CodyAgent
-import com.sourcegraph.cody.statusbar.CodyAutocompleteStatusService
-import com.sourcegraph.common.UpgradeToCodyProNotification
+import com.sourcegraph.cody.config.notification.AccountSettingChangeActionNotifier
+import com.sourcegraph.cody.config.notification.AccountSettingChangeContext
 import com.sourcegraph.config.ConfigUtil
 
 class CodyPersistentAccountsHost(private val project: Project?) : CodyAccountsHost {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/config/CodyPersistentAccountsHost.kt
```diff
@@ -22,12 +21,11 @@ class CodyPersistentAccountsHost(private val project: Project?) : CodyAccountsHo
       // Notify Cody Agent about config changes.
       CodyAgent.getServer(project)
           ?.configurationDidChange(ConfigUtil.getAgentConfiguration(project))
-      val codyToolWindowContent = CodyToolWindowContent.getInstance(project)
-      codyToolWindowContent.refreshPanelsVisibility()
-      codyToolWindowContent.embeddingStatusView.updateEmbeddingStatus()
-      UpgradeToCodyProNotification.autocompleteRateLimitError.set(null)
-      UpgradeToCodyProNotification.chatRateLimitError.set(null)
-      CodyAutocompleteStatusService.resetApplication(project)
+
```
pkukielka, line 24: Is `addAccount` called every time plugin is started?
Previously that logic was also called if user was already logged in, and panel was created for the first time.
How will it work now if it will be only called on account settings change?

mkondratek: I did a quick search and it seems `addAccount` is not called unless a user performs the login action.

> Previously that logic was also called if user was already logged in, and pans was created for the first time.

I do not understand. The whole reset logic is called here as before but from `AccountSettingChangeListener`. Here, from `addAccount` and there, from `AccountConfigurable.kt::apply`.

pkukielka: Yes, I'm talking about `AccountConfigurable.kt::apply.`.
That function was called in various situations, e.g.:

```
    if (connectionAndAuthState == ConnectionAndAuthState.COULD_NOT_CONNECT) {
      // ...
      emptyText.appendLine(
          "Click here to configure your Sourcegraph Cody + Code Search settings.",
          new SimpleTextAttributes(STYLE_PLAIN, JBUI.CurrentTheme.Link.Foreground.ENABLED),
          __ ->
              ShowSettingsUtil.getInstance()
                  .showSettingsDialog(project, AccountConfigurable.class));

```

If I understand this code correctly clicking on that link will create panel which in turn will instantiate `AccountConfigurable`. Previously it was causing reset logic to be executed, and now I think it's not anymore?

pkukielka: Resolved offline, at the beginning of the `apply` method there is notification to the listener, so everything should work as expected.

pkukielka (approved):
LGTM

PR #239, opened 2023-12-29
Disable Commands while a message is being generated

mkondratek:
We have no issue for that and the Commands are to be migrated to commands (we are still using recipes actually) but this is a quick and small fix. 

## Test plan
1. run command/write a prompt
2. verify that commands are disabled during the message generation but are reenabled after that


mkondratek:
https://github.com/sourcegraph/jetbrains/assets/19799111/5b04bf54-21de-4102-9f94-2ac9cb3e8ada

pkukielka (approved):
LGTM

PR #240, opened 2023-12-29
Wrap "Already upgraded..." text in the subscription tab

mkondratek:
We have no issue for this one but the text exceeding the line width is not visible - these changes fix that.

## Test plan
- play with Cody tool window width 
- verify that the text is wrapped


mkondratek:
| Before |      After      |
|----------|-------------|
| image | image |

pkukielka (approved):
LGTM

PR #262, opened 2024-01-08
Update CONTRIBUTING.md

beyang:
- remove sourcegraph/sourcegraph dependency
- prefer `asdf` to install deps
- break into subsections

## Test plan

Dev docs only


steveyegge (approved):

CONTRIBUTING.md
```diff
@@ -11,26 +11,16 @@ our [issue tracker](https://github.com/sourcegraph/cody/issues/new/choose).
...
   Confirm that you have Java 11 installed with `java -version`.
-- Clone `https://github.com/sourcegraph/sourcegraph`
 - Clone `https://github.com/sourcegraph/cody` in a sibling directory.
-  The toplevel directories for sourcegraph/sourcegraph and sourcegraph/cody must be next to each other.
-- Install the following two IntelliJ plugins to format Java and Kotlin on file save
-    - https://plugins.jetbrains.com/plugin/8527-google-java-format
-    - https://plugins.jetbrains.com/plugin/14912-ktfmt
+- [Install `asdf`](https://asdf-vm.com/guide/getting-started.html)
+  - Run `asdf install` in the root directory of this repository
+  - If not using `asdf`, you can install the programs in `.tool-versions` through other means, but ensure the versions match.
```
mkondratek: I'd rather not reference `nodejs` & `pnpm` (`.tool-versions`) so directly. They are the agent's requirements (not plugin's). Hence, `asdf` mention seems to be redundant here too. 

The agent has its own `.tool-versions`. Having the same lib version declarations in these two files may lead to inconsistency. What if the agent specifies  `nodejs 20.4.0` but the plugin tells it's `18.18.2` (this is the actual case right now)?

olafurpg: I think it makes sense to keep .tool-versions in sync with the agent .tool-versions. I have hit on issues where building the agent from source fails if I don't have the right dependencies where I run `./gradlew :runIde`

olafurpg (approved):

CONTRIBUTING.md
pkukielka: Why do we need both .tool-windows if `jetbrains` on itself does not need pnpm? 
I hit the same issues when I tried first time to build everything, but at that time I haven't got `asdf` installed.
My understanding is that when `grade` will run `pnpm` in the cody dir, it should automatically read `asdf` settings from `cody` and use proper runtime automatically. Isn't that the case?

As for `asdf` installation and other dependencies needed for cody I thin we should simply link to the:
https://github.com/sourcegraph/cody/blob/main/doc/dev/index.md
This way we won't need to manually keep both docs in sync.

olafurpg: Agree that we should ideally pick up the asdf settings from the Cody repo. I haven't gotten this working, but I haven't spent much time investigating it either because I've been able to work around it by having the right environment when running Gradle.

I'm fine either way we go.

BTW, I use rtx instead of asdf because it it's more intuitive and works more reliably for me. rtx is compatible with asdf `.tool-version` files.

PR #274, opened 2024-01-09
Disable displaying feature limits

vdavid:
The problem is that we're displaying 160 chats in the error message:

image

while on the /cody/manage page, we display 20:

image

Cody Gateway only knows the LLM limits (which is 160 for Free users), and that's what it returns in the error header, so that's all the client knows as well. We should display the official limit of 20 instead, but we can't know it for now on the client side.

So my quick solution is just to hide this info on the UI.

## Test plan

Relying on the JetBrains team to test this


kalanchan:
raised [here](https://sourcegraph.slack.com/archives/C05SZB829D0/p1704824393085649)

cc @sourcegraph/jetbrains-plugin @danielmarquespt

mkondratek (approved):
pls test

PR #284, opened 2024-01-10
Fix slow op on refreshSubscriptionTab + restart the agent on jetbrains' and agent's user id mismatch

mkondratek:
## Test plan 
- put a breakpoint in `addNewSubscriptionTab` and set `agentUserId` a dummy value so it mismatch the `jetbrainsUserId`(that will test the error handling)
  - internal ide error should not appear 
  - agent should restart and a new proper matching id should be assigned 

- in `addNewSubscriptionTab` increase the timeout & runIde
  - UI should not freeeze 🥶

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -4,23 +4,24 @@ import com.intellij.icons.AllIcons
 import com.intellij.ide.ui.laf.darcula.ui.DarculaButtonUI
 import com.intellij.openapi.actionSystem.*
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.invokeLater
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.fileEditor.FileEditorManager
 import com.intellij.openapi.project.DumbAwareAction
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.VerticalFlowLayout
-import com.intellij.openapi.util.Disposer
 import com.intellij.ui.SimpleTextAttributes
 import com.intellij.ui.components.JBPanelWithEmptyText
 import com.intellij.ui.components.JBTabbedPane
 import com.intellij.util.IconUtil
+import com.intellij.util.concurrency.annotations.RequiresBackgroundThread
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.ui.JBUI
-import com.intellij.vcs.log.runInEdtAsync
 import com.intellij.xml.util.XmlStringUtil
 import com.sourcegraph.cody.agent.CodyAgent.Companion.getInitializedServer
 import com.sourcegraph.cody.agent.CodyAgent.Companion.isConnected
+import com.sourcegraph.cody.agent.CodyAgentManager
 import com.sourcegraph.cody.agent.CodyAgentManager.tryRestartingAgentIfNotRunning
 import com.sourcegraph.cody.agent.CodyAgentServer
 import com.sourcegraph.cody.agent.protocol.*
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -111,20 +112,21 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
 
     addWelcomeMessage()
-    refreshSubscriptionTab()
+    ApplicationManager.getApplication().executeOnPooledThread { refreshSubscriptionTab() }
     loadNewChatId()
   }
 
+  @RequiresBackgroundThread
   fun refreshSubscriptionTab() {
-    runInEdtAsync(Disposer.newCheckedDisposable()) {
```
mkondratek, line 119: tbh, I think we shouldn't have it called on Edt (whatever means "Async" here). for an experiment I switched `Thread.sleep(200)` to wait a few seconds instead - I was freezing the UI 😕 

I marked this method as `RequiresBackgroundThread` and called it on     `ApplicationManager.getApplication().executeOnPooledThread` where we call it.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -111,20 +112,21 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
-        if (tabbedPane.tabCount < SUBSCRIPTION_TAB_INDEX + 1) {
-          addNewSubscriptionTab(server)
-        } else {
+    tryRestartingAgentIfNotRunning(project)
+    getInitializedServer(project).thenAccept { server ->
+      if (tabbedPane.tabCount < SUBSCRIPTION_TAB_INDEX + 1) {
+        addNewSubscriptionTab(server)
+      } else {
+        ApplicationManager.getApplication().invokeLater {
           tabbedPane.remove(SUBSCRIPTION_TAB_INDEX)
```
mkondratek, line 127: this is an UI actions so we must get back to UI thread (`invokeLater`)

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -163,7 +165,9 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
       }
       if (jetbrainsUserId != agentUserId) {
         if (agentUserId != null) {
-          logger.error("User id in JetBrains is different from agent")
+          logger.warn("User id in JetBrains is different from agent: restarting agent...")
```
mkondratek, line 168: changed to warn (so a user does not see an internal error anymore) + restarting the agent (which should be a fix)

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -181,9 +185,11 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
                   }
                   .get()
           if (isCurrentUserPro != null) {
-            val subscriptionPanel = createSubscriptionTab(isCurrentUserPro)
-            tabbedPane.insertTab(
-                "Subscription", null, subscriptionPanel, null, SUBSCRIPTION_TAB_INDEX)
+            ApplicationManager.getApplication().invokeLater {
+              val subscriptionPanel = createSubscriptionTab(isCurrentUserPro)
+              tabbedPane.insertTab(
```
mkondratek, line 190: UI actions - moved to UI thread

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -111,20 +112,21 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
-    runInEdtAsync(Disposer.newCheckedDisposable()) {
-      tryRestartingAgentIfNotRunning(project)
-      getInitializedServer(project).thenAccept { server ->
-        if (tabbedPane.tabCount < SUBSCRIPTION_TAB_INDEX + 1) {
-          addNewSubscriptionTab(server)
-        } else {
+    tryRestartingAgentIfNotRunning(project)
+    getInitializedServer(project).thenAccept { server ->
+      if (tabbedPane.tabCount < SUBSCRIPTION_TAB_INDEX + 1) {
+        addNewSubscriptionTab(server)
```
pkukielka, line 124: I looked into `addNewSubscriptionTab` and now it seems to be doing both long operations and UI operations.
Now long operations are on the background thread, but UI operations (like `tabbedPane.insertTab`) are not on the dispatch thread.

Sa1to (approved):
LGTM

exigow (approved):
Looks good :+1:. I confirm that the agent is forced to restart after ID mismatch. Subscription panel is not freezing.

PR #285, opened 2024-01-10
Update codyCommit

mkondratek:
This PR is important b/c it brings a crucial fix to the telemetry 

## Test plan 
- test everything 🤡

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -88,6 +88,10 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
       stopGeneratingButton.isVisible = false
       sendButton.isEnabled = promptPanel.textArea.text.isNotBlank()
       ensureBlinkingCursorIsNotDisplayed()
+      recipesPanel.components.filterIsInstance<JButton>().forEach {
+        it.isEnabled = true
+        it.toolTipText = null
+      }
     }
     stopGeneratingButton.isVisible = false
     stopGeneratingButtonPanel.add(stopGeneratingButton)
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -68,9 +68,6 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
     recipesPanel.layout = BoxLayout(recipesPanel, BoxLayout.Y_AXIS)
     tabbedPane.insertTab("Commands", null, recipesPanel, null, RECIPES_TAB_INDEX)
 
-    // Initiate filling recipes panel in the background
-    refreshRecipes()
-
     // Chat panel
     messagesPanel.layout = VerticalFlowLayout(VerticalFlowLayout.TOP, 0, 0, true, true)
     val chatPanel = ChatScrollPane(messagesPanel)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/ExtensionConfiguration.kt
```diff
@@ -9,6 +9,8 @@ data class ExtensionConfiguration(
     val autocompleteAdvancedProvider: String? = null,
     val autocompleteAdvancedServerEndpoint: String? = null,
     val autocompleteAdvancedAccessToken: String? = null,
+    val customConfiguration: Map<String, Any> =
+        mapOf("cody.useContext" to "embeddings", "cody.experimental.symfContext" to false),
```
mkondratek: @olafurpg , does it make sense?

olafurpg: This looks right 👍🏻

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -114,6 +111,9 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
     addWelcomeMessage()
     ApplicationManager.getApplication().executeOnPooledThread { refreshSubscriptionTab() }
     loadNewChatId()
+
+    // Initiate filling recipes panel in the background
+    refreshRecipes()
```
mkondratek, line 118: moved from above - probably no actual impact but looks better at the end. I will provide some cleanup to this class in a separate PR

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -215,6 +215,7 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
   }
 
   private fun loadCommands() {
+    tryRestartingAgentIfNotRunning(project)
```
mkondratek, line 225: I tried `:runIde` a few times - with a rate-limited account I saw the issue with commands not loading - this line seems to be a fix

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -91,6 +88,10 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
       stopGeneratingButton.isVisible = false
       sendButton.isEnabled = promptPanel.textArea.text.isNotBlank()
       ensureBlinkingCursorIsNotDisplayed()
+      recipesPanel.components.filterIsInstance<JButton>().forEach {
+        it.isEnabled = true
+        it.toolTipText = null
+      }
```
mkondratek, line 94: unrel fix

src/main/kotlin/com/sourcegraph/cody/agent/ExtensionConfiguration.kt
mkondratek: we decided to just cherry pick the telemetry fix - these adjustments are redundant

pkukielka (approved):
LGTM

exigow (changes requested):
Chat input seems to be infinitely locked on "Connecting to agent" on private repositories. On public repos like sourcegraph/jetbrains/cody/etc this problem doesn't exists (chat connects to the agent and after a moment the text input becomes usable).

Selection_033

I've double-checked this and it appears that the current main branch doesn't have this issue, so the regression must have been introduced with this PR (which seems very strange since the changes are cherry-picked and are solely related to telemetry).

Does anyone have similar experience?

EDIT/NOTE: Sometimes it's working if we launch a private repo first. The problem is fully reproducible if we open another private repo within the same "session".

exigow (approved):
Works great. The last fix nailed it. :+1:

Sa1to (approved):

PR #291, opened 2024-01-12
Add "General commands availability in context menu" to TESTING.md

exigow:
I am documenting this because I wasn't fully aware of this functionality (personally, I'm trying to avoid using mouse during coding, and this menu is only accessible with a right-click), especially since it recently turned out that this menu wasn't working at all.

## Test plan

N/A


pkukielka (approved):
LGTM

Issue #135, opened 2023-11-20
Add single-click bug report and improve repository discoverability


exigow:
Discussion: https://sourcegraph.slack.com/archives/C059N5FRYG3/p1700485424699019

I noticed that it is currently difficult to report a bug or discover code if user is not associated with Sourcegraph. I think we should improve that if we want to attract open source developers.

Things to do:
1. We are missing single-click _Report Exception_ button when error occurs - to enable this, we can implement `<errorHandler>` with automatic stacktrace attachment (maybe similar to https://github.com/exigow/intellij-gdscript/blob/master/src/main/kotlin/error/ErrorGithubSubmitter.kt?) ![Selection_013](https://github.com/sourcegraph/jetbrains/assets/5013708/ae53666f-5a0d-44e7-9521-007e195c7464)
2. Issue Tracker on marketplace plugin page is pointing to previous repo - this can be only updated by the plugin admin
3. links in docs are pointing to old repo

taylorsperry:
I love the idea of enabling users to send us error reports with logs that help us debug. Would it be possible for those reports to automatically create issues that have `bug` labels so they make it onto our board? Or what's the flow that you imagine? 

I can check on the marketplace and docs updates!

exigow:
@taylorsperry,

> I love the idea of enabling users to send us error reports with logs that help us debug. Would it be possible for those reports to automatically create issues that have bug labels so they make it onto our board? Or what's the flow that you imagine?

Correct! Errors will be appropriately tagged and will have the prefix `bug: ` according to our convention.

PR: https://github.com/sourcegraph/jetbrains/pull/296

> I can check on the marketplace and docs updates!

Thanks! Looks like links in docs are already updated by @kalanchan so we only need to update `Issue Tracker`. Issue Tracker link can be edited on marketplace if you have admin rights: https://plugins.jetbrains.com/plugin/9682-sourcegraph-cody--code-search

To clarify, this link currently points to `https://github.com/sourcegraph/sourcegraph/issues/new?labels=team/integrations,jetbrains-ide&title=JetBrains:+` which is **old JetBrains repository**. It should point to new repository: `https://github.com/sourcegraph/jetbrains/issues/new`.

taylorsperry:
Closing as done

PR #296, opened 2024-01-12
Add "Open GitHub Issue" error submitter

exigow:
Unplanned, but I've made decision to add it, especially since implementing it was a small effort. Convenient issue reporting will help us receive more exceptions, so we'll be able to react to them by Feb GA and increase the discoverability of the project. Maybe someone will be engaged enough to participate in our open source project, who knows?

Related to: #135

## Test plan

1. Add `throw RuntimeException()` somewhere in the plugin code to simulate exception.
2. Run the plugin and trigger that exception. An error notification will appear. 
3. Open notification, then click `Open GitHub Issue`.
4. Browser will open at the `issues/new` with useful context (plugin version, IDE version, stacktrace, etc). You should be able to submit that issue (but you shouldn't do it if the exception is only for testing purposes!!).

## Demo

Before:

Selection_038

After: 

Selection_035
Selection_039

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/error/CodyError.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/error/CodyError.kt
@@ -1,0 +1,9 @@
+package com.sourcegraph.cody.error
+
+data class CodyError(
+    val title: String?,
+    val pluginVersion: String?,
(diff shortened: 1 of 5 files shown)


pkukielka (approved):
LGTM

PR #297, opened 2024-01-12
Create clickable links from the listed context files

### LINKED ISSUE #255: Make contextFiles a clickable link

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/255

## Test plan 
- `:runIde`
- ask Cody about the code / run explain command 
- see the context file
- text the links

image

Note that we have issue with unrelated files shown in the context read files list: https://github.com/sourcegraph/jetbrains/issues/298


src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
```diff
@@ -0,0 +1,73 @@
...
+    this.layout = BorderLayout()
+
+    val margin = JBInsets.create(Insets(TEXT_MARGIN, TEXT_MARGIN, TEXT_MARGIN, TEXT_MARGIN))
+    val contextFileNames =
+        contextMessages
+            .stream()
+            .map<ContextFile>(ContextMessage::file)
+            .filter { obj: ContextFile? -> Objects.nonNull(obj) }
+            .map(ContextFile::fileName)
+            .collect(Collectors.toSet())
```
exigow: We can remove Java stream & collect.

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
```diff
@@ -0,0 +1,73 @@
...
+    contextFileNames.forEach { fileName: String ->
+      val filePanel = createFileWithLinkPanel(project, fileName)
+      accordionSection.contentPanel.add(filePanel, fileIndex.getAndIncrement())
+    }
+    add(accordionSection, BorderLayout.CENTER)
+  }
+
+  private fun createFileWithLinkPanel(project: Project, fileName: String): JPanel {
+    val fileWithoutRedundantPrefix =
+        fileName.removePrefix("../../../..").removePrefix(project.baseDir.path)
```
exigow: Can you pls elaborate on this `removePrefix("../../../..")`?  It looks like a underlying problem. Are we getting such paths from the agent?

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
exigow: plus use filterNotNull + toSet()

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
```diff
@@ -0,0 +1,73 @@
...
+    val anAction =
+        object : DumbAwareAction() {
+          override fun actionPerformed(anActionEvent: AnActionEvent) {
+            val file = project.baseDir.findFileByRelativePath(fileWithoutRedundantPrefix)
```
exigow: Looks like `com.intellij.openapi.project.Project.getBaseDir()` is deprecated.

Maybe `getBasePath()` is better alternative?

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
```diff
@@ -0,0 +1,73 @@
...
+            logger.info("Opening a file from the used context (fileName=$fileName file=$file)")
+            if (file != null) {
+              FileEditorManager.getInstance(project).openFile(file, /*focusEditor=*/ true)
```
exigow: IMO we can ~remove this `/*focusEditor=*/` or~ replace with `val focusEditor = true`.

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
```diff
@@ -0,0 +1,73 @@
...
+            .stream()
+            .map<ContextFile>(ContextMessage::file)
+            .filter { obj: ContextFile? -> Objects.nonNull(obj) }
+            .map(ContextFile::fileName)
+            .collect(Collectors.toSet())
+
+    val accordionSection = AccordionSection("Read ${contextFileNames.size} files")
+    accordionSection.isOpaque = false
+    accordionSection.border = EmptyBorder(margin)
+    val fileIndex = AtomicInteger(0)
```
exigow: Suggestion: `AtomicInteger` -> `Int`

(code is synchronous)

exigow: Or even better: `forEachIndexed`! :smile:

exigow (changes requested):
Looks good! :+1: I've requested some small changes (see: comments).

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: what can I do instead?

mkondratek: got it, act

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: We cannot do any better here.

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: ok

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: nvm, I was using a wrong repo as a context

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: thanks!

pkukielka: Wasn't `findFileByRelativePath` one of the methods which sometimes were taking longer than expected?
If that is the case, remember we should use the same constraints for the listener callbacks and actions as we use for UI threads - this means no long running actions. Just FYI.

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
pkukielka: That looks weird. Why exactly `../../../..`?

exigow (approved):
Thanks for the fixes.

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: or... there is a problem 🤔

mkondratek: image

installed on my IDE & run in jetbrains project

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: nah, it was `getRelativePathToProjectRoot`. but I will take a close look 🕵️

mkondratek: fixed

src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
mkondratek: nah, I probably had the former bad version or bad context repo - it works properly now.

there is another problem, however - I will start a discussion below

mkondratek:
So the remaining unknown seems to be the remote files. Take a look at the agent's response:

image

It contains a doubled entry of the local file:
```
{
  "fileName": "/Users/mkondratek/IdeaProjects/runIde/jetbrains/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt"
},
```
and two other entries with repo & revision:
```
{
  "fileName": "src/main/kotlin/com/sourcegraph/cody/chat/CodyOnboardingGuidancePanel.kt",
  "repoName": "github.com/sourcegraph/jetbrains",
  "revision": "14891e64eb192789df371dd43ae819a77aa22cd6"
},
{
  "fileName": "src/main/java/com/sourcegraph/cody/CodyToolWindowContent.java",
  "repoName": "github.com/sourcegraph/jetbrains",
  "revision": "14891e64eb192789df371dd43ae819a77aa22cd6"
},
```

I assume that the file referencing the local file should link it directly. 

What about the "remote files"? We can link the local files to them BUT there can be inconsistencies b/w local and remote (eg. the agent uses a file that exists in the remote but it does not exist locally). How should we handle it?

gradle.properties
```diff
@@ -10,12 +10,12 @@ pluginSinceBuild=221.1
 pluginUntilBuild=*
 # IntelliJ Platform Properties -> https://github.com/JetBrains/gradle-intellij-plugin#intellij-platform-properties
 platformType=IC
-platformVersion=2022.1
+platformVersion=2023.3.1
```
pkukielka: Rollback this please!

pkukielka (changes requested):
Please do not bump IJ platform version!

exigow:
@danielmarquespt, we haven't consulted this change with you so please take a look on attached demo in first comment & eventually confirm. The design has slightly changed (paths now have "link" color + no border/outline).

Before:
image

mkondratek:
it should be all good now

mkondratek:
> @danielmarquespt, we haven't consulted this change with you so please take a look on attached demo in first comment & eventually confirm. The design has slightly changed (paths now have "link" color + no border/outline).
> 
> Before: image

image

pkukielka (approved):
LGTM

PR #303, opened 2024-01-15
Reimplement agent lifecycle management

pkukielka:
## Test plan

This PR will require extensive testing of all existing features.
We need to go through whole https://github.com/sourcegraph/jetbrains/blob/main/TESTING.md
Also ideally use this changes locally for some time to try cach all kind of timing/lifecycle bugs.

## Changes

Agent lifecycle is now managed fully by `CodyAgentService`.

Classes using agent DO NOT try to validate it state, restart, wait few seconds before doing some action, etc.
They also not do any assumptions about agent state anymore.
All they know/get is a `CompletableFuture` which will contain the agent when it's ready.

If agent will loose connection or crash in some other way it should be restarted automatically by `CodyAgentService`.

Other than that lifetime management and accesability of several objects components (like `CodyAgentCodebase` or server object) were changed.

Also, maybe not a huge win but:
 30 files changed, 576 insertions(+), **659** deletions(-)
:)


src/main/java/com/sourcegraph/config/ApiAuthenticator.java
```diff
@@ -1,20 +0,0 @@
-package com.sourcegraph.config;
```
pkukielka, line 1: It was not used in any place.

src/main/java/com/sourcegraph/cody/CodyFileEditorListener.java
```diff
@@ -1,78 +1,50 @@
...
-      return;
-    }
-
-    ApplicationManager.getApplication()
-        .executeOnPooledThread(
-            () -> {
-              CodyAgent.getInitializedServer(source.getProject())
-                  // The timeout has been increased from 3 to 12.
-                  // This more like workaround than a fix to:
-                  // https://github.com/sourcegraph/jetbrains/issues/169
```
mkondratek, line 37: I assume we can say that this PR actually fixes: https://github.com/sourcegraph/jetbrains/issues/169

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
```diff
@@ -25,18 +24,11 @@ class AccountSettingChangeListener(project: Project) : ChangeListener(project) {
...
-              // Stopping the agent is idempotent, so it's OK if we call stopAgent multiple times.
-              CodyAgentManager.stopAgent(project)
-            }
-
             // Notify Cody Agent about config changes.
-            val agentServer = CodyAgent.getServer(project)
-            if (ConfigUtil.isCodyEnabled() && agentServer != null) {
-              agentServer.configurationDidChange(ConfigUtil.getAgentConfiguration(project))
+            CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
+              if (ConfigUtil.isCodyEnabled()) {
```
mkondratek, line 29: so previously depending on this boolean we were starting or stopping the agent. Is the behavior different now? so is the agent always started?

src/main/kotlin/com/sourcegraph/common/UpgradeToCodyProNotification.kt
```diff
@@ -85,12 +85,14 @@ private constructor(title: String, content: String, shouldShowUpgradeOption: Boo
...
-        val server = CodyAgent.getServer(project)
-        if (server != null) {
-          val codyProFeatureFlag = server.evaluateFeatureFlag(GetFeatureFlag("CodyProJetBrains"))
-          return codyProFeatureFlag.get()!!
-        }
+        return CodyAgentService.getAgent(project)
+            .thenCompose { agent ->
+              agent.server.evaluateFeatureFlag(GetFeatureFlag("CodyProJetBrains"))
+            }
+            // TODO We should avoid using getNow if possible
```
mkondratek, line 92: can we have an issue for this one? or does it make sense to just add `timeout(1, SECONDS)`?

src/main/kotlin/com/sourcegraph/cody/context/EmbeddingStatusView.kt
```diff
@@ -83,21 +80,23 @@ class EmbeddingStatusView(private val project: Project) : JPanel() {
...
-    val icon = embeddingStatus.getIcon()
-    if (icon != null) {
-      embeddingStatusContent.icon = icon
-      embeddingStatusContent.preferredSize =
-          Dimension(icon.iconWidth + 10, embeddingStatusContent.height)
-    }
-    val tooltip = embeddingStatus.getTooltip(project)
-    if (tooltip.isNotEmpty()) {
-      embeddingStatusContent.setToolTipText(tooltip)
+    ApplicationManager.getApplication().invokeLater {
```
mkondratek, line 83: I see all the logic is wrapped in `invokeLater`. Shouldn't be a caller's responsibility to call it on a proper thread? We could then mark it as `@RequiresEdt`.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -0,0 +1,103 @@
...
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.project.Project
+import com.sourcegraph.cody.statusbar.CodyAutocompleteStatusService
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+import java.util.function.Consumer
+import javax.annotation.concurrent.GuardedBy
+
+@Service(Service.Level.APP)
+class CodyAgentService : Disposable {
```
mkondratek, line 17: beautiful 😍

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -146,14 +144,20 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
...
         ApplicationManager.getApplication().invokeLater {
           promptPanel.textArea.isEnabled = true
           promptPanel.textArea.emptyText.text = "Ask a question about this code..."
         }
         callback.invoke()
+      } catch (e: ExecutionException) {
+        // Agent cannot gracefully recover when connection is lost, we need to restart it
+        logger.warn("Failed to load new chat, restarting agent", e)
+        CodyAgentService.getInstance(project).restartAgent(project)
+        Thread.sleep(5000)
```
mkondratek, line 160: is it needed to sleep here? `startAgent` has it's 15s timeout already ⌚

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -222,13 +226,9 @@ class CodyToolWindowContent(private val project: Project) : UpdatableChat {
   }
 
   private fun loadCommands() {
-    tryRestartingAgentIfNotRunning(project)
-    getInitializedServer(project).thenAccept { server ->
-      if (server == null) {
-        setRecipesPanelError()
-      }
+    CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
```
mkondratek, line 232: I guess we can simplify the code here even more. Let's stick to the rule that it is the caller's responsibility to use a proper thread, hence:
1. remove `CodyAgentService.applyAgentOnBackgroundThread(project) { ... }` wrapping here
2. remove `ApplicationManager.getApplication().executeOnPooledThread { ... }` wrapping above (line 225, in `refreshRecipes`)
3. revamp `loadCommands` so it takes `agent` as a param
3. wrap `loadCommands(agent)` into `CodyAgentService.applyAgentOnBackgroundThread(project) { ... }` in line 225, `refeshRecipes`)

src/main/kotlin/com/sourcegraph/cody/chat/Chat.kt
```diff
@@ -32,106 +32,93 @@ class Chat {
...
+      val isFirstMessage = AtomicBoolean(false)
+      agent.client.onFinishedProcessing = Runnable { chat.finishMessageProcessing() }
+      agent.client.onChatUpdateMessageInProgress = Consumer { agentChatMessage ->
+        val agentChatMessageText = agentChatMessage.text ?: return@Consumer
+        val chatMessage =
+            ChatMessage(Speaker.ASSISTANT, agentChatMessageText, agentChatMessage.displayText)
+        if (isFirstMessage.compareAndSet(false, true)) {
+          val contextMessages =
+              agentChatMessage.contextFiles
+                  ?.stream()
```
exigow: We can remove stream + collect

src/main/java/com/sourcegraph/cody/CodyFileEditorListener.java
pkukielka: There is a good chance for that.
Fas there any deterministic reproduction for the issue above?

src/main/kotlin/com/sourcegraph/common/UpgradeToCodyProNotification.kt
pkukielka: Timeout does not change anything there, because there is no quarantee if it will take 0, 2 or 10 sec.
I will create an issue.

src/main/java/com/sourcegraph/cody/CodyFileEditorListener.java
mkondratek: I am afraid there was not 😕 I could have been related to the network

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
pkukielka: I wanted to avoid constant heavier utilisation of the CPU in case of the lack of the network.
In fact it would be better to have some listener which would say if network is back, I will check if we can do that. I will also move that code to agent service.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
pkukielka: 1. I agree that we can simplify this code a bit
2. I do not agree with `Let's stick to the rule that it is the caller's responsibility to use a proper thread` as a general rule, it can lead to too many errors/omissions. That is, unless we explicitly wrap methods bory in something like CompletableFuture, but with a distinct subtypes for EDT and Background (which I think could work nicely but it's a bit additional work and I'm not sure if it's currently worth it).
3. I'm happy to discuss this on the standup or a dedicated call.

mkondratek: ok, feel free to do it your way

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
pkukielka: That is something I take liberty to remove but we have to test it. Agent is always started (or if not yet `CodyAgentService.applyAgentOnBackgroundThread` will wait for it to be).
I think starting/stoping should take place when cody settings change, as that is where you can enable/disable cody (and we have proper code  for that in settings).

But if someone see a reason why that code was needed there I will be happy to hear it.

src/main/kotlin/com/sourcegraph/cody/context/EmbeddingStatusView.kt
pkukielka: I have mixing feelings about that, let's discuss on the next standup.
I think this can go as it is for now.

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -1,225 +1,147 @@
...
-
-    val processBuilder = ProcessBuilder(command)
-    if (java.lang.Boolean.getBoolean("cody.accept-non-trusted-certificates-automatically") ||
-        ConfigUtil.getShouldAcceptNonTrustedCertificatesAutomatically()) {
-      processBuilder.environment()["NODE_TLS_REJECT_UNAUTHORIZED"] = "0"
+class CodyAgent
+private constructor(
+    val client: CodyAgentClient,
+    val server: CodyAgentServer,
+    val launcher: Launcher<CodyAgentServer>,
```
exigow, line 31: This lanucher is unused - can we remove that?

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
```diff
@@ -9,19 +11,18 @@ import com.sourcegraph.common.ProjectFileUtils
 import com.sourcegraph.config.ConfigUtil
 import com.sourcegraph.vcs.RepoUtil
 
-class CodyAgentCodebase(private val underlying: CodyAgentServer, val project: Project) {
+@Service(Service.Level.PROJECT)
+class CodyAgentCodebase(val project: Project) {
+
+  fun initializeRepoName() {
+    onRepositoryNameChange(RepoUtil.findRepositoryName(project, null))
+  }
```
exigow, line 21: This method should be placed below class fields.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentCodebase.kt
```diff
@@ -9,19 +11,18 @@ import com.sourcegraph.common.ProjectFileUtils
 import com.sourcegraph.config.ConfigUtil
 import com.sourcegraph.vcs.RepoUtil
 
-class CodyAgentCodebase(private val underlying: CodyAgentServer, val project: Project) {
+@Service(Service.Level.PROJECT)
+class CodyAgentCodebase(val project: Project) {
```
exigow, line 15: Project should be private.

exigow (approved):
Looks good :+1:.

I've added some minor comments and I'm looking forward for Cody commit bump to start testing. I understand the idea of these changes, especially after your presentation. I like this CodyAgentService which really simplifies usages (as confirmed by the net number of lines) and and forces us to think more about threading.

Sa1to (approved):
LGTM 🎉 FYI rebased https://github.com/sourcegraph/jetbrains/pull/183 on this branch

mkondratek (approved):

PR #308, opened 2024-01-16
Update codyCommit

mkondratek:
## Test plan 
- test all

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
@@ -1,7 +1,26 @@
 package com.sourcegraph.cody.agent.protocol
 
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import java.lang.reflect.Type
+import java.net.URI
+
 data class ContextFile(
-    val fileName: String,
+    val uri: URI,
     val repoName: String?,
     val revision: String?,
 )
+
+val contextFileDeserializer =
+    JsonDeserializer { jsonElement: JsonElement, _: Type, _: JsonDeserializationContext ->
+      val jsonObject = jsonElement.asJsonObject
+
+      val uriObj = jsonObject["uri"].asJsonObject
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
```diff
@@ -1,7 +1,26 @@
...
 
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import java.lang.reflect.Type
+import java.net.URI
+
 data class ContextFile(
-    val fileName: String,
+    val uri: URI,
```
mkondratek, line 10: [API changed](https://github.com/sourcegraph/cody/pull/2653).

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
```diff
@@ -1,7 +1,26 @@
...
     val repoName: String?,
     val revision: String?,
 )
+
+val contextFileDeserializer =
```
mkondratek, line 15: For some reason, deserialization did not work out of the box.

exigow: IMO, we shouldn't do that in this way. Are these deserialization issues caused by the String->URI change or something else has been changed/introduced?

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
```diff
@@ -1,7 +1,26 @@
...
+import com.google.gson.JsonElement
+import java.lang.reflect.Type
+import java.net.URI
+
 data class ContextFile(
-    val fileName: String,
+    val uri: URI,
     val repoName: String?,
     val revision: String?,
 )
```
exigow, line 13: We should test this String->URI change on Windows. Recently, there were issues because the agent has a URI in a different RFC than those built-in in Java.

pkukielka (approved):
LGTM

PR #313, opened 2024-01-18
Fix context files presentation on Windows

mkondratek:
I should have tested it in the scope of: https://github.com/sourcegraph/jetbrains/pull/297. Anyway, it fixes the links in Windows.

## Test plan 
- [x] Test on Windows + MacOS
- [x] run command/ prompt chat & see the files
- [x] files should be clickable and navigate correctly

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
@@ -22,7 +22,7 @@ val contextFileDeserializer =
               uriObj["scheme"]?.asString,
               uriObj["host"]?.asString,
               uriObj["path"]?.asString,
-              uriObj["fragment"]?.asString)
+              /* fragment= */ null)
       val repoName = jsonObject["repoName"]?.asString
       val revision = jsonObject["revision"]?.asString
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/ContextFilesMessage.kt
```diff
@@ -35,7 +36,7 @@ class ContextFilesMessage(val project: Project, contextMessages: List<ContextMes
               if (it.repoName != null) {
                 "${project.basePath}/${it.uri.path}"
               } else {
-                it.uri.path
+                Paths.get(it.uri).absolutePathString()
```
mkondratek, line 39: so that macos & windows paths are resolved correctly (thanks @pkukielka for this idea)

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
```diff
@@ -22,7 +22,7 @@ val contextFileDeserializer =
               uriObj["scheme"]?.asString,
               uriObj["host"]?.asString,
               uriObj["path"]?.asString,
-              uriObj["fragment"]?.asString)
+              /* fragment= */ null)
```
mkondratek, line 25: redundant for us + it breaks `Paths.get`

pkukielka (approved):
LGTM

PR #318, opened 2024-01-19
IntelliJ reacts to attribution enable/disable from site config

### LINKED ISSUE #59427: Guardrails can be turned on/off from enterprise instance
```[tasklist]
### Tasks
- [x] Backend
- [x] VS Code - chat
- [ ] IntelliJ - chat
- [ ] Autocomplete
```

cbart:
Closes sourcegraph/sourcegraph#59427

This change adds `CurrentConfigFeatures` singleton/observer that allows the UI to react on whether attribution is turned on or off.

Background:

- Attribution search ([RFC WIP](https://docs.google.com/document/d/1zSxFDQPxZcn5b6yKx40etpJayoibVzj_Gnugzln1weI/edit#heading=h.trqab8y0kufp), [PRD](https://docs.google.com/document/d/1c3CLC7ICDaG63NOWjO6zWm-UElwuXkFrKbCKTw-7H6Q/edit#heading=h.fsopaopl4q9u)) is performing a dotcom search on a snippet of code generated by Cody
- The goal is to inform the user that _this code was found in the open source_.
- For chat, this looks like an indicator - we show the user that search is running and then
  - either it finds nothing - then we signal there is no risk
  - or it finds the snippet - then we signal where the snippet was found in tooltip
  - or it cannot finish the search - then we show attribution is not available

## TODO
- [x] unit tests!!!
- [ ] green SCIP CI

## Test plan

- TODO
- Corresponding VS Code tests
- Manual
- [Loom](https://www.loom.com/share/cf541aa3fb734390a9163a46a4dde447?sid=a16c6d92-6e51-409b-b7c8-80a6a92a4c3c)


src/main/java/com/sourcegraph/cody/agent/CurrentConfigFeatures.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+public class CurrentConfigFeatures implements Consumer<ConfigFeatures> {
+
+    public static CurrentConfigFeatures INSTANCE = new CurrentConfigFeatures(new ConfigFeatures(true, false));
```
pkukielka: FYI Idiomatic way in IntelliJ to create instance is using project or app level services. 
If you will search for `@Service` you will find some examples.
Difference between static instance like yours, and services, is that lifetime and initialisation order/time of services is managed by IJ and allow you to avoid some pitfalls connected to it. You also get some additional benefits like if you inherit from `Disposable` and override `dispose` method you can ensure proper cleaning in the right moment.
It might not make too big difference in your case, just saying :)

src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
```diff
@@ -139,14 +146,70 @@ public void mouseExited(@NotNull EditorMouseEvent event) {
...
     CodeEditorPart codeEditorPart = new CodeEditorPart(layeredEditorPane, editor);
     codeEditorPart.updateLanguage(language);
     return codeEditorPart;
   }
 
+  @Nullable private TransparentButton attributionIndicator() {
+    AttributionButton guardrailsButton = new AttributionButton("Attribution search running...");
+    guardrailsButton.setEnabled(false);
+    // TODO ACTUALLY GET ATTRIBUTION
+    CompletableFuture.runAsync(() -> {
```
pkukielka: `runAsync` is sometimes problematic, because it uses current executor for spawning new thread, and it's not always clear what that will be as in IJ there are multiple various threads types/executors mixed together. Idiomatically we are trying to always use `ApplicationManager.getApplication().invokeLater` if all we need is just a background thread.

src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
```diff
@@ -139,14 +146,70 @@ public void mouseExited(@NotNull EditorMouseEvent event) {
...
+      try {
+        Thread.sleep(5000);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
+    }).thenAccept((v) -> {
+      guardrailsButton.setText("No attribution");
```
pkukielka: It's not allowed to do UI changes on the background threads (one of the IJ development pitfals).
You need to switch to dispatch thread (also called EDT) to do that operation:

```kotlin
    guardrailsButton.setText("No attribution");
}
```

Very useful read on that topics (kind of 'must have' for IJ devs in my opinion):
https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html

src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
```diff
@@ -139,14 +146,70 @@ public void mouseExited(@NotNull EditorMouseEvent event) {
...
     return codeEditorPart;
   }
 
+  @Nullable private TransparentButton attributionIndicator() {
+    AttributionButton guardrailsButton = new AttributionButton("Attribution search running...");
+    guardrailsButton.setEnabled(false);
+    // TODO ACTUALLY GET ATTRIBUTION
+    CompletableFuture.runAsync(() -> {
+      try {
+        Thread.sleep(5000);
```
pkukielka: How it works?
Looking at the code it seems something should interrupt this thread if attribution is found, and if not we want to change text after 5 sec?
Where does that interruption come from?

cbart: Apologies for confusion! This is just a placeholder to demonstrate the behavior. For reals we'll invoke a command to the extension here, and wait for the result to update the UI.

I'm going to work on that next.

src/main/java/com/sourcegraph/cody/agent/CurrentConfigFeatures.java
cbart: Done!

src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
cbart: I will leave an `invokeLater` and will send a follow-up PR with implementation of actually fetching attribution.

src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
cbart: I've left an `invokeLater` there. In a follow-up PR, I'll actually implement it with [Olaf's advice](https://sourcegraph.slack.com/archives/C059N5FRYG3/p1705655269541249?thread_ts=1705588365.312009&cid=C059N5FRYG3).

cbart:
All comments addressed, please take another look! This should be good to review now.

src/main/java/com/sourcegraph/cody/agent/CurrentConfigFeatures.java
```diff
@@ -0,0 +1,100 @@
...
+  public void update(ConfigFeatures configFeatures) {
+    this.features.set(configFeatures);
+    observers.forEach((observer) -> observer.update(configFeatures));
+  }
+
+  /**
+   * Given listener will be given new {@link ConfigFeatures} whenever they arrive. Observation
+   * relationship is ended once the returned cleanup {@link Runnable} is {@link Runnable#run()}.
+   */
+  public Runnable attach(ConfigFeaturesObserver observer) {
```
pkukielka: Minor thing, but you could use `CancellationToken` here.

```kotlin
val cancellationToken = CancellationToken()
cancellationToken.onCancel { observers.remove(id) }
return cancellationToken
```

A bit more code but I think it is a bit more clear regarding the intent.

cbart: Done!

pkukielka (approved):
LGTM

PR #326, opened 2024-01-22
Add support for local symf context

### LINKED ISSUE #259: Implement UI for managing local symf context and triggering the refresh

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/259

## Test plan

Manual testing using full QA guide.

Other than that:
1) Enable context
2) Ask repo-specific question
3) Check if response is based on the understanding of the local context

4) Disable context
5) Ask repo-specific question
6) Check if response is lacking previous insight

7) Click on refresh button and check if reindexing is in progress
8) Click on help button and check if it opens proper documentation page


src/main/resources/CodyBundle.properties
```diff
@@ -44,30 +37,33 @@ chat.rate-limit-error.upgrade=\
...
   <html>\
     You''ve used all{0} autocomplete suggestions for the month. \
     Upgrade to Cody Pro for unlimited autocompletes, chats, and commands.<br><br>\
     (Already upgraded to Pro? Restart your IDE for changes to take effect)\
   </html>
 UpgradeToCodyProNotification.content.explain=To ensure that Cody can stay operational for all Cody users, please come back tomorrow for more chats, commands, and autocompletes.
-
 tab.subscription.already-pro=(Already upgraded to Pro? Restart your IDE for changes to take effect)
+context-panel.panel-name="Chat Context"
+context-panel.reindex-button-name=""Trigger Reindexing""
```
exigow: double quote

exigow: Plus "Chat Context" is also quoted. These quotes are visible in UI.

exigow: This issue concerns other labels (see: "1 local project" is also quoted):

image

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
```diff
@@ -36,13 +34,18 @@ class AccountSettingChangeListener(project: Project) : ChangeListener(project) {
...
-                embeddingStatusView.updateEmbeddingStatus()
               }
 
               UpgradeToCodyProNotification.autocompleteRateLimitError.set(null)
               UpgradeToCodyProNotification.chatRateLimitError.set(null)
               CodyAutocompleteStatusService.resetApplication(project)
+
               ApplicationManager.getApplication().executeOnPooledThread { refreshSubscriptionTab() }
+
+              ApplicationManager.getApplication().executeOnPooledThread {
```
mkondratek: any reason to run it on a separate pooled thread than `refreshSubscriptionTab()` above?

src/main/kotlin/com/sourcegraph/cody/context/ui/ContextRepositoriesCheckboxRenderer.kt
```diff
@@ -0,0 +1,35 @@
...
+      tree: JTree?,
+      value: Any?,
+      selected: Boolean,
+      expanded: Boolean,
+      leaf: Boolean,
+      row: Int,
+      hasFocus: Boolean
+  ) {
+    when (value) {
+      is CheckedTreeNode -> {
```
exigow, line 21: This `when` has only one `is CheckedTreeNode` condition. Are we planning to expand it? Because if not, I think we should simplify it and make a simple `if`.

src/main/resources/CodyBundle.properties
pkukielka: Good point, I will fix it!

src/main/kotlin/com/sourcegraph/cody/context/ui/ContextRepositoriesCheckboxRenderer.kt
```diff
@@ -0,0 +1,35 @@
...
+        when (val userObject = value.userObject) {
+          is Project -> {
+            textRenderer.appendHTML(
+                "<b>${userObject.name}</b> - <i>${userObject.basePath}</i>",
```
mkondratek, line 25: - [x] we will need to verify it on the latest supported version of the IDE

src/main/resources/CodyBundle.properties
```diff
@@ -44,30 +37,33 @@ chat.rate-limit-error.upgrade=\
...
+context-panel.in-progress="Running Cody 'Keyword Search' indexer..."
```
mkondratek: + pls double check the rendering of `'` - sometimes it is needed to write `''` to see a single `'`

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -0,0 +1,128 @@
...
+        .createPanel()
+  }
+
+  init {
+    layout = VerticalFlowLayout(VerticalFlowLayout.BOTTOM, 14, 0, true, false)
+    tree.expandRow(0)
+
+    add(toolbarPanel)
+  }
+}
```
exigow, line 67: It's a bit hard to figure out what the execution order in this class. It looks like it was written backwards. :smile:

My proposals:
* Move init to top (as is should be). 
* Classes here don't share states so my suggestion is to replace anonymous classes with private static classes like "RefreshButton()", "HelpButton()". This inner classes should be places at bottom.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -0,0 +1,128 @@
...
+          }
+        }
+
+        override fun isEnabled(): Boolean {
+          return !isReindexingInProgress.get()
+        }
+
+        override fun isDumbAware(): Boolean {
+          return true
+        }
```
exigow: Simple expressions like this can be shortened to smth like `override fun isDumbAware() = true`.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -0,0 +1,128 @@
...
+            isEnhancedContextEnabled.set(checked)
+            CodyAgentService.getInstance(project).restartAgent(project)
+          }
+        }
+    localProject.add(currentRepo)
+    treeRoot.add(localProject)
+    CheckboxTree(ContextRepositoriesCheckboxRenderer(), treeRoot)
+  }
+
+  private val toolbarPanel = run {
```
exigow, line 44: Why these variables (toolbarPanel, tree) are initialized within `run` block? Can we use simple, plain functions here? Like `private fun createToolbarPanel() {...}`).

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
pkukielka: Yes thank you! This is a merge error. It was supposed to be `invokeLater`!

src/main/kotlin/com/sourcegraph/cody/context/ui/ContextRepositoriesCheckboxRenderer.kt
pkukielka: We will be changing it soon when we will start supporting multi-repo for enterprise.
Whole model will be a little bit richer most likely. 
I will fix rewrite it then (in a couple of days max :))

src/main/resources/CodyBundle.properties
pkukielka: Already fixed, thanks.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
pkukielka: What you propose is actually nt possible :)
Order of execution/creation is exactly as written.
Because of that `init` cannot be moved to the beginning as it would then operate on not-yey-existing-variables.
Classes I can move.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
pkukielka: We could. Run block is basically static initializer.
So e.g. in scala it would look like that:

```
val x = {
   val a = 1
   val b = 2
   a + b
}
// x was initialised to 3
```

I kind of intentionally make them initialised statically to show their creation is not context dependent or time sensitive. When you see a function you cannot reason so much about it.

exigow: > I kind of intentionally make them initialised statically to show their creation is not context dependent or time sensitive.

Good point!

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
exigow: Thanks for explanation. I still believe that replacing anonymous classes with private static inner classes will be a better solution. Ofc I leave the decision up to you.

src/main/kotlin/com/sourcegraph/cody/context/ui/ContextRepositoriesCheckboxRenderer.kt
exigow: Thanks!

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
pkukielka: Actually I removed it altogether, it is not needed anymore after I discussed some design changes with Olaf.

src/main/kotlin/com/sourcegraph/cody/context/ui/ContextRepositoriesCheckboxRenderer.kt
pkukielka: Looks fine on `2023.1.1`, `2023.2.2` and `2023.3.2`.
`2022.1` is fine as well.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
pkukielka: I moved them to the separate classes.
If we want to make them finer classes I'm fine with that as well, I can move it later.

exigow (changes requested):
Can you please share some ideas for repo-specific questions? I'm trying to ask for smth like 'show me the class where service X is used' or 'list the classes that are written in Kotlin' or 'show me usages of X', but I'm missing something.

Also: I think that we should update testing guide with this feature ("Ask repo-specific question with local-only context").

mkondratek (approved):
lgtm

exigow (approved):
From our private conversation, it emerged that to properly test, we must ask a precise question like "what is the first line of the CodyEditorFactoryListener class?". I was able to correctly test this feature with this tip.

However, I have two small topics:

* We should bump the version to the Cody commit if this feature requires the latest version. I had to manually bump the version, and it was not obvious to me. It seems to me that developers should not be expected to work on CODY_DIR that is built on the latest main, especially since we often have to go back in history. This causes some conflicts on merge, but this can be simply fixed.
* We should add to `testing.md` information that to objectively test this feature, we must specify the question in precise way, e.g., "what is the first line of the CodyEditorFactoryListener class?". This also was not obvious to me.

Looks good! :+1:

exigow:
I've spotted some bug: if you click refresh while a response is still being generated, an error `java.util.concurrent.CancellationException` occurs. After this error, there are different problems with the chat.

Issue #344, opened 2024-01-24
bug: Non-EDT thread exception in new enhanced context panel


steveyegge:
### Installation Information

Latest nightly as of 1/24.

### Describe the bug

com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: Access is allowed from Event Dispatch Thread (EDT) only; see https://jb.gg/ij-platform-threading for details
Current thread: Thread[ApplicationImpl pooled thread 41,4,main] 2110361532 (EventQueue.isDispatchThread()=false)
SystemEventQueueThread: Thread[AWT-EventQueue-0,6,main] 122581038
	at com.intellij.openapi.application.impl.ApplicationImpl.createThreadAccessException(ApplicationImpl.java:1083)
	at com.intellij.openapi.application.impl.ApplicationImpl.throwThreadAccessException(ApplicationImpl.java:1078)
	at com.intellij.openapi.application.impl.ApplicationImpl.assertIsDispatchThread(ApplicationImpl.java:1066)
	at com.intellij.ui.SpeedSearchBase.isPopupActive(SpeedSearchBase.java:210)
	at com.intellij.ui.speedSearch.SpeedSearchSupply.getSupply(SpeedSearchSupply.java:55)
	at com.intellij.ui.speedSearch.SpeedSearchSupply.getSupply(SpeedSearchSupply.java:44)
	at com.intellij.ui.speedSearch.SpeedSearchUtil.applySpeedSearchHighlightingFiltered(SpeedSearchUtil.java:185)
	at com.intellij.ui.ColoredTreeCellRenderer.rendererComponentInner(ColoredTreeCellRenderer.java:129)
	at com.intellij.ui.ColoredTreeCellRenderer.getTreeCellRendererComponent(ColoredTreeCellRenderer.java:58)
	at com.intellij.ui.CheckboxTreeBase$CheckboxTreeCellRendererBase.getTreeCellRendererComponent(CheckboxTreeBase.java:164)
	at com.intellij.ui.tree.ui.DefaultTreeUI.getRenderer(DefaultTreeUI.java:189)
	at com.intellij.ui.tree.ui.DefaultTreeUI$2.getNodeDimensions(DefaultTreeUI.java:582)
	at java.desktop/javax.swing.tree.AbstractLayoutCache.getNodeDimensions(AbstractLayoutCache.java:497)
[...]

PR #345, opened 2024-01-24
Expand context panel tree on EDT

steveyegge:
This is a fix for [the recent exception I'm seeing in nightly.](https://github.com/sourcegraph/jetbrains/issues/344)

I unfortunately was not able to repro with a local build. So I'm not sure how it's getting called from a non-EDT thread -- some sort of race condition, possibly. But this looks like the right fix -- I tried it with/without, and the new context panel expands the tree properly for me in both cases.

## Test plan

Manual testing. Passed.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
@@ -1,6 +1,7 @@
 package com.sourcegraph.cody.context.ui
 
 import com.intellij.openapi.actionSystem.ActionToolbarPosition
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.VerticalFlowLayout
 import com.intellij.ui.CheckboxTree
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -60,8 +61,12 @@ class EnhancedContextPanel(private val project: Project) : JPanel() {
 
   init {
     layout = VerticalFlowLayout(VerticalFlowLayout.BOTTOM, 14, 0, true, false)
-    tree.expandRow(0)
 
-    add(toolbarPanel)
+    // Fix for https://github.com/sourcegraph/jetbrains/issues/344
```
pkukielka, line 65: I think that fix is right.
Looks like `EnhancedContextPanel ` is created in `ChatPanel` constructor and `ChatPanel` in turn in the `CodyToolWindowContent`. 

`CodyToolWindowContent` is a IJ project-level service object, and its lifecycle is controlled by IJ. 
I found this info in IJ [documentation](https://plugins.jetbrains.com/docs/intellij/plugin-services.html#retrieving-a-service):

> Getting a service doesn't need a read action and can be performed from any thread. If a service is requested from several threads, it will be initialised in the first thread, and other threads will be blocked until it is fully initialised.

So you are probably right, it's a race. We need to be on a safe side there.

pkukielka (approved):
LGTM

Issue #350, opened 2024-01-25
bug: java.lang.Exception


rsachetto:
Plugin version: ```5.2.20247```
IDE version: ```CL-233.13871```
Exception: ```java.lang.Exception```
Stacktrace:
```text
java.lang.Exception
	at com.jetbrains.rd.platform.util.IdeaRdLoggerFactor$getLogger$1.log(IdeaRdLoggerFactor.kt:28)
	at com.jetbrains.rd.util.SwitchLogger.log(Logger.kt:38)
	at com.jetbrains.rd.framework.impl.RdCall.sync(RdTask.kt:463)
	at com.jetbrains.rd.framework.IRdCall$DefaultImpls.sync$default(TaskInterfaces.kt:123)
	at com.jetbrains.cidr.radler.format.RadExternalFormatter.doFormat(RadExternalFormatter.kt:49)
	at com.jetbrains.cidr.radler.format.RadExternalFormatter.format(RadExternalFormatter.kt:30)
	at com.intellij.psi.codeStyle.ExternalFormatProcessor.formatRangeInFile(ExternalFormatProcessor.java:119)
	at com.intellij.psi.codeStyle.ExternalFormatProcessor.formatRangeInFile(ExternalFormatProcessor.java:127)
	at com.intellij.formatting.service.ExternalFormatProcessorAdapter.lambda$formatRanges$0(ExternalFormatProcessorAdapter.java:56)
	at com.intellij.psi.impl.source.codeStyle.CoreCodeStyleUtil.postProcessRanges(CoreCodeStyleUtil.java:114)
	at com.intellij.formatting.service.ExternalFormatProcessorAdapter.formatRanges(ExternalFormatProcessorAdapter.java:55)
	at com.intellij.formatting.service.FormattingServiceUtil.formatRanges(FormattingServiceUtil.java:123)
	at com.intellij.psi.impl.source.codeStyle.CodeStyleManagerImpl.reformatText(CodeStyleManagerImpl.java:148)
	at com.intellij.psi.impl.source.codeStyle.CodeStyleManagerImpl.reformatText(CodeStyleManagerImpl.java:112)
	at com.intellij.psi.impl.source.codeStyle.CodeStyleManagerImpl.reformatText(CodeStyleManagerImpl.java:105)
	at com.sourcegraph.utils.CodyFormatter$Companion.formatStringBasedOnDocument(CodyFormatter.kt:42)
[...]

PR #362, opened 2024-01-27
Fix for issue 350 - exception on completion

steveyegge:
Wrapped in a write action.

    Also includes some very minor code cleanups.

https://github.com/sourcegraph/jetbrains/issues/350

## Test plan

Tested manually.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -3,6 +3,7 @@ package com.sourcegraph.cody.autocomplete
 import com.intellij.codeInsight.hint.HintManager
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.command.CommandProcessor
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -256,13 +255,13 @@ class CodyAutocompleteManager {
               null
             }
             .completeOnTimeout(null, 3, TimeUnit.SECONDS)
-            .get()
-
-        resetApplication(project)
-        resultOuter.complete(null)
+            .thenRun {
```
mkondratek, line 258: lgtm, but pls double check that autocomplete works fine

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
steveyegge: I checked!

PR #367, opened 2024-01-29
Implement user-selected context files

beyang:
Adds the ability for the user to at-mention context files in the chat input.

https://github.com/sourcegraph/jetbrains/assets/1646931/dd5aa1ff-1f6b-4ea0-98e7-0f4f21fd737a

Not implemented:
* Support for at-mentioned symbols

## Questions

* Are there any e2e or integration tests I should update?

## Test plan

- [ ] Open up a project and open Cody chat
- [ ] At-mention files in the project
- [ ] Try the following methods of selecting at-mentioned files:
  - [ ] Tab to select
  - [ ] Enter to select
  - [ ] Clicking on item to select
- [ ] Try at-mentioning more than one file
- [ ] File should not be included if it is deleted from the text of the message

### DIFF
--- a/src/main/java/com/sourcegraph/cody/PromptPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/PromptPanel.kt
@@ -1,6 +1,5 @@
 package com.sourcegraph.cody
 
-import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.actionSystem.CustomShortcutSet
 import com.intellij.openapi.actionSystem.KeyboardShortcut
@@ -10,6 +9,8 @@ import com.intellij.ui.DocumentAdapter
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.ui.JBUI
 import com.intellij.util.ui.UIUtil
+import com.sourcegraph.cody.agent.WebviewMessage
+import com.sourcegraph.cody.agent.protocol.ContextFile
 import com.sourcegraph.cody.chat.ChatSession
 import com.sourcegraph.cody.chat.CodyChatMessageHistory
 import com.sourcegraph.cody.chat.ui.SendButton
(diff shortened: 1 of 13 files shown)


src/main/java/com/sourcegraph/cody/PromptPanel.kt
```diff
@@ -73,87 +84,222 @@ class PromptPanel(private val chatSession: ChatSession) : JLayeredPane() {
...
+  // if the path contains more than three components, display the last three
+  val path = contextFile.uri.path
 
-    val UP = KeyboardShortcut(KeyStroke.getKeyStroke(KeyEvent.VK_UP, 0), null)
-    val DOWN = KeyboardShortcut(KeyStroke.getKeyStroke(KeyEvent.VK_DOWN, 0), null)
-    val DEFAULT_SUBMIT_ACTION_SHORTCUT = CustomShortcutSet(JUST_ENTER)
-    val POP_UPPER_MESSAGE_ACTION_SHORTCUT = CustomShortcutSet(UP)
-    val POP_LOWER_MESSAGE_ACTION_SHORTCUT = CustomShortcutSet(DOWN)
+  // split path on separator (OS agnostic)
+  val pathComponents = path.split(File.separator)
```
pkukielka: It's safer to avoid string operations on paths.
More idiomatic way to do it could be:
```kotlin
val pathLength = path.getNameCount()
val shortenPath = Paths.get("..").
    resolve(path.subpath(pathLength - 3, pathLength))
```

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -63,8 +65,19 @@ public void webviewPostMessage(WebviewPostMessageParams params) {
         && extensionMessage.getType().equals(ExtensionMessage.Type.TRANSCRIPT)) {
       ApplicationManager.getApplication().invokeLater(() -> onNewMessage.accept(params));
     } else {
-      logger.debug("onNewMessage is null or message type is not transcript");
-      logger.debug(String.format("webview/postMessage %s: %s", params.getId(), extensionMessage));
+      var listener = this.webviewMessageListeners.get(params.getId());
+      if (listener != null) {
+        listener.accept(params.getMessage());
+      } else {
+        logger.warn(
```
pkukielka: Do you think it makes sense to change it back to warn?
I recently switched if from warn to debug because I think this is not a real warning (we consciously do not handle other messages). I think we should setup it in the way that we on dev machines [always see all debug logs](https://plugins.jetbrains.com/docs/intellij/ide-infrastructure.html#logging) but it does not leak to the users.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -63,8 +65,19 @@ public void webviewPostMessage(WebviewPostMessageParams params) {
...
+            String.format("webview/postMessage %s: %s", params.getId(), params.getMessage()));
+      }
     }
   }
+
+  private final Map<String, Consumer<ExtensionMessage>> webviewMessageListeners = new HashMap<>();
+
+  public void onWebviewMessage(String panelID, Consumer<ExtensionMessage> callback) {
```
pkukielka: Looks like `webviewMessageListeners` will be accessed from multiple threads. 
We need to make it `ConcurrentHashMap`.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -46,6 +44,25 @@ private constructor(
 
   init {
     cancellationToken.get().dispose()
+
+    CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
```
pkukielka: You will need to add this code also in `restoreAgentSession`.
The way it works right now is that we do not re relay on stable agent connection or even agent process existence in any given moment in time (as assumptions in that area were was major source of instability in past). `applyAgentOnBackgroundThread` was added specifically to enforce that contract as it waits for agent to be operational before executing any operation on it (and that part of your code is fine).
But after successful startup, agent can be shut down and restarted later (e.g due to crash or account change), and in such case if we rely on any state in/out of agent we need to make sure to restore it. In `restoreAgentSession` we are already doing this for current session restoring chat history in agent and obtain new sessionId.

pkukielka:
It works smooth, nice! 

The only one minot thing I spotted UI-wise is that when there is only one element left on the file list it seems a bit clipped.
If it's possible to make file box just 5px longer it would fix it completely.

image

src/main/java/com/sourcegraph/cody/PromptPanel.kt
beyang: I don't quite get the recommendation here—converting `uri.path` to a `Path` object seems inadvisable, since the former uses posix style path separators, whereas `Path` seems like it would be OS-specific. This does flag an issue where I was using `File.separator` to split the path string, where  I should just be using `/` (and then replacing with `File.separator` in the return value).

pkukielka: Ah, I haven't noticed that we start with uri and not path, and so we have are quarantined to have well defined schema we want to preserve.
You can ignore my comment then.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -43,15 +41,38 @@ private constructor(
...
+  private val newAgentSessionCallbacks =
+      Collections.synchronizedList(mutableListOf<Consumer<CodyAgent>>())
 
   init {
     cancellationToken.get().dispose()
-  }
+    doOnceAndOnNewAgentSession { agent ->
 
-  fun getPanel(): ChatPanel = chatPanel
+      // Register webview message listener on the agent.
```
pkukielka: Agent connection is fully managed in the CodyAgentService already.
But `restoreAgentSession` is necessary to reload history of the every chat inside the agent (as agent is stateless, and if restarted looses any context).

The callback you added is different thing, and indded it would couple those object lifecycles a bit.
On the bright side I already resolved that issue by the way of the rewrite I did for multi-chat. 
Specifically, look at this part:
https://github.com/sourcegraph/jetbrains/blob/main/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt#L24
There is new service called `AgentChatSessionService` which allow you to get chat session by sessionId/panelId. It's used to dispatch messages based on id to the correct `AgentChatSession` instance. Currently it calls `receiveMessage` on it, but you can add second callback (similar to `agent.client.onNewMessage`) which will in the same way dispatch to `receiveWebviewExtensionMessage`.

pkukielka (approved):
LGTM!

PR #433, opened 2024-01-29
Add multi-chat with parallel sessions architecture

exigow:
:construction: WIP :construction:

what works:
* i'm able to restore serialized session and add follow-up questions with correct response (simply put: happy path for this feature)
* tree is automatically updated

TODO:
1. ~~tree needs to be notified for update~~ fixed
    * ~~add smth similar to `historyTree.addLeafToTree(sessionId)` and use when first message appears (used for title)~~
    * ~~delete is also not implemented yet~~
3. implement commands (currently only messages works)
4. context files? (theyre currently broken - decide what next)
5. serialize model


src/main/java/com/sourcegraph/cody/CodyToolWindowFactory.java
```diff
@@ -29,15 +29,7 @@ public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindo
...
-                  .createContent(toolWindowContent.getContentPanel(), "", false);
-          content.setPreferredFocusableComponent(
-              toolWindowContent.getPreferredFocusableComponent());
-          toolWindowContent.addToTabbedPaneChangeListener(
-              () -> {
-                content.setPreferredFocusableComponent(
-                    toolWindowContent.getPreferredFocusableComponent());
-                return null;
-              });
+                  .createContent(toolWindowContent.getAllContentPanel(), "", false);
```
pkukielka, line 32: I think that is change from my PR, no?

exigow:
Moved to https://github.com/sourcegraph/jetbrains/pull/438.

PR #435, opened 2024-01-30
Fix: Context files are not showing up in the chat

### LINKED ISSUE #328: Context files are not showing up in the chat

mkondratek:
## Test plan
- chat and see the context files (try clickable links)


src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagesPanel.kt
```diff
@@ -24,10 +26,14 @@ class MessagesPanel(private val project: Project) :
...
 
     if (componentCount > 0) {
-      val lastPanel = components.last() as? JPanel
-      val lastMessage = lastPanel?.getComponent(0) as? SingleMessagePanel
-      if (message.id == lastMessage?.getMessageId()) {
-        lastMessage.updateContentWith(message)
+      val messageToUpdate =
+          components
+              .mapNotNull { (it as? JPanel) }
+              .filter { it.componentCount > 0 }
```
mkondratek: unrel fix - previously `getComponent(0)` could throw index-out-of-bound

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagesPanel.kt
```diff
@@ -24,10 +26,14 @@ class MessagesPanel(private val project: Project) :
...
+              .map { it.getComponent(0) as? SingleMessagePanel }
+              .find { it?.getMessageId() == message.id }
```
mkondratek: unrel fix - `find` instead of `last` (so we can update the message wherever it is)

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagesPanel.kt
```diff
@@ -43,6 +49,32 @@ class MessagesPanel(private val project: Project) :
     repaint()
   }
 
+  @RequiresEdt
+  @Synchronized
+  fun addOrUpdateContext(message: ChatMessage) {
+    removeBlinkingCursor()
+    createUsedContextFilesPanel(message)?.let(::addComponentToChat)
+    add(BlinkingCursorComponent.instance)
```
mkondratek: I remove the blinking cursor and re-add it just after the context files panel. We receive context files much faster than the actual response. Trust me, it looks good now.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -59,6 +59,8 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
 
   private val logger = LoggerFactory.getLogger(ChatSession::class.java)
 
+  private val theMostRecentContextFilesMessageId = AtomicReference<UUID>(UUID.randomUUID())
```
pkukielka, line 62: I don't like that state variable. It is used for updating single message context but it pollutes whole class. We should think about:
a) removing it completely somehow
b) if not possible then maybe moving to single message - then you don't need fake start UUID, just a boolean

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -179,6 +201,25 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+  @RequiresEdt
+  private fun addAssistantUsedContextToChat(message: ChatMessage) {
+    synchronized(this) {
+      val findResult = messages.indexOfLast { it.id == message.id }
+      if (findResult >= 0) {
+        messages[findResult] = message
+
+        // The first response does not include the context files. We must check
+        // for the second version of it when it's updated with the proper data.
+        val isFirstUpdate = theMostRecentContextFilesMessageId.getAndSet(message.id) != message.id
```
pkukielka, line 213: You don't need that class-wide `theMostRecentContextFilesMessageId` variable.
You can just check if last message in `messages` have the same UUID. If it's the same, it's an update.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -179,6 +201,25 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
     }
   }
 
+  @RequiresEdt
+  private fun addAssistantUsedContextToChat(message: ChatMessage) {
+    synchronized(this) {
+      val findResult = messages.indexOfLast { it.id == message.id }
```
pkukielka, line 207: I think more idiomatic for kotlin would be using find and check if not null, but it's a nitpick.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -155,8 +157,28 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
               extensionMessage.isMessageInProgress == false) {
             getCancellationToken().dispose()
           } else {
-            if (lastMessage?.text != null && extensionMessage.chatID != null) {
-              addAssistantResponseToChat(lastMessage.text, lastMessage.displayText)
+
+            val penultimateMessage = extensionMessage.messages?.dropLast(1)?.lastOrNull()
```
pkukielka: I do not even know what `penultimate` means :) Can you use some other name?

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -155,8 +157,28 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+            if (extensionMessage.chatID != null) {
```
pkukielka: We should not care about `chatId` anywhere in our logic. Why is that needed?

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -155,8 +157,28 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+              if (penultimateMessage != null) {
+                val messageId =
+                    UUID.nameUUIDFromBytes(
+                        (extensionMessage.messages.count() - 1).toString().toByteArray())
```
pkukielka: Why `-1`? It is different formula than used in `addAssistantResponseToChat`, this will cause us issues.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ChatMessage.kt
```diff
@@ -32,9 +31,6 @@ data class ChatMessage(
     val displayText: String? = null,
     val contextFiles: List<ContextFile>? = null,
     val error: ChatError? = null,
-    // Internal ID used for identifying updates of the message
-    // All partial messages which are part of the same response are required to have the same ID
-    val id: UUID = UUID.randomUUID()
```
mkondratek, line 37: note: this is great 🚀

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Speaker.kt
```diff
@@ -1,12 +1,13 @@
...
 import com.google.gson.annotations.SerializedName
 
 enum class Speaker(val speaker: String) {
   @SerializedName("human") HUMAN("human"),
   @SerializedName("assistant") ASSISTANT("assistant")
 }
 
-val SpeakerSerializer = JsonSerializer<Speaker> { src, _, _ -> JsonPrimitive(src.speaker) }
+val speakerDeserializer = JsonDeserializer { src, _, _ ->
+  Speaker.values().find { it.speaker == src.asString }
```
mkondratek, line 12: still, not the best but rather simple and working - I assume our problems with this field come from the fact that we mix serialization and inheritance (here in the client and in the agent)

mkondratek:
closing in favor of https://github.com/sourcegraph/jetbrains/pull/520

PR #438, opened 2024-01-30
Add multi-chat with parallel sessions architecture

exigow:
### Organize multiple chats

You should be able to organize multiple chats and follow up previous conversations.
Cody should "remember" your questions and chat responses, even after closing IDE.

#### Happy path

1. Start a new chat
2. Send message similar to `my favorite color is blue`
3. Close IDE
4. Run IDE and open previous conversation
5. Ask `what's my favorite color?`
6. Response should be similar to `your favorite color is blue`

#### (optional) Test ideas

Useful tips:

* Transcripts in `runIde` mode are located in `build/idea-sandbox/config/options/cody_history.xml`.
* You can force-save transcript by using the `Ctrl` + `S`.

Test ideas:

1. Delete "active" chat. You should be able to delete the currently opened chat. Messages should be removed from Chat tab.
2. Restore historical chat, focus on chat input field and use UP/DOWN keys to cycle between previous questions.
3. Press "new chat" as fast as you can. Especially during the IDE startup.
4. Switch between chats as fast as you can.
5. Press "new chat" while being inside Subscription panel or something other than Chat tab. Tabs should switch automatically.
6. Use commands/recipes inside empty, new chat. Verify serialization/deserialization.
7. Ask about codebase to force response with listed context files and verify if everything is correctly serialized/deserialized. Links to context files should be clickable.
8. Remove all chats using history UI. Tree presentation is empty and branches like "Today" are removed from panel. File with transcripts should also disappear.
9. Use only the keyboard. For example, navigate transcripts with arrows, delete, enter.
10. Start typing while being focused on Chat History to perform search-by-title.
11. Open multiple chats and ask few simultaneous questions in several sessions at once.


src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -39,7 +42,7 @@ interface ChatSession {
 }
 
 class AgentChatSession
-private constructor(private val project: Project, newSessionId: CompletableFuture<SessionId>) :
+private constructor(private val project: Project, private val internalId: String, newSessionId: CompletableFuture<SessionId>, ) :
```
pkukielka: Can we move `internalId` to be last param and add it default with UUID?

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -236,12 +240,33 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+        println("chat session added internalId=${chatSession.internalId}")
       synchronized(chatSessions) { chatSessions.add(chatSession) }
       return chatSession
     }
 
-    private fun createNewPanel(
+      fun createFromState(project: Project, state: ChatState): AgentChatSession {
+          val sessionId = createNewPanel(project) { it.server.chatNew() }
+          val chatSession = AgentChatSession(project, state.internalId!!, sessionId)
+          chatSessions.add(chatSession)
```
pkukielka: Please synchronise this and add it as a last thing in this method. As quickly as you add it, it is possible to receive messages into this session, and we usually od not want that until all else is initialised.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -236,12 +240,33 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+              val parsed = when (val speaker = message.speaker) {
+                  MessageState.SpeakerState.HUMAN -> Speaker.HUMAN
+                  MessageState.SpeakerState.ASSISTANT -> Speaker.ASSISTANT
+                  else -> error("unrecognized speaker $speaker")
+              }
+              val chatMessage = ChatMessage(parsed, message.text, id = UUID.randomUUID())
+              chatSession.messages.add(chatMessage)
+              chatSession.chatPanel.addOrUpdateMessage(chatMessage)
+          }
+          CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
```
pkukielka, line 263: That will be kind of a race.

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.sourcegraph.cody.history.state.HistoryState
+import com.sourcegraph.cody.history.state.MessageState
+import java.time.LocalDateTime
+
+@State(
+    name = "com.sourcegraph.cody.history.HistoryService", storages = [Storage("cody_history.xml")])
+@Service(Service.Level.PROJECT)
+class HistoryService : SimplePersistentStateComponent<HistoryState>(HistoryState()) {
+
+  private val listeners = mutableListOf<(ChatState) -> Unit>()
```
pkukielka, line 15: Don't you need to `synchronize` operation on this list?

src/main/kotlin/com/sourcegraph/cody/history/state/MessageState.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.intellij.openapi.components.BaseState
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+class MessageState : BaseState() {
+
+//  var messageId: String? by string()
+  var text: String? by string()
+  var speaker by enum<SpeakerState>()
+//  var contextFiles by list<String>()
```
pkukielka: We might need context files as well.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -28,9 +30,15 @@ class CodyToolWindowContent(private val project: Project) {
...
   private var codyOnboardingGuidancePanel: CodyOnboardingGuidancePanel? = null
   private val signInWithSourcegraphPanel = SignInWithSourcegraphPanel(project)
-  private val historyTree = JPanel()
+  private val historyTree = HistoryTree(::selectHistory, onDelete = {
+      println("on delete")
+  })
   private val tabbedPane = JBTabbedPane()
 
+  private fun selectHistory(state: ChatState) {
+    addChatSession(AgentChatSession.createFromState(project, state))
```
pkukielka: How does that work if someone select the same chats back and forth multiple times?
Won't it duplicate existing sessions?
I would thought we need to do something like:

```
val session = AgentChatSession.getSessionByInternalId(state.id) ?: AgentChatSession.createFromState(project, state)
addChatSession(session)
```

(better name for `addChatSession ` would probably be `selectChatSession`, and I know I'm guilty of that bad name ;))

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
exigow: Thanks, now added.

exigow (commented):
EDIT: Fixed.

~~Bug:~~

~~To reproduce: Create a new chat, ask a question that forces a long response, open a new chat, and then return to the first one.~~

~~The last message stops being updated in real-time.~~

~~However, everything is fine in the model, and after reloading the chat, the text will be supplemented with more characters. Looks like i need to identify messages somehow to connect token stream again to messages panel after restore.~~

~~There is a message ID in protocol (`ChatMessage.id`), but AFAIK it's added for internal use and it is going to be removed. @mkondratek, is that correct?~~

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
exigow: Good point! Fixed.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
exigow: Fixed.

exigow (commented):
EDIT: Fixed.

~~Bug: up/down arrows are not working after restore.~~

exigow (commented):
EDIT: Fixed + added test scenario.

~~Bug: Remove all chats and group "Today" is still visible in tree panel.~~

~~Empty groups should be removed when empty.~~

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
exigow: Good catch! That fixed one of my problems.

exigow (commented):
EDIT: Fixed after Cody bump.

~~Bug: installing plugin manually on `IU-233.11799.241` causes deadlock on start-up. IDE is not responding.~~

~~Log: https://gist.github.com/exigow/d63e142476137f547be4ce4b0aa8d067~~

~~That's nteresting, because everything is fine via `runIde`.~~

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -135,6 +138,23 @@ class CodyToolWindowContent(private val project: Project) {
     allContentLayout.show(allContentPanel, MAIN_PANEL)
   }
 
+  private fun selectHistory(state: ChatState) {
+    val session =
+        AgentChatSession.getSessionByInternalId(state.internalId!!)
```
pkukielka: I don't think we need those `!!`, do we?

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -135,6 +138,23 @@ class CodyToolWindowContent(private val project: Project) {
...
+            ?: AgentChatSession.createFromState(project, state)
+    addChatSession(session)
+  }
+
+  private fun deleteHistory(state: ChatState) {
+    HistoryService.getInstance().remove(state.internalId!!)
+    val session = AgentChatSession.getSessionByInternalId(state.internalId!!)
+    if (session != null) {
+      AgentChatSession.removeSession(session)
+      val isVisible = session.getPanel().chatSession.getInternalId() == state.internalId!!
```
pkukielka: `session.getPanel().chatSession.getInternalId()` is kind of internal info of the UI panel, we are doing it a bit backwards I think. Maybe we should just add `AtomicReference<ChatSession> currentSession` here in `CodyToolWindowContent` and keep it up to date?
In the long term reaching deep into objects like that will make the code spaghetti again :P

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -236,7 +254,29 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
     @RequiresEdt
     fun createNew(project: Project): AgentChatSession {
       val sessionId = createNewPanel(project) { it.server.chatNew() }
-      val chatSession = AgentChatSession(project, sessionId)
+      val chatSession = getSession(sessionId.get()) ?: AgentChatSession(project, sessionId)
```
pkukielka: We are using this construct in 3 or 4 places already, across two classes.
Can we wrap it in a method?
Could be something like:
`getOrCreateSession(sessionId: SessionId, () -> AgentChatSession): AgentChatSession`

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -236,7 +254,29 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+      synchronized(chatSessions) { chatSessions.add(chatSession) }
+      return chatSession
+    }
+
+    fun createFromState(project: Project, state: ChatState): AgentChatSession {
```
pkukielka, line 259: Missing `@RequiresEdt` or `invokeLater` inside.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -236,7 +254,29 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+      val chatSession =
+          getSession(sessionId.get()) ?: AgentChatSession(project, sessionId, state.internalId!!)
+      for (message in state.messages) {
+        val parsed =
+            when (val speaker = message.speaker) {
+              MessageState.SpeakerState.HUMAN -> Speaker.HUMAN
+              MessageState.SpeakerState.ASSISTANT -> Speaker.ASSISTANT
+              else -> error("unrecognized speaker $speaker")
+            }
+        val chatMessage = ChatMessage(parsed, message.text, id = UUID.randomUUID())
```
pkukielka: `id = UUID.randomUUID()` is a default, you do not need to provide it explicitly.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
```diff
@@ -236,7 +254,29 @@ private constructor(private val project: Project, newSessionId: CompletableFutur
...
+        chatSession.messages.add(chatMessage)
```
pkukielka, line 270: You can replace those two lines with one, and simpler one:
`addMessage(chatMessage)`

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -18,7 +18,7 @@ import javax.swing.BorderFactory
 import javax.swing.JButton
 import javax.swing.JPanel
 
-class ChatPanel(project: Project, chatSession: ChatSession) :
+class ChatPanel(project: Project, val chatSession: ChatSession) :
```
pkukielka, line 21: Revert please :)
Panel should not expose that, it's sign of the problem with the architecture.

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -0,0 +1,55 @@
...
+    message.text = chatMessage.text
+    message.speaker =
+        when (chatMessage.speaker) {
+          Speaker.HUMAN -> MessageState.SpeakerState.HUMAN
+          Speaker.ASSISTANT -> MessageState.SpeakerState.ASSISTANT
+        }
+    return message
+  }
+
+  private fun getChatOrCreate(internalId: String): ChatState {
```
pkukielka, line 43: Can we use proper type instead of string?
I mean, I know it's a string now, but we can keep it UUID, and this way we will never accidentally mix it up with `sessionId`.

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  fun listenOnUpdate(listener: (ChatState) -> Unit) {
+    synchronized(listeners) { listeners += listener }
+  }
+
+  fun update(internalId: String, chatMessages: List<ChatMessage>) {
+    val found = getChatOrCreate(internalId)
+    found.messages = chatMessages.map(::convertToMessageState).toMutableList()
+    found.setUpdatedTimeAt(LocalDateTime.now())
+    listeners.forEach { it(found) }
```
pkukielka: You synchronised it in one place, but you need to do it in all of them.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
exigow: Nullability is leaking from the model and it's caused by used solution. We simply have to always be prepared for the situation where a tag is missing during deserialization (see: BaseState). I'm simply failing fast here with `!!`.

Theoretically, we could add non-null getters (`fun getInternalId(): String`) on the model, but it wouldn't change anything since we still have to throw an exception. Alternatively, we could pass an empty string (`state.internalId ?: ""`), but that would make debugging more difficult in case of potential issues.

What do you suggest?

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
exigow: I agree. Fixed!

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
pkukielka: Something being `null` when we do not expect it should be rare case, no? I mean, some serialisation error or whatever?

If that is the case, then I suggest we validate the model when we load if from disk, and if it's incoherent in any way we simply delete that session. There is no point in showing users chats which they can't open anyway.
This way you can be sure there are no nulls in the model later and interfaces can be clear.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
exigow: Fixed.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
exigow: Fixed, thanks.

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
exigow: Good point! Fixed.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
exigow: Related to/duplicate of https://github.com/sourcegraph/jetbrains/pull/438#discussion_r1472447882.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
exigow: Correct, intuitively we're doing the same thing here, but it causes issues with deserialization:

1. When transcript has one question and one response, then last response is not visible in panel.
2. When transcript has more questions/responses, everything is OK.

I didn't delve into what the cause is (probably that condition inside the method).

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
exigow: I removed the null checks from the code and slightly rewrote the state to be less defensive (less `!!` + less potential NPE's, especially when we're parsing dates).

Ultimately, I would like the state to be completely hidden outside of HistoryService however this required more work. I have reviewed other implementations and it seems that the problem is not solved at all (and nullable leaks into the logic) or a persistent state component hides the state from the world, providing custom setters/getters on component layer (and I like this idea!).

We can refactor this later.

Example: https://github.com/JetBrains/intellij-community/blob/master/plugins/git4idea/src/git4idea/config/GitVcsSettings.java

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
pkukielka: That does not make sense to me, and possible hides some other issue.
I think we need to understand what/why is happening.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
pkukielka: TBH I do not see how it's related ;)

exigow (commented):
Bug:

```
java.lang.Throwable: TreeUI should be accessed only from EDT
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:376)
	at com.intellij.ui.tree.ui.DefaultTreeUI.isValid(DefaultTreeUI.java:216)
	at com.intellij.ui.tree.ui.DefaultTreeUI.getRowForPath(DefaultTreeUI.java:766)
	at java.desktop/javax.swing.plaf.basic.BasicTreeUI.updateLeadSelectionRow(BasicTreeUI.java:2847)
	at java.desktop/javax.swing.plaf.basic.BasicTreeUI$Handler.treeStructureChanged(BasicTreeUI.java:4409)
	at java.desktop/javax.swing.tree.DefaultTreeModel.fireTreeStructureChanged(DefaultTreeModel.java:616)
	at java.desktop/javax.swing.tree.DefaultTreeModel.reload(DefaultTreeModel.java:320)
	at com.sourcegraph.cody.history.HistoryTree.updatePresentation(HistoryTree.kt:79)
	at com.sourcegraph.cody.history.HistoryTree.access$updatePresentation(HistoryTree.kt:28)
	at com.sourcegraph.cody.history.HistoryTree$4.invoke(HistoryTree.kt:59)
	at com.sourcegraph.cody.history.HistoryTree$4.invoke(HistoryTree.kt:59)
	at com.sourcegraph.cody.history.HistoryService.update(HistoryService.kt:25)
	at com.sourcegraph.cody.chat.AgentChatSession.addMessage(ChatSession.kt:179)
	at com.sourcegraph.cody.chat.AgentChatSession.access$addMessage(ChatSession.kt:42)
	at com.sourcegraph.cody.chat.AgentChatSession$Companion.createFromCommand(ChatSession.kt:245)
	at com.sourcegraph.cody.CodyToolWindowContent$commandsPanel$1.invoke(CodyToolWindowContent.kt:46)
	at com.sourcegraph.cody.CodyToolWindowContent$commandsPanel$1.invoke(CodyToolWindowContent.kt:45)
	at com.sourcegraph.cody.commands.ui.CommandsTabPanel$executeCommandWithContext$1$1.invoke(CommandsTabPanel.kt:30)
	at com.sourcegraph.cody.commands.ui.CommandsTabPanel$executeCommandWithContext$1$1.invoke(CommandsTabPanel.kt:29)
	at com.sourcegraph.cody.autocomplete.CodyEditorFactoryListener$Util.informAgentAboutEditorChange$lambda$2(CodyEditorFactoryListener.kt:175)
	at com.sourcegraph.cody.agent.CodyAgentService$Companion$applyAgentOnBackgroundThread$1.invoke$lambda$0(CodyAgentService.kt:92)
	at com.intellij.openapi.application.impl.ApplicationImpl$2.run(ApplicationImpl.java:249)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at com.intellij.util.concurrency.ContextCallable.call(ContextCallable.java:32)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at com.intellij.util.concurrency.ContextRunnable.run(ContextRunnable.java:27)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:702)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:699)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1.run(Executors.java:699)
	at java.base/java.lang.Thread.run(Thread.java:840)
```


Todo after discussion: `CodyToolWindowContent.executeOnInstanceIfNotDisposed` should `invokeLater` inside + simplify usages of `executeOnInstanceIfNotDisposed`.

exigow:
Bug: Tree is not properly refreshed when I'm continuing old conversation. Chat should be placed as first entry in "Today" tree node after sending follow-up question.

exigow (commented):
EDIT: Fixed by Piotr.

~~Bug:~~

~~Chat History is switching automatically to Chat panel after removing chat from tree.~~

~~EDIT: This `isVisible` condition is always true. My mistake. I need to identify session which is currently visible in panel.~~

```
  private fun deleteHistory(state: ChatState) {
    HistoryService.getInstance().remove(state.internalId)
    val session = AgentChatSession.getSessionByInternalId(state.internalId)
    if (session != null) {
      AgentChatSession.removeSession(session)
      val isVisible = session.getPanel().chatSession.getInternalId() == state.internalId
      if (isVisible) addChatSession(AgentChatSession.createNew(project))
    }
  }
```

pkukielka (approved):
LGTM

Sa1to (approved):

exigow:
TODO: Select leaf below after delete.

exigow:
TODO: Delete all.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
exigow: Fixed by Piotr.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
exigow: Fixed by Piotr.

src/main/kotlin/com/sourcegraph/cody/chat/ChatSession.kt
pkukielka: Ok, that code is fine.
The reason it is not working with just `addMessage` is that you call `HistoryService.getInstance().update(internalId, messages)` at the end of it.
You cannot iterate through history and at the same time add a new one.

PR #271, opened 2024-01-09
Fix high CPU utilisation when Cody panel is visible

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/261

## Test plan

Described in: https://github.com/sourcegraph/jetbrains/blob/main/TESTING.md#colour-scheme-changes
During the testing keep Activity Monitor open and check if CPU usage remains low (it can olpy jump up for a moment during the theme change).

PR #440, opened 2024-01-30
Initial implementation for Inline Edit and Document Code

steveyegge:
Committing before it is ready because Olaf would like to play with it. Current state: Document Code makes a successful roundtrip to the agent. And we have Edit Instructions dialog. The rest is not wired up yet.

## Test plan

Locally tested, but the code is by no means ready to use yet.


steveyegge:
Note: Please do not approve or merge this PR.

Update March 18th -- the PR is ready for review & merge.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -20,12 +26,52 @@ public class CodyAgentClient {
...
+    onEditTaskDidChange.accept(params);
+  }
+
+  public void setOnTextDocumentEdit(Consumer<TextDocumentEditParams> callback) {
+    onTextDocumentEdit = callback;
+  }
+
+  @JsonRequest("textDocument/edit")
+  public CompletableFuture<Boolean> textDocumentEdit(TextDocumentEditParams params) {
+    var future = new CompletableFuture<Boolean>();
```
pkukielka: That could be val I think :)
As a side note if you never set result to `false` and only ned it to indicate there were no error maybe `CompletableFuture<Unit>` could be sufficient.

src/main/kotlin/com/sourcegraph/cody/edit/EditCodeInlayRenderer.kt
```diff
@@ -0,0 +1,61 @@
...
+
+class EditCodeInlayRenderer(private val component: JPanel, private val requiredHeight: Int) :
+    EditorCustomElementRenderer {
+
+  private val rendererPane = CellRendererPane()
+
+  init {
+    // Call validate() initially to ensure the layout is updated
+    rendererPane.add(component)
+    component.validate()
```
pkukielka: General comment so I won't forget later:
be careful with EDT calls in init, as then every object creation of `EditCodeInlayRenderer` need to be in the EDT as well, and in past we had hard to spot bugs because of this.

src/main/kotlin/com/sourcegraph/cody/edit/EditCodeInlayRenderer.kt
```diff
@@ -0,0 +1,61 @@
...
+
+  override fun paint(
+      inlay: Inlay<*>,
+      g: Graphics,
+      targetRegion: Rectangle,
+      textAttributes: TextAttributes
+  ) {
+    // Various failed attempts to get something to render in the inlay area.
+    rendererPane.bounds = targetRegion
+    component.setBounds(0, 0, targetRegion.width, targetRegion.height)
```
pkukielka: Probably won't be the case in your situation, but please double check CPU consumption when changing bounds or recalculating layout in `paint` or `repaint` function.
I introduced [nasty bug which spiked our CPU usage](https://github.com/sourcegraph/jetbrains/pull/271/files) significantly in past because of this.

steveyegge:
This branch now has the full implementation, months in the making, of Document Code and Edit Code. It requires the [stevey/extension-client-api](https://github.com/sourcegraph/cody/pull/3445) Cody PR in order to function properly.

This is essentially just a "Happy Path Only" implementation. There is quite a bit more work left to do, but if you are careful, you can use the feature pretty reliably.

- Document Code and Edit Code are available from the right-click context menu (only)
- They are disabled by default; there is a new Setting for enabling them
- Cancel, Undo, and Accept all work

There are still missing features (e.g. Diff, Retry), issues with concurrent editing tasks, behavioral annoyances (such as not auto-formatting the code after rewriting it), and so on. I will create GH issues for each of them.

## Test Plan

Will start on e2e integration testing framework next.

CONTRIBUTING.md
```diff
@@ -352,3 +355,30 @@ To set this up, in the Before launch section of any Run/Debug configuration, do
 You only need to do most of these steps once, when you create the
 External Tool. Then you can use it in any run config that spawns the
 agent, to rebuild it first.
+
+## Debugging VS Code
+
+Sometimes, the Agent behaves differently when called from IntelliJ than
+when called from VS Code, and you may need to debug the same code paths
```
dominiccooney: What does in mean for the agent to be called from VSCode? When the extension calls VSCode APIs and hits agent shims?

CONTRIBUTING.md
```diff
@@ -352,3 +355,30 @@ To set this up, in the Before launch section of any Run/Debug configuration, do
...
+through the Agent twice. First, when called from the JetBrains extension
+side, and again when called from the VS Code extension side.
+
+To accomplish the latter, you can use VSC to debug itself. It works very
```
dominiccooney: VSC => VSCode, ditto below.
```suggestion
To accomplish the latter, you can use VSCode to debug itself. It works very
```

CONTRIBUTING.md
```diff
@@ -352,3 +355,30 @@ To set this up, in the Before launch section of any Run/Debug configuration, do
...
+configurations for debugging VSC Cody in the file `.vscode/launch.json`
+in `sourcegraph/cody`, such as `Launch VS Code Extension (Desktop)`.
+
+You do not launch VSC run configurations from the command palette.
+Instead, use ctrl-shift-D to open the Run and Debug view, and you can
+see the configuration dropdown at the top.
+
+## Known Issues
+
+- Force-stopping the target often corrupts the indexes, requiring an
```
dominiccooney: Naive question but what are "the" indexes?

Nit but can we tighten this up... a list with one bullet, with a sub bullet, and workarounds mentioned in the sub bullet and the next paragraph?

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -26,14 +33,112 @@ public class CodyAgentClient {
...
+  @JsonRequest("textDocument/edit")
+  public CompletableFuture<Boolean> textDocumentEdit(TextDocumentEditParams params) {
+    return onEventThread(
+        () -> {
+          if (onTextDocumentEdit != null) {
+            onTextDocumentEdit.accept(params);
+          } else {
+            logger.warn("No callback registered for textDocument/edit");
+          }
+          return true;
```
dominiccooney, line 95: Unrelated to this change, but I think a lot of our future hopes and dreams hinge on this boolean.

The VSCode edit API can fail, and this boolean indicates the success or failure of the edit. When we have concurrent edits (in the sense that the Agent's model of the document isn't caught up to the JetBrains reality because the user has edited but the extension sends an edit), we could detect that and fail here.

Punchline is that extension code doesn't handle those failures! We would need to have the extension handle that. But there's probably some *not terribly burdensome* abstraction, a "retryable edit", that could do that. Interested in your thoughts here.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -26,14 +33,112 @@ public class CodyAgentClient {
...
+  @Nullable private Consumer<TextDocumentEditParams> onTextDocumentEdit;
+
+  // Callback for the "workspace/executeCommand" request from the agent.
+  @Nullable private Consumer<DisplayCodeLensParams> onDisplayCodeLens;
+
+  // Callback for the "workspace/edit" request from the agent.
+  @Nullable private Consumer<WorkspaceEditParams> onWorkspaceEdit;
+
+  public void setOnEditTaskDidUpdate(@Nullable Consumer<EditTask> callback) {
+    onEditTaskDidUpdate = callback;
```
dominiccooney, line 49: This is fine for now, but longer term is this inviting disaster with overwriting a single callback? Should this throw if the callback is already set, or can the wire-up move to some run once initialization step?

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -26,14 +33,112 @@ public class CodyAgentClient {
...
+
+  public void setOnEditTaskDidDelete(@Nullable Consumer<EditTask> callback) {
+    onEditTaskDidDelete = callback;
+  }
+
+  @JsonNotification("editTask/didUpdate")
+  public void editTaskDidUpdate(EditTask params) {
+    onEventThread(
+        () -> {
+          if (onEditTaskDidUpdate != null) {
```
dominiccooney, line 60: For brevity, what about providing a default callback which logs, making the callback non-nullable everywhere, and just calling directly here?

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
```diff
@@ -47,24 +152,24 @@ private <T> CompletableFuture<T> onEventThread(Supplier<T> handler) {
     return result;
   }
 
+  // Webviews
```
dominiccooney, line 135: Curious where this comes from, seems unrelated to edits.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ClientCapabilities.kt
```diff
@@ -0,0 +1,13 @@
...
+data class ClientCapabilities(
+    var completions: String? = null,
+    var chat: String? = null,
+    var git: String? = null,
+    var progressBars: String? = null,
+    var edit: String? = null,
+    var editWorkspace: String? = null,
+    var codeLenses: String? = null,
+    // 'lenses' for code lenses, must have codeLenses = 'enabled'
+    var fixupControls: String? = null,
```
dominiccooney: This is scree, I removed this and made the Agent always do client-side controls. (...which we can change if we want. But this is unused and should be removed without changes to add it back on the Agent side.)

src/main/kotlin/com/sourcegraph/cody/agent/protocol/CodyTaskState.kt
```diff
@@ -0,0 +1,22 @@
+package com.sourcegraph.cody.agent.protocol
+
+enum class CodyTaskState(val id: Int) {
+  Idle(0),
```
dominiccooney: Don't tear your hair out, I didn't notice this either, but these don't match the extension side after 74d2cd8763 (which is really old, actually) and there may have been changes after that. Current definition:

```
    /**
     * The task has been created, but not yet started.
     */
    idle = 1,
    /**
     * The task has been started, but we have not yet received an actionable
     * response from the LLM.
     */
    working = 2,
    /**
     * We have received an response from the LLM, and we intent to apply the
     * response to the document as we receive it.
     * Similar to `applying` but we do not wait for the LLM to finish responding.
     */
    inserting = 3,
    /**
     * We have received a complete response from the LLM, and we are in the process
     * of appplying the full response to the document.
     */
    applying = 4,
    /**
     * The response has been applied to the document, and we are attempting to format
     * it using the users' preferred formatter.
     */
    formatting = 5,
    /**
     * The response has been applied to the document and we are satisifed enough to present it to the user.
     * The user hasn't technically accepted it, and they can still act on the response.
     * E.g. Undo the change, Retry the change, View the diff.
     */
    applied = 6,
    /**
     * Terminal state. The response has been "accepted" by the user. This is either by:
     * - Clicking "Accept" via the CodeLens
     * - Saving the document
     * - Making an edit within the range of the response (implied acceptance)
     */
    finished = 7,
    /**
     * Terminal state. We received an error somewhere in the process.
     * We present this error to the user, the response can be "discarded" by the user by:
     * - Clicking "Discard" via the CodeLens
     * - Saving the document
     * - Making an edit within the range of the response (implied acceptance)
     */
    error = 8,
    /**
     * Additional state currently only used for the `test` command.
     * This state is used to signify that an Edit is no longer idle, but waiting for
     * some additional information before it is started (e.g. a file name from the LLM)
     */
    pending = 9,
```

src/main/kotlin/com/sourcegraph/cody/agent/protocol/CodyTaskState.kt
```diff
@@ -0,0 +1,22 @@
...
+  Inserting(2),
+  Applying(3),
+  Formatting(4),
+  Applied(5),
+  Finished(6),
+  Error(7),
+  Pending(8)
+}
+
+@Suppress("unused")
```
dominiccooney: Let's delete it

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolCommand.kt
```diff
@@ -0,0 +1,9 @@
+package com.sourcegraph.cody.agent.protocol
+
+data class ProtocolCommand(
+    val title: TitleParams? = null,
+    val command: String? = null,
+    val tooltip: String? = null,
+    // First argument is always task ID (string).
```
dominiccooney: Now that we're not using commands for accept, cancel, undo—what do we need this for?

If we do need it, this is not true for many commands. Let's elaborate the comment a bit, maybe this is a TODO.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -3,7 +3,7 @@ package com.sourcegraph.cody.agent.protocol
 import com.intellij.openapi.vfs.VirtualFile
 import com.sourcegraph.cody.agent.protocol.util.Rfc3986UriEncoder
 
-class TextDocument
+class ProtocolTextDocument
```
dominiccooney, line 6: Why is this named this way? It looks awkward given it is in a package "protocol" and it has a bunch of sibling types that are TextDocumentFoo.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
```diff
@@ -12,4 +14,15 @@ data class Range(val start: Position, val end: Position) {
...
+
+  fun toOffsets(document: Document): Pair<Int, Int> {
+    return Pair(start.toOffset(document), end.toOffset(document))
+  }
+
+  // Return true if the agent gave us all zeroes for the range.
+  fun isZero() = start.isZero() && end.isZero()
+
+  companion object {
+    fun nullRange() = Range(Position(0, 0), Position(0, 0))
```
dominiccooney: Mildly nervous about this "null range", "zero position", etc. idea. Because if you start a new file and tell Cody to do stuff, the position is going to be like this.

Could we make the protocol actually use undefined or null for missing, and trust when zeros do appear that we meant it?

src/main/kotlin/com/sourcegraph/cody/edit/DocumentCodeSession.kt
```diff
@@ -0,0 +1,33 @@
...
+  override fun retry() {
+    // TODO: The actual prompt is displayed as ghost text in the text input field.
+    // E.g. "Write a brief documentation comment for the selected code <etc.>"
+    // We need to send the prompt along with the lenses, so that the client can display it.
+    EditCommandPrompt(controller, editor, "Edit instructions and Retry").displayPromptUI()
+  }
+
+  // Brings up a diff view showing the changes the AI made.
+  override fun diff() {
+    // The FixupController issues a vscode.diff command to show the smart diff in the
```
dominiccooney: No longer true. The VSCode code lens UI does this, but we're using the "client UI" now and I made the vscode.diff an implementation detail of the VSCode code lens UI. (Of course, it is software, we can do whatever we want. But that's where the Agent/extension/VSCode side changes are at right now.)

src/main/kotlin/com/sourcegraph/cody/edit/EditCodeInlayRenderer.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.intellij.openapi.editor.EditorCustomElementRenderer
+import com.intellij.openapi.editor.Inlay
+import com.intellij.openapi.editor.markup.TextAttributes
+import com.intellij.ui.JBColor
+import java.awt.Graphics
+import java.awt.Rectangle
+import javax.swing.CellRendererPane
+import javax.swing.JPanel
+
+class EditCodeInlayRenderer(private val component: JPanel, private val requiredHeight: Int) :
```
dominiccooney: Sorry to be dense, but where is this used?

src/main/kotlin/com/sourcegraph/cody/edit/EditCodeInlayRenderer.kt
```diff
@@ -0,0 +1,61 @@
...
+      g: Graphics,
+      targetRegion: Rectangle,
+      textAttributes: TextAttributes
+  ) {
+    // Various failed attempts to get something to render in the inlay area.
+    rendererPane.bounds = targetRegion
+    component.setBounds(0, 0, targetRegion.width, targetRegion.height)
+    component.validate()
+    rendererPane.validate()
+    rendererPane.background = JBColor.WHITE
```
dominiccooney, line 43: Theming?

src/main/kotlin/com/sourcegraph/cody/edit/EditCodeInlayRenderer.kt
dominiccooney: I'm unsure where this is used, but I assume `component` isn't part of the layout, it is just being used to paint into the renderer pane. So laying it out here should have less potential for mayhem.

src/main/kotlin/com/sourcegraph/cody/edit/EditCodeInlayRenderer.kt
dominiccooney: Theming?

dominiccooney (commented):
I've only looked through half of this but some feedback inline.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
pkukielka: That is old chat API which was removed since, it should be removed from there.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolCommand.kt
pkukielka: If it would be the case it's still needed, and it is required for some commands but not for the others, why do not make it separate param only for commands which need it?

src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextEdit.kt
```diff
@@ -0,0 +1,15 @@
+package com.sourcegraph.cody.agent.protocol
+
+data class TextEdit(
+    // This tag will be 'replace', 'insert', or 'delete'.
+    val type: String,
+
+    // Valid for replace & delete.
+    val range: Range? = null,
```
pkukielka, line 8: Wouldn't it make sense to create two different types of edits, one position based and one renge based? This way we don't need this comments and all fields would always be required and used.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/WorkspaceEditOperation.kt
```diff
@@ -0,0 +1,18 @@
...
+
+data class WorkspaceEditParamsOptions(
+    val overwrite: Boolean = false,
+    val ignoreIfNotExists: Boolean = false,
+    val recursive: Boolean = false
+)
+
+data class WorkspaceEditOperation(
+    val type: String, // all
+    val uri: String? = null, // created, delete, edit
```
pkukielka, line 11: Can we make sure URI parameters are sent as `URI`s and not strings?
Generally I think we should be extra careful around uri/path objects, they were source of many conversion problems in paths between VSCode <-> Intellij, especially adding additional platforms as windows.

src/main/kotlin/com/sourcegraph/cody/edit/DocumentCodeActionHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+class DocumentCodeActionHandler : EditorActionHandler() {
+  private val logger = Logger.getInstance(DocumentCodeActionHandler::class.java)
+
+  override fun isEnabledForCaret(editor: Editor, caret: Caret, dataContext: DataContext?): Boolean {
+    return CodyEditorUtil.isEditorValidForAutocomplete(editor) &&
+        ConfigUtil.isExperimentalInlineEditEnabled()
+  }
+
+  override fun doExecute(editor: Editor, where: Caret?, dataContext: DataContext?) {
+    val fixupService = editor.project?.getService(FixupService::class.java)
```
pkukielka, line 25: Can we move that to `FixupService` as `getInstance` method?
I think that is what we do for all other services.

src/main/kotlin/com/sourcegraph/cody/edit/EditCommandPrompt.kt
```diff
@@ -0,0 +1,248 @@
...
+import java.awt.event.KeyEvent
+import javax.swing.JComponent
+import javax.swing.JLabel
+import javax.swing.JPanel
+import javax.swing.SwingUtilities
+import javax.swing.event.DocumentEvent
+import javax.swing.event.DocumentListener
+
+/** Pop up a user interface for giving Cody instructions to fix up code at the cursor. */
+class EditCommandPrompt(val controller: FixupService, val editor: Editor, val dialogTitle: String) {
```
pkukielka, line 27: Could you please annotate methods which requires dispatch thread with `@RequiresEdt`?
It can give us additional safety net in the newest IJ versions.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+
+  // Sessions for which we have not yet received a task ID, but may receive an edit anyway.
+  private var pendingSessions: MutableSet<FixupSession> = mutableSetOf()
+
+  // The last text the user typed in without saving it, for continuity.
+  private var lastPrompt: String = ""
+
+  init {
+    // JetBrains docs say avoid heavy lifting in the constructor, so pass to another thread.
+    backgroundThread {
```
pkukielka: `CodyAgentService.withAgent(project)` spawns actions on background thread itself, so you don't need to additionally wrap it in `backgroundThread`.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+      CodyAgentService.withAgent(project) { agent ->
+        ApplicationManager.getApplication().invokeLater {
```
pkukielka: That `invokeLater` seems to be unnecessary:
1) It won't have effect on the thread in which callbacks will be called
2) You ensure proper thread for callbacks, e.g. in `editTaskDidUpdate` using `onEventThread`, so you are covered

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+    if (!CodyEditorUtil.isEditorValidForAutocomplete(editor)) {
+      logger.warn("Inline edit invoked when editing not available")
+      return false
+    }
+    return true
+  }
+
+  // TODO: get model list from protocol
+  fun getModels(): List<String> = listOf("GPT-4", "GPT-3.5")
```
pkukielka, line 121: FYI we are already doing it for LLM selector so maybe you can reuse that code.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+import com.sourcegraph.utils.CodyEditorUtil
+
+/** Controller for commands that allow the LLM to edit the code directly. */
+@Service(Service.Level.PROJECT)
+class FixupService(val project: Project) : Disposable {
+  private val logger = Logger.getInstance(FixupService::class.java)
+
+  // We only use this for multiplexing task updates from the Agent to concurrent sessions.
+  // TODO: Consider doing the multiplexing in CodyAgentClient instead.
+  private var activeSessions: MutableMap<String, FixupSession> = mutableMapOf()
```
pkukielka, line 20: Is it possible to have more than one fixup running at the same time, e.g. in different chat window?
If so, this most likely need to be changed to be multithread-safe.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+            } else {
+              session.taskDeleted()
+            }
+          }
+
+          agent.client.setOnWorkspaceEdit { params ->
+            for (op in params.operations) {
+              when (op.type) {
+                "create-file" -> {
+                  logger.warn("Workspace edit operation created a file: ${op.uri}")
```
pkukielka: Should those be `warnings` or just `info`?

src/main/kotlin/com/sourcegraph/cody/edit/DocumentCodeActionHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+    if (fixupService == null) {
+      logger.warn("FixupService not found")
```
pkukielka: It's not really `FixupService` which was missing, but rather `Project`, right?
So maybe something like:

```kotlin
val project = editor.project
if (project == null) {
    logger.warn("No project found, cannot run DocumentCodeActionHandler::startDocumentCode")
} else {
    FixupService.getInstance(project).startDocumentCode(editor)
}
```

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
pkukielka: I would say let's remove `setOnEditTaskDidUpdate` method and move `onEditTaskDidUpdate` update code into `CodyAgentService::onStartup` as other similar callback.
Current code won't work is some scenarios, e.g. after agent restart after a crash or account change.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+                  logger.warn("Workspace edit operation deleted a file: ${op.uri}")
+                }
+                "edit-file" -> {
+                  if (op.edits == null) {
+                    logger.warn("Workspace edit operation has no edits")
+                  } else {
+                    // If there is a pending session, assume that it is the one that caused the
+                    // edit.
+                    val session: FixupSession? =
+                        if (pendingSessions.isNotEmpty()) {
```
pkukielka: I'm not sure if I understand this. Why do we want to performa anything on a `Fixup` which was not fully initialised yet (is pending and have no `taskId`)?

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  fun removeSession(session: FixupSession) {
+    pendingSessions.remove(session)
+    activeSessions.remove(session.taskId)
+  }
+
+  override fun dispose() {
+    activeSessions.values.forEach { it.dispose() }
```
pkukielka, line 147: We don't need to dispose pending ones?

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -0,0 +1,166 @@
...
+  fun setCurrentModel(model: String) {
+    lastSelectedModel = model
+  }
+
+  fun getLastPrompt(): String = lastPrompt
+
+  fun addSession(session: FixupSession) {
+    val taskId = session.taskId
+    if (taskId == null) {
+      pendingSessions.add(session)
```
pkukielka, line 140: So each `FixupSession` sis upposed to be send exactly two times?
Once when it's created (pending) and once when it's updated to active?

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,354 @@
...
+import java.util.concurrent.CompletionException
+import java.util.concurrent.TimeUnit
+
+/**
+ * Common functionality for commands that let the agent edit the code inline, such as adding a doc
+ * string, or fixing up a region according to user instructions.
+ */
+abstract class FixupSession(val controller: FixupService, val editor: Editor) : Disposable {
+  private val logger = Logger.getInstance(FixupSession::class.java)
+  val project = editor.project!!
```
pkukielka: Could we please do not assume that. We had a bugs in past related to `Project` and use of "!!"

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  fun commandCallbacks(): Map<String, () -> Unit> = lensActionCallbacks
+
+  @RequiresEdt
+  private fun triggerDocumentCodeAsync() {
+    // This caret lookup requires us to be on the EDT.
+    val caret = editor.caretModel.primaryCaret.offset
+
+    FixupService.backgroundThread {
```
pkukielka: No need for `backgroundThread`, `CodyAgentService.withAgent(project)` will spawn it itself.

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,354 @@
...
+      }
+    }
+  }
+
+  // We're consistently triggering the 'retrieved codebase context before initialization' error
+  // in ContextProvider.ts. It's a different initialization path from completions & chat.
+  // Calling onFileOpened forces the right initialization path.
+  private fun workAroundUninitializedCodebase(editor: Editor) {
+    val file = FileDocumentManager.getInstance().getFile(editor.document)!!
+    val project = editor.project!!
```
pkukielka: Please let's do not rely on `!!`, it's not always the case that project is present (e.g I think it might throw an exception when closing/disposal of the project started but some code is still running, we have sen this in past).

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,354 @@
...
+    var range = selectionRange
+    if (range == null) {
+      // Be defensive, as the protocol has been fragile with respect to selection ranges.
+      logger.warn("No selection range for session: $this")
+      // Last-ditch effort to show it somewhere other than top of file.
+      val position = Position(editor.caretModel.currentCaret.logicalPosition.line, 0)
+      range = Range(start = position, end = position)
+    } else {
+      // The actual insertion point is on the line above.
+      val position = Position(range.start.line - 1, 0)
```
pkukielka: What if the insertion line is 0, is `range.start.line` equal to -1 then?

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensSpinner.kt
```diff
@@ -0,0 +1,61 @@
+package com.sourcegraph.cody.edit.widget
+
+import com.intellij.openapi.application.ApplicationManager
+import java.awt.FontMetrics
+import java.awt.Graphics2D
+import java.awt.geom.AffineTransform
+import javax.swing.Icon
+import javax.swing.Timer
+
+class LensSpinner(group: LensWidgetGroup, private val icon: Icon) : LensWidget(group) {
```
pkukielka, line 10: I'm not sure if I get purpose of this class correctly, but IntelliJ supports [animated icons](https://plugins.jetbrains.com/docs/intellij/icons.html#animated-icons) for spinners (see: `AnimatedIcon.Default`).
If your use case requires something completely custom could elaborate on why?

src/main/kotlin/com/sourcegraph/cody/edit/EditCommandPrompt.kt
```diff
@@ -0,0 +1,248 @@
...
+    val file = FileDocumentManager.getInstance().getFile(editor.document)?.name ?: "unknown file"
+    topRow.add(JLabel("Editing $file at $line:$col"), BorderLayout.CENTER)
+
+    val southRow = JPanel(BorderLayout())
+
+    val historyLabel =
+        JLabel().apply {
+          text =
+              if (promptHistory.isNotEmpty()) {
+                "↑↓ for history"
```
pkukielka, line 213: Please make sue that ascii is also properly rendered on windows

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,354 @@
...
+import java.util.concurrent.CancellationException
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CompletionException
+import java.util.concurrent.TimeUnit
+
+/**
+ * Common functionality for commands that let the agent edit the code inline, such as adding a doc
+ * string, or fixing up a region according to user instructions.
+ */
+abstract class FixupSession(val controller: FixupService, val editor: Editor) : Disposable {
```
pkukielka: It looks to me that two things you really need instead of editor are document and project.
Maybe it would make sense to pass them as non null constructor parameters instead of editor?

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,354 @@
...
+            "delete" -> ReplaceUndoableAction(this, edit, marker)
+            "insert" -> InsertUndoableAction(this, edit, marker)
+            else -> logger.warn("Unknown edit type: ${edit.type}")
+          }
+        }
+      }
+    }
+  }
+
+  private fun createMarkerForEdit(doc: Document, edit: TextEdit): RangeMarker? {
```
pkukielka: It seems to me that logic from there could be just moved inside `ReplaceUndoableAction` and `InsertUndoableAction`, and you wouldn't have to pass marker through the constructor.

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,354 @@
...
+    val project = editor.project ?: return
+    if (project.isDisposed) return
+    val fileEditor = getEditorForDocument(editor.document, project)
+    val undoManager = UndoManager.getInstance(project)
+    if (undoManager.isUndoAvailable(fileEditor)) {
+      undoManager.undo(fileEditor)
+    }
+  }
+
+  private fun getEditorForDocument(document: Document, project: Project): FileEditor? {
```
pkukielka: A lot of method signatures like this one would simplify a lot if you will make `document: Document`, and  `project: Project` parameters of `FixupSession` constructor.

CONTRIBUTING.md
steveyegge: This is fair feedback; I'll rewrite it to use the right terminology.

CONTRIBUTING.md
steveyegge: Fixed.

CONTRIBUTING.md
steveyegge: Fixed.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
steveyegge: Yes, let's do it -- I've seen it fail to apply once (no code returned by LLM) and would love to get more feedback to the user when this happens.

src/main/java/com/sourcegraph/cody/agent/CodyAgentClient.java
steveyegge: I agree, but there are 8 of these callbacks (I think my PR only added one), and we should rearchitect all of them in one go. I've added [issue 1136](https://github.com/sourcegraph/jetbrains/issues/1136) to track it.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ClientCapabilities.kt
steveyegge: Removed.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/CodyTaskState.kt
steveyegge: Fixed!

src/main/kotlin/com/sourcegraph/cody/agent/protocol/CodyTaskState.kt
steveyegge: done

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolCommand.kt
steveyegge: I've ripped out the entire code path; it's easy enough to put back in if we ever need it.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
steveyegge: Apologies; there were 3 unused methods here and this was one. I've done a sweep for other warnings as well.

src/main/kotlin/com/sourcegraph/cody/edit/DocumentCodeSession.kt
steveyegge: fixed

src/main/kotlin/com/sourcegraph/cody/edit/EditCodeInlayRenderer.kt
steveyegge: Was a leftover, not used - removed.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextEdit.kt
steveyegge: Probably, but it's a bigger change; maybe we can wait until after this PR is merged?

src/main/kotlin/com/sourcegraph/cody/agent/protocol/WorkspaceEditOperation.kt
steveyegge: They have been bouncing back and forth in origin/main between URI and String and causing me no end of merge headaches. My understanding is that URI has trouble serializing in some cases.

It works now, and I don't want to break it, so let's plan an API cleanup pass afterwards.

src/main/kotlin/com/sourcegraph/cody/edit/EditCommandPrompt.kt
steveyegge: Done!

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
steveyegge: Good catch - fixed!

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
steveyegge: Great catch - was the result of some refactoring. Fixed.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
steveyegge: Yikes, forgot that's not wired up yet -- I've added a [GH issue](https://github.com/sourcegraph/jetbrains/issues/1137) to track it. Thanks.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
steveyegge: It is, and there's a tracking bug to fix all the concurrency issues with the existing code. (Or, if we don't have time to get it done for release, we need to block concurrent edit tasks.)

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
steveyegge: They're really TODOs in disguise. Create/Rename/Delete file ops need to be supported before the first release, and there is a [tracking issue](https://github.com/sourcegraph/jetbrains/issues/1119) for it.

src/main/kotlin/com/sourcegraph/cody/edit/DocumentCodeActionHandler.kt
steveyegge: Ha! Good catch; fixed.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
steveyegge: This is a problem with the protocol that needs to be fixed; when we get editing operations, we don't know which task they are associated with (no taskId is passed).

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
steveyegge: Good call; fixed.

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
steveyegge: Is it OK if I throw an exception there, then, if it's null?

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
steveyegge: Fixed.

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
steveyegge: Fixed.

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
steveyegge: Good catch. Left over hack around a backend bug that's since been fixed; I've removed the -1.

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensSpinner.kt
steveyegge: I've taken a look and I would love to replace this code with the AnimatedIcon. But I judge it to be more than an hour's work, so I've filed [issue 1142](https://github.com/sourcegraph/jetbrains/issues/1142) to track it.

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
steveyegge: Although I do use the editor in at least three other situations in this file, you are absolutely correct that passing in non-null project and document parameters to the constructor will significantly simplify the rest of the code in the file. So I went ahead and did that, and pushed a fix to this branch.

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  fun cancel() {
+    CodyAgentService.withAgent(project) { agent ->
+      agent.server.commandExecute(CommandExecuteParams(COMMAND_CANCEL, listOf(taskId!!)))
+    }
+    if (performedEdits) {
+      undo()
+    } else {
+      finish()
```
pkukielka, line 232: That `finish()` can be executed before `agent.server.commandExecute(CommandExecuteParams(COMMAND_CANCEL, listOf(taskId!!)))`, as that agent code waits for agent to be available and executes on the background thread asynchronously. 
Just wanted to confirm that tis is fine?

pkukielka (approved):
Let's merge it and do a remaining work on `main` :)

PR #447, opened 2024-01-31
Fix internal error on quick project close & open

mkondratek:
## Test plan
- quickly open and close a project (then reopen to see any internal error)


src/main/kotlin/com/sourcegraph/cody/initialization/EndOfTrialNotificationScheduler.kt
```diff
@@ -48,12 +51,14 @@ class EndOfTrialNotificationScheduler private constructor(val project: Project)
             val codyProTrialEnded =
                 agent.server
                     .evaluateFeatureFlag(GetFeatureFlag.CodyProTrialEnded)
-                    .get(4, TimeUnit.SECONDS) == true
+                    .orTimeout(4, TimeUnit.SECONDS)
+                    .getNow(false) == true
```
mkondratek: I've seen the error coming from this place. I managed to get it twice but it's a race (hard to reproduce). This approach will not throw anything.

src/main/kotlin/com/sourcegraph/cody/initialization/EndOfTrialNotificationScheduler.kt
```diff
@@ -48,12 +51,14 @@ class EndOfTrialNotificationScheduler private constructor(val project: Project)
...
 
             val useSscForCodySubscription =
                 agent.server
                     .evaluateFeatureFlag(GetFeatureFlag.UseSscForCodySubscription)
-                    .get(4, TimeUnit.SECONDS) == true
+                    .orTimeout(4, TimeUnit.SECONDS)
+                    .getNow(false) == true
```
mkondratek: Same as above. I did the same thing in other places where it seemed relevant.

src/main/kotlin/com/sourcegraph/cody/statusbar/CodyAutocompleteStatusService.kt
```diff
@@ -38,9 +38,11 @@ class CodyAutocompleteStatusService : CodyAutocompleteStatusListener, Disposable
 
   override fun onCodyAutocompleteStatusReset(project: Project) {
     ApplicationManager.getApplication().executeOnPooledThread {
-      val notify = didStatusChanged(project)
-      if (notify) {
-        updateCodyStatusBarIcons()
+      if (!project.isDisposed) {
```
mkondratek, line 41: Another issue - the same scenario (a quick open & close project) - this one is (almost) deterministic. Should not be a case anymore.

src/main/kotlin/com/sourcegraph/common/UpgradeToCodyProNotification.kt
```diff
@@ -77,9 +76,7 @@ private constructor(title: String, content: String, shouldShowUpgradeOption: Boo
               else -> CodyBundle.getString("UpgradeToCodyProNotification.title.explain")
             }
 
-        ApplicationManager.getApplication().invokeLater {
-          UpgradeToCodyProNotification(title, content, shouldShowUpgradeOption).notify(project)
-        }
+        UpgradeToCodyProNotification(title, content, shouldShowUpgradeOption).notify(project)
```
mkondratek, line 79: Unrel: redundant `invokeLater` - `notify` handles the threading already.

src/main/java/com/sourcegraph/vcs/RepoUtil.java
```diff
@@ -159,8 +159,8 @@ private static String doReplacements(
                 .thenCompose(
                     agent ->
                         agent.getServer().convertGitCloneURLToCodebaseName(new CloneURL(cloneURL)))
-                .completeOnTimeout(/* value= */ null, /* timeout= */ 4, TimeUnit.SECONDS)
-                .get();
+                .orTimeout(4, TimeUnit.SECONDS)
```
pkukielka: Why that change?
I do not understand how `.orTimeout(...).getNow(null)` is supposed to work, `orTimeout ` throws exception after a given time, but `getNow` returns immediately. That `orTimeout` will do nothing there. I think `.completeOnTimeout(...).get()` was in fact proper solution.
Please elaborate :)

pkukielka (changes requested):
I added one comment but it applies to a few places. Please do not merge before claryfying.

src/main/java/com/sourcegraph/vcs/RepoUtil.java
mkondratek: Hmmm... I did think about it. Probably getnow. Ignores the timeout 🤔 anyway. My point is I want to ensure that the timeout applies and get will not throw

mkondratek:
The icon is more visible and has the same colors as everywhere else.

image
image

pkukielka (approved):
LGTM

PR #451, opened 2024-02-01
Fix chat history sorting after posting message to the old chat

Sa1to:
It fixes issue with chat history sorting when new message is posted in the old chat: https://github.com/sourcegraph/jetbrains/pull/438#issuecomment-1919594604
## Test plan
* Create new chat and post message
* Create second chat and post message
* Check history tree
* Post message in the first chat
* Old chat should be moved to the top of the tree

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/history/HistoryTree.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/HistoryTree.kt
@@ -72,11 +72,7 @@ class HistoryTree(
         model.reload(root)
       } else {
         val period = root.periods().find { it.periodText == periodText }!!
-        val extended = period.leafs() + LeafNode(chat)
-        val sorted = extended.sortedByDescending { it.chat.getUpdatedTimeAt() }
-        period.removeAllChildren()
-        for (child in sorted) period.add(child)
-        model.reload(period)
+        addChatToPeriodAndSort(period, chat)
       }
     } else {
       val period =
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/history/HistoryTree.kt
```diff
@@ -78,6 +78,17 @@ class HistoryTree(
...
         model.reload(period)
       }
+    } else {
+      val period =
+          root
+              .periods()
+              .flatMap { it.leafs() }
+              .find { it.chat.internalId == chat.internalId }!!
+              .parent as PeriodNode
+      val sorted = period.leafs().sortedByDescending { it.chat.getUpdatedTimeAt() }
```
exigow: `period.leafs().sortedByDescending { it.chat.getUpdatedTimeAt() }` is now good candidate for a function (maybe smth like `Period.sortLeafs()`?).

pkukielka: I'm not sure if that logic is correct. What if you will update chat from different period, e.g. from yesterday?

exigow: Tip: You probably know this, but for the record: You can easily reproduce this by editing `cody_history.xml` and changing `updatedAt` to some old date.

pkukielka: I think what I reported is a small bug but it might not be super easy to fix.
Let's create a ticket and merge this one please.

pkukielka (approved):
LGTM

PR #452, opened 2024-02-01
Fix `TreeUI should be accessed only from EDT` exception

exigow:


src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
```diff
@@ -23,10 +23,10 @@ class AccountSettingChangeListener(project: Project) : ChangeListener(project) {
 
             // Notify Cody Agent about config changes.
             if (ConfigUtil.isCodyEnabled()) {
-              CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
-                agent.server.configurationDidChange(ConfigUtil.getAgentConfiguration(project))
-              }
-              CodyToolWindowContent.executeOnInstanceIfNotDisposed(project) {
+              CodyToolWindowContent.invokeLaterIfNotDisposed(project) {
```
pkukielka: The kind of does not make sense. We use invoke later to immediately switch to background thread.
Maybe we need two methods.

exigow: Updated after internal discussion.

exigow:
Duplicate of https://github.com/sourcegraph/jetbrains/pull/460

PR #455, opened 2024-02-01
Minor fixes for history tab removal process

pkukielka:
## Test plan

Full pre-release QA will be done on this change today.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -70,12 +70,12 @@ class CodyToolWindowContent(private val project: Project) {
     switchToChatSession(AgentChatSession.createNew(project))
   }
 
-  fun switchToChatSession(chatSession: AgentChatSession) {
+  fun switchToChatSession(chatSession: AgentChatSession, showChatWindow: Boolean = true) {
     ApplicationManager.getApplication().invokeLater {
       currentChatSession.getAndSet(chatSession)
       chatContainerPanel.removeAll()
       chatContainerPanel.add(chatSession.getPanel())
-      tabbedPane.selectedIndex = CHAT_TAB_INDEX
+      if (showChatWindow) tabbedPane.selectedIndex = CHAT_TAB_INDEX
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -54,20 +54,18 @@ private constructor(
       sessionId.get().getNow(null) == thatSessionId
 
   fun restoreAgentSession(agent: CodyAgent) {
-    synchronized(this) {
```
pkukielka, line 57: I removed synchronisation because it does not have any sense on the EDT

src/main/kotlin/com/sourcegraph/cody/history/HistoryTree.kt
```diff
@@ -86,11 +87,21 @@ class HistoryTree(
...
-      val period = leaf.parent as PeriodNode
-      model.removeNodeFromParent(leaf)
+    tree.selectedLeafOrNull()?.let { selectedLeaf ->
+      val period = selectedLeaf.parent as PeriodNode
+      val selectedIndex = period.getIndex(selectedLeaf)
+      onDelete(selectedLeaf.chat)
+      model.removeNodeFromParent(selectedLeaf)
       if (model.getChildCount(period) == 0) model.removeNodeFromParent(period)
+
+      val leafsCount = period.childCount
```
pkukielka, line 97: This logic will preserve selection:
* on the same index if possible
* on the previous index if removed element was the last one

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -140,7 +140,9 @@ class CodyToolWindowContent(private val project: Project) {
     HistoryService.getInstance().remove(state.internalId)
     if (AgentChatSessionService.getInstance(project).removeSession(state)) {
       val isVisible = currentChatSession.get()?.getInternalId() == state.internalId
-      if (isVisible) switchToChatSession(AgentChatSession.createNew(project))
+      if (isVisible) {
+        switchToChatSession(AgentChatSession.createNew(project), showChatWindow = false)
```
pkukielka, line 144: When we remove element we don't want to switch to the chat window.

mkondratek (approved):
parówal

PR #458, opened 2024-02-01
Cody action group should be visible even during the indexing

pkukielka:
## Test plan

Manual testing:
1) Trigger clearing caches from File menu
2) Check if Cody context menu is present and enabled during the indexing


Sa1to (approved):
LGTM

mkondratek (approved):

PR #463, opened 2024-02-01
Do not check for subscription data when no account signed in

mkondratek:
## Test plan
- no account signned in -> open a project


exigow (approved):
LGTM :+1:

PR #474, opened 2024-02-02
Bugfix: Disable New Chat action when there are no active accounts

exigow:
Also: Added `CodyAuthenticationManager.hasNoActiveAccount(project)` and used everywhere (?).

## Test plan

1. Remove account -> icon should be hidden & action shortcut should be disabled
2. Add account -> icon appears & action shortcut works again

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/NewChatAction.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/NewChatAction.kt
@@ -6,6 +6,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.wm.ToolWindowManager
 import com.sourcegraph.cody.CodyToolWindowContent
 import com.sourcegraph.cody.CodyToolWindowFactory
+import com.sourcegraph.cody.config.CodyAuthenticationManager
 
 class NewChatAction : DumbAwareAction() {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/NewChatAction.kt
```diff
@@ -17,6 +18,12 @@ class NewChatAction : DumbAwareAction() {
     showToolbar(project)
   }
 
+  override fun update(event: AnActionEvent) {
+    val project = event.project ?: return
+    val hasActiveAccount = !CodyAuthenticationManager.instance.hasNoActiveAccount(project)
```
pkukielka, line 23: How often is update called? And how expensive `CodyAuthenticationManager.instance.hasNoActiveAccount` is?
It's quite important as I think it's called on the EDT now, right?

exigow: Update is called very often: Any interaction with a toolbar fires this method few times per second (clicking, hovering with mouse, etc).

It's fast and "offline" service with persistent state. It's `cody_accounts.xml`.

pkukielka (approved):
LGTM

PR #475, opened 2024-02-02
Add "Delete All Chats" popup action

### LINKED ISSUE #371: GA: Users can create new chats, view old chats (persisted across sessions and grouped by timestamp), and delete chats (individually or all chats).

exigow:
Fixes #371

## Test plan

1. The Remove All action should be always enabled.
2. You should be able to use the Remove All action while tokens are still being generated.
3. The Chat panel should be automatically cleared, and sessions terminated.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/history/HistoryTree.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/HistoryTree.kt
@@ -28,7 +28,8 @@ import javax.swing.tree.TreeSelectionModel
 
 class HistoryTree(
     private val onSelect: (ChatState) -> Unit,
-    private val onDelete: (ChatState) -> Unit
+    private val onRemove: (ChatState) -> Unit,
+    private val onRemoveAll: () -> Unit
 ) : SimpleToolWindowPanel(true, true) {
 
   private val model = DefaultTreeModel(buildTree())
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/history/HistoryTree.kt
```diff
@@ -160,32 +177,31 @@ class HistoryTree(
...
-      private val tree: SimpleTree,
       text: String,
-      icon: Icon? = null,
+      icon: Icon?,
+      private val isEnabled: () -> Boolean,
       private val action: () -> Unit
   ) : AnAction(text, null, icon) {
+
     override fun update(e: AnActionEvent) {
-      super.update(e)
```
pkukielka, line 169: Is removal of `super.update(e)` on purpose? What does it change?

exigow: Implementation of `super.update(e)`:

```java
    public void update(@NotNull AnActionEvent e) {
        if (e == null) {
            $$$reportNull$$$0(10);
        }

    }
```

So is basically non-null assertion. We don't have to do that because `AnActionEvent` is non-null in our Kotlin code (not `AnActionEvent?`). That's why i believe i can remove that.

pkukielka (approved):
LGTM

PR #497, opened 2024-02-06
Add commands shortcuts

### LINKED ISSUE #408: The test command can be triggered from the sidebar, right-click menu, or keyboard shortcut

mkondratek:
Docs updated. 
Subscription renamed to My Account.
Shortcuts added.
Mnemonics for commands added.

Fixes https://github.com/sourcegraph/jetbrains/issues/408, https://github.com/sourcegraph/jetbrains/issues/412, https://github.com/sourcegraph/jetbrains/issues/415.

## Test plan

| Command       | Windows / Linux                                      | MacOs                                                |
|---------------|------------------------------------------------------|------------------------------------------------------|
| Explain Code  | <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>1</kbd>     | <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>E</kbd> |
| Smell Code    | <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>2</kbd> | <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>S</kbd> |
| Generate Test | <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>3</kbd> | <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>T</kbd> |


pkukielka (approved):
LGTM!

TESTING.md
```diff
@@ -119,13 +119,23 @@ Prerequisite: You have to be **signed in**. This is important because we expect
...
 #### Expected behaviour
 
-* List of commands is immediately available after the toolbar is displayed. **No refresh is required.**
+* List of commands is immediately available after the tool window is displayed. **No refresh is required.**
+
+### General commands availability from keyboard shortcuts
+
+1. Explain Code: <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>E</kbd>
+2. Smell Code: <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>S</kbd>
+3. Generate Test: <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>T</kbd>
```
exigow: `control` != `Ctrl`?

mkondratek: control works on macOs

I am trying to find a corresponding (or alternative shortcut on windows). Please let me know if you find something on Linux.

exigow (changes requested):
Unfortunately, these shortcuts don't work for me on Linux. I've tried all the "special keys". I'm using CM Storm Quick Fire XT keyboard.

I've did some search and I've found some claims that the Meta key is not found on modern keyboards and Mac keyboard has a completely different set of modifiers. Can you please verify?

BTW out of curiosity, I've scrolled through all the shortcuts in Settings > Keymap, and none of them uses the 'meta' modifier. I think we should look for a more OS-agnostic (and less confusing :p) key combination.

TESTING.md
exigow: Thanks! But what's the difference between `control` and `meta`? 

(My confusion arises from the fact that in the Settings, it says 'meta', while in the testing docs, it's 'control' - sorry if I'm asking stupid questions.)

mkondratek:
Shortcuts updated for Linux & Windows.

src/main/kotlin/com/sourcegraph/cody/chat/actions/BaseCommandAction.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.intellij.openapi.fileEditor.FileEditorManager
+import com.intellij.openapi.project.Project
+import com.sourcegraph.cody.CodyToolWindowContent
+import com.sourcegraph.cody.autocomplete.CodyEditorFactoryListener
+import com.sourcegraph.cody.chat.AgentChatSession
+import com.sourcegraph.cody.commands.CommandId
+
+abstract class BaseCommandAction : BaseChatAction() {
+
+  abstract fun myCommandId(): CommandId
```
exigow, line 13: We can use `abstract val myCommandId: CommandId` here, no need to define this as function

src/main/kotlin/com/sourcegraph/cody/commands/CommandId.kt
```diff
@@ -1,7 +1,7 @@
 package com.sourcegraph.cody.commands
 
-enum class CommandId(val displayName: String) {
-  Explain("Explain Code"),
-  Smell("Smell Code"),
-  Test("Generate Test")
+enum class CommandId(val displayName: String, val mnemonic: Char) {
+  Explain("Explain Code", 'E'),
+  Smell("Smell Code", 'S'),
+  Test("Generate Test", 'T')
```
exigow, line 6: Great idea with those mnemonic keys!

BTW I see a potential UX issue: 

* Mnemonic combination is Alt + E, S, or T
* Shortcut version is Alt + Shift + 1, 2, or 3

IMO we should pick one and use the same keys. If you're not sure which ones to choose, it seems to me that your suggestion (E, S, and T) is definitely a better version to to memorize (+ it is compatible with mnemonic letters).

TESTING.md
```diff
@@ -119,13 +119,25 @@ Prerequisite: You have to be **signed in**. This is important because we expect
...
-* List of commands is immediately available after the toolbar is displayed. **No refresh is required.**
+* List of commands is immediately available after the tool window is displayed. **No refresh is required.**
+
+### General commands availability from keyboard shortcuts
+
+| Command       | Windows / Linux                                      | MacOs                                                |
+|---------------|------------------------------------------------------|------------------------------------------------------|
+| Explain Code  | <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>1</kbd>     | <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>E</kbd> |
+| Smell Code    | <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>2</kbd> | <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>S</kbd> |
+| Generate Test | <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>3</kbd> | <kbd>control</kbd> + <kbd>Shift</kbd> + <kbd>T</kbd> |
```
exigow, line 136: We need to reformat this.

TESTING.md
```diff
@@ -119,13 +119,25 @@ Prerequisite: You have to be **signed in**. This is important because we expect
...
+
+#### Expected behaviour
+
+* Commands work as executed from any other place (the tool window opens if needed).
```
exigow, line 141: I think we should add some info about mnemonics and how this works/what to expect (it was not obvious for me).

TESTING.md
exigow: `meta` removed, closing thread

exigow (approved):
Looks good and works on Linux now! :+1: 

I've opened few tiny threads. I'm approving in advance because I believe you will address my comments (especially [this one](https://github.com/sourcegraph/jetbrains/pull/497#discussion_r1480602388)).

exigow:
OH NO AUTOMERGE.

@mkondratek

TESTING.md
mkondratek: why? what's wrong here?

TESTING.md
mkondratek: Let me explain. 

Mnemonics are a common solution applied in IntelliJ and its plugins (and probably many other tools. They help to "click" buttons and choose options without the mouse (or clicking TAB key). Having the focus on some window you click `option` on macOS (idk what is the key on Linux) - having the key pressed you see the buttons have underlined characters - ideally each button has a different character. Then you can click the specific (underlined) key to "click" the given button. 

Try it in IntelliJ's commit panel. Pressing `option` there underlines `m` in `Amend`, `i` in `Commit` and `P` in `Commit and Push...`. 

---

I do not think we need to describe it here. It is very simple. Only nerds use that 🤓 It was not requested officially. It's just a nice trick useful for ones that like to use it.

src/main/kotlin/com/sourcegraph/cody/chat/actions/BaseCommandAction.kt
mkondratek: I will fix that in other PR

src/main/kotlin/com/sourcegraph/cody/commands/CommandId.kt
mkondratek: Please see my comment above.

tl;dr
Only letters that are part of the action texts can be used for mnemonics. Mnemonics is not the same thing as shortcuts.

PR #503, opened 2024-02-06
Isolate chats between accounts

### LINKED ISSUE #374: Chats should not persist between accounts. If I create a chat while signed into my Free account through Sourcegraph.com, I should not see that chat when I sign into my account through sourcegraph.sourcegraph.com.

exigow:
Fixes #374

## Test Plan

Prerequisite: You need two working accounts. Preferably one Free, and one Enterprise.

1. Switch to first account.
2. Send a message.
3. Switch to second account.
4. Send a message.

These two chats should be isolated between different accounts. Both accounts should have one conversation each. It doesn't matter if you're on FREE or Enterprise: every chat now is assigned to account.

You should also be able to switch between accounts while tokens are still being generated.

Note about migration: Old chats will preserved in the new schema. This works so that orphaned chats will be assigned to the currently active account.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
@@ -179,7 +179,7 @@ private constructor(
     }
     messages.add(message)
     chatPanel.addOrUpdateMessage(message)
-    HistoryService.getInstance().update(internalId, messages)
+    HistoryService.getInstance().update(project, internalId, messages)
   }
 
   @RequiresEdt
--- a/TESTING.md
+++ b/TESTING.md
@@ -18,6 +18,7 @@
     - [ ] [Autoscroll to latest message](#autoscroll-to-latest-message)
     - [ ] [Read chat history without interruptions](#read-chat-history-without-interruptions)
     - [ ] [Organize multiple chats](#organize-multiple-chats)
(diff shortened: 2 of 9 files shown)


exigow:
~TODO: What if users have serialized chats without an assigned userId? Should they see them on each account, or should be skipped during deser?~ fixed

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -176,7 +176,7 @@ private constructor(
     }
     messages.add(message)
     chatPanel.addOrUpdateMessage(message)
-    HistoryService.getInstance().update(internalId, messages)
+    HistoryService.getInstance().update(project, internalId, messages)
```
pkukielka, line 182: `HistoryService` should have project automatically passed as a parameter to the class, as it is project-level service (current implementation of `HistoryService.getInstance` is incorrect as even IDE suggests).
I mostly fixed that on my PR with multi-repo.

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
exigow: Good point!

PR #505, opened 2024-02-06
Add Open Chat action

### LINKED ISSUE #373: Users can focus on an open chat with a keyboard shortcut.

exigow:
Fixes #373

## Test Plan

* Use Alt + MINUS to open Chat.

More test scenarios in TESTING.md.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -259,6 +259,9 @@ Test ideas:
 10. Start typing while being focused on Chat History to perform search-by-title.
 11. Open multiple chats and ask few simultaneous questions in several sessions at once.
 12. Open new chat with <kbd>Alt</kbd> + <kbd>=</kbd> shortcut (or <kbd>Option</kbd> + <kbd>=</kbd> on Mac).
+13. Open existing chat with shortcut <kbd>Alt</kbd> + <kbd>-</kbd> (or <kbd>Option</kbd> + <kbd>-</kbd> on Mac) and start typing question. Tab should be switched automatically on Chat.
+14. Second click <kbd>Alt</kbd> + <kbd-</kbd> should hide tool window if focused (similar behavior as other tool windows).
+15. Click <kbd>Esc</kbd> while being focused inside Cody tool window. You should be automatically focused on code.  
 
 ## Code Search
 
--- a/src/main/java/com/sourcegraph/cody/PromptPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/PromptPanel.kt
@@ -119,6 +119,8 @@ class PromptPanel(
     }
   }
 
+  fun focus() = textArea.requestFocusInWindow()
+
   private fun didUseShortcut(shortcut: CustomShortcutSet) {
     if (contextFilesListView.model.size > 0) {
       when (shortcut) {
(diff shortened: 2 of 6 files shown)


exigow (commented):
This feature works, but not in the way I would like it to.

First problem: From what I've found, each toolbar has its "Activation${toolbarName}ToolWindow" action by default. We have this shortcut and it's automatically defined in Settings > Keymap > Tool Windows > Cody, but no keys are assigned. The correct solution would be to assign keys to this action. My changes mimic this behavior by adding an action that manually displays the toolbar (outcome of this is two actions in settings doing the ~same thing).

This causes second problem: Pressing the shortcut key again doesn't hide the toolbar, but refocuses it again. It's huge problem IMO because this is last missing part for using Cody with keyboard only). Built-in toolbar shortcut gives that feature for free.

exigow:
About `Activation${toolbarName}ToolWindow`: https://intellij-support.jetbrains.com/hc/en-us/community/posts/206127339-Adding-a-keyboard-shortcut-to-a-custom-toolwindow

src/main/java/com/sourcegraph/cody/PromptPanel.kt
```diff
@@ -119,6 +119,8 @@ class PromptPanel(
     }
   }
 
+  fun focus() = textArea.requestFocusInWindow()
```
pkukielka, line 122: Is that really needed? Prompt panel has `AncestorListener` listener which focuses on it automatically when it becomes visible.

pkukielka (approved):
LGTM

src/main/java/com/sourcegraph/cody/PromptPanel.kt
exigow: Thanks! I'll verify that

TESTING.md
```diff
@@ -247,6 +247,9 @@ Test ideas:
 10. Start typing while being focused on Chat History to perform search-by-title.
 11. Open multiple chats and ask few simultaneous questions in several sessions at once.
 12. Open new chat with <kbd>Alt</kbd> + <kbd>=</kbd> shortcut (or <kbd>Option</kbd> + <kbd>=</kbd> on Mac).
+13. Open existing chat with shortcut <kbd>Alt</kbd> + <kbd>-</kbd> (or <kbd>Option</kbd> + <kbd>-</kbd> on Mac) and start typing question. Tab should be switched automatically on Chat.
```
mkondratek, line 262: 💟

TESTING.md
```diff
@@ -247,6 +247,9 @@ Test ideas:
...
+14. Second click <kbd>Alt</kbd> + <kbd-</kbd> should hide toolbar if focused (similar behavior as other toolbars).
```
mkondratek: It is "wool window" not "toolbar" - pls rename

src/main/java/com/sourcegraph/cody/PromptPanel.kt
exigow: @pkukielka I've commented this code and message panel is not always focused (especially when tool window is already opened). I need to add this method.

PR #507, opened 2024-02-06
Add llm selection dropdown

### LINKED ISSUE #384: GA: PLG pro users can choose between the supported LLMs for a given chat (but they cannot change LLMs within a single chat)

Sa1to:
It fixes https://github.com/sourcegraph/jetbrains/issues/384 https://github.com/sourcegraph/jetbrains/issues/387 https://github.com/sourcegraph/jetbrains/issues/389
## Test plan
Scenario 1
* Login as Pro user
* Create new chat
* User is able to change default LLM

Scenario 2
* Login as Pro user
* Open chat with already sent messages
* User see default LLM but is not able to change it

Scenario 3
* Login as Free user
* Open new chat
* User doesn't see LLM dropdown

Scenario 4
* Login as Enterprise user
* Open new chat 
* User see default LLM but is not able to change it
*


src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -174,9 +178,21 @@ private constructor(
...
+          messages.add(message)
+          chatPanel.addOrUpdateMessage(message, shouldAddBlinkingCursor = true)
+          HistoryService.getInstance().update(internalId, messages, selectedModel)
+        }
+      }
+    } else {
+      messages.add(message)
+      chatPanel.addOrUpdateMessage(message, shouldAddBlinkingCursor = true)
+      HistoryService.getInstance().update(internalId, messages)
+    }
```
exigow: I think it shouldn't be possible for user to send any message without a selected model (from logical perspective). Is this branching (if-else) really necessary here?

Perhaps we can generalize this problem? This fragment is IMO smell: `selectedModel = chatPanel.modelDropdown.selectedItem?.toString() ?: ""`. Why not `... ?: "anthropic/claude-2.0"`?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -190,6 +206,7 @@ private constructor(
 
   companion object {
     private val logger = LoggerFactory.getLogger(AgentChatSession::class.java)
+    private const val defaultModel = "anthropic/claude-2.0"
```
exigow: NIT: top-level immutable properties should be uppercase (logger is mutable, DEFAULT_MODEL is not).

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
Sa1to: For the second part i agree but regarding the first part: this if/esle branch is here because we need to set model only when first message is sent. For all other messages we need to use already selected model.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -47,10 +51,31 @@ class ChatPanel(project: Project, val chatSession: ChatSession) :
     add(lowerPanel, BorderLayout.SOUTH)
   }
 
+  fun addModelDropdown(project: Project, sessionId: String, selectedModel: String) {
+    CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
+      agent.server.isCurrentUserPro().thenApply { isUserPro ->
```
pkukielka: Can we change that `getIsCurrentUserPro` function to return `Boolean` and not `Boolean`?
We can just add `?: false` at the end of it.
Then `if (isUserPro == true)` will turn into `if (isUserPro)` here and in other places where we use it.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -47,10 +51,31 @@ class ChatPanel(project: Project, val chatSession: ChatSession) :
...
+        if (isUserPro == true) {
+          add(modelDropdown, BorderLayout.NORTH)
+          if (selectedModel == "") {
+            val chatModels = agent.server.chatModels(ChatModelsParams(sessionId))
+            chatModels.thenApply { response ->
+              response.models.forEach { model -> modelDropdown.addItem(model.model) }
```
pkukielka: That need to be run in EDT, no?

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -47,10 +51,31 @@ class ChatPanel(project: Project, val chatSession: ChatSession) :
     add(lowerPanel, BorderLayout.SOUTH)
   }
 
+  fun addModelDropdown(project: Project, sessionId: String, selectedModel: String) {
+    CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
+      agent.server.isCurrentUserPro().thenApply { isUserPro ->
+        if (isUserPro == true) {
+          add(modelDropdown, BorderLayout.NORTH)
```
pkukielka: EDT?

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -67,7 +70,10 @@ class CodyToolWindowContent(private val project: Project) {
 
   fun removeAllChatSessions() {
     AgentChatSessionService.getInstance(project).removeAllSessions()
-    switchToChatSession(AgentChatSession.createNew(project))
+    UIUtil.invokeLaterIfNeeded {
+      chatContainerPanel.removeAll()
+      chatContainerPanel.add(JBLabel(AnimatedIcon.Default()))
```
mkondratek: I do not like that loading screen on account change. Why do we need it?  We do have a default model? There is no loading screen on startup. If we need it to wait for the models list why don't we just use the default (since we define it) OR display it later like for startup?

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
```diff
@@ -45,7 +45,7 @@ interface CodyAgentServer {
   @JsonRequest("featureFlags/getFeatureFlag")
   fun evaluateFeatureFlag(flagName: GetFeatureFlag): CompletableFuture<Boolean?>
 
-  @JsonRequest("graphql/currentUserIsPro") fun isCurrentUserPro(): CompletableFuture<Boolean?>
+  @JsonRequest("graphql/currentUserIsPro") fun isCurrentUserPro(): CompletableFuture<Boolean>
```
mkondratek, line 48: what if the agent `return null`? is it possible?

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -23,6 +23,17 @@ class HistoryService(private val project: Project) :
     synchronized(listeners) { listeners += listener }
   }
 
+  @Synchronized
+  fun updateChatMessages(
```
mkondratek, line 27: is it ok we do writing operations in a synchronized method within EDT?

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
```diff
@@ -29,7 +31,13 @@ class AccountSettingChangeListener(project: Project) : ChangeListener(project) {
               CodyToolWindowContent.executeOnInstanceIfNotDisposed(project) {
                 removeAllChatSessions()
               }
-              CodyAgentService.getInstance(project).restartAgent(project)
+              CodyAgentService.getInstance(project).restartAgent(project).thenApply {
+                CodyToolWindowContent.executeOnInstanceIfNotDisposed(project) {
+                  UIUtil.invokeLaterIfNeeded {
```
mkondratek: I'd rather invert this wrapping ➡️ 
`executeOnInstanceIfNotDisposed` inside `invokeLaterIfNeeded`

project could dispose before `invokeLaterIfNeeded` executes

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -47,12 +56,71 @@ class ChatPanel(project: Project, chatSession: ChatSession) :
...
+                      else model.displayName))
+            }
+            modelDropdown.isEnabled = false
+          }
+        }
+      } else {
+        agent.server.isCurrentUserPro().thenApply { isUserPro ->
+          if (isUserPro) {
+            UIUtil.invokeLaterIfNeeded {
+              modelDropdown.setRenderer(CodyModelComboboxRenderer())
```
mkondratek: Use property access syntax

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -47,12 +56,71 @@ class ChatPanel(project: Project, chatSession: ChatSession) :
     lowerPanel.add(promptPanel)
     lowerPanel.add(contextView)
     add(lowerPanel, BorderLayout.SOUTH)
+    //    chatSession.getSessionId()?.let { addModelDropdown(project, it, selectedModel)}
+  }
+
+  fun addModelDropdown(project: Project, sessionId: String, selectedModel: ChatModel?) {
```
mkondratek, line 63: this method is a mess 😕

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -177,9 +184,31 @@ private constructor(
     if (messages.lastOrNull()?.id == message.id) {
       messages.removeLast()
     }
-    messages.add(message)
-    chatPanel.addOrUpdateMessage(message)
-    HistoryService.getInstance(project).updateChatMessages(internalId, messages)
+    if (messages.size == 0) {
+      val displayModelName = (chatPanel.modelDropdown.selectedItem as CodyModelComboboxItem).name
```
mkondratek:
```
java.lang.NullPointerException: null cannot be cast to non-null type com.sourcegraph.cody.ui.CodyModelComboboxItem
	at com.sourcegraph.cody.chat.AgentChatSession.addMessage(AgentChatSession.kt:188)
	at com.sourcegraph.cody.chat.AgentChatSession.sendMessage(AgentChatSession.kt:86)
	at com.sourcegraph.cody.PromptPanel.didSubmitChatMessage(PromptPanel.kt:150)
	at com.sourcegraph.cody.PromptPanel.didUseShortcut(PromptPanel.kt:134)
	at com.sourcegraph.cody.PromptPanel.access$didUseShortcut(PromptPanel.kt:31)
	at com.sourcegraph.cody.PromptPanel$6.actionPerformed(PromptPanel.kt:114)
```

1. `runIde`
2. having dot com account 
3. write something (before the combo box loads)

mkondratek (dismissed):
Commands do not work 😕

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
Sa1to: When we are switching an account we need to load LLM dropdown before user is able to send a message. If i use default model (hardcoded one)  it might not be in line with default model for enterprise users...

pkukielka: That is not exactly true. If you use default model (just by not specifying it) it will always pick correct default.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -37,25 +40,28 @@ private constructor(
...
         messages
             .toList()
             .dropWhile { it.speaker == Speaker.ASSISTANT }
             .fold(emptyList<ChatMessage>()) { acc, msg ->
               if (acc.lastOrNull()?.speaker == msg.speaker) acc else acc.plus(msg)
             }
-    val restoreParams = ChatRestoreParams(model, messagesToReload, UUID.randomUUID().toString())
+    val restoreParams =
+        ChatRestoreParams(
+            selectedModel?.agentName ?: "", messagesToReload, UUID.randomUUID().toString())
```
pkukielka: Is `""` really correct, shouldn't it be null to use the default? Have you tested this path?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -177,9 +183,41 @@ private constructor(
     if (messages.lastOrNull()?.id == message.id) {
       messages.removeLast()
     }
-    messages.add(message)
-    chatPanel.addOrUpdateMessage(message)
-    HistoryService.getInstance(project).updateChatMessages(internalId, messages)
+    if (messages.size == 0) {
```
pkukielka: Please encapsulate getting the model into separate method (if needed can be just with callback, or it can return `CompletableFuture`)

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -177,9 +183,41 @@ private constructor(
...
+                  WebviewMessage(command = "chatModel", model = selectedModel?.agentName)
+                )
+              )
+              .get()
+            messages.add(message)
+            chatPanel.addOrUpdateMessage(message)
+            ApplicationManager.getApplication().invokeLater {
+              HistoryService.getInstance(project)
+                .updateChatMessages(
+                  internalId, messages, selectedModel?.displayName ?: DEFAULT_MODEL.displayName
```
pkukielka: That is kind of incorrect. If we don't know what model we selected, in case of `selectedModel` being null, shouldn't we also serialise it as a null?
And then when restored it will use the default again (and you must support this path anyway for historic reasons).

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -177,9 +183,41 @@ private constructor(
...
+                )
+            }
+          }
+        }
+      }else {
```
pkukielka: Reformat

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
Sa1to: We can get null from an agent but If we get null then `isCurrentUserPro != true` I guess both are fine i dont really care 🤷   
https://github.com/sourcegraph/jetbrains/pull/507#discussion_r1481033113

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -49,10 +58,68 @@ class ChatPanel(project: Project, chatSession: ChatSession) :
...
+      val activeAccountType = CodyAuthenticationManager.instance.getActiveAccount(project)
+      //    Display available model for Enterprise user without possibility to change
+      if (activeAccountType != null && !activeAccountType.isDotcomAccount()) {
+        val chatModels = agent.server.chatModels(ChatModelsParams(sessionId))
+        chatModels.thenApplyAsync { response ->
+          ApplicationManager.getApplication().invokeLater {
+            modelDropdown.setRenderer(CodyModelComboboxRenderer())
+            add(modelDropdown, BorderLayout.NORTH)
+            response.models.forEach { modelProvider ->
+              val model = ChatModel.fromAgentName(modelProvider.model) ?: ChatModel.UNKNOWN_MODEL
```
pkukielka: Compiler tells me that:
> Task :compileKotlin
w: file:///Users/pkukielka/Work/sourcegraph/jetbrains/src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt:72:72 Elvis operator (?:) always returns the left operand of non-nullable type ChatModel
w: file:///Users/pkukielka/Work/sourcegraph/jetbrains/src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt:92:68 Elvis operator (?:) always returns the left operand of non-nullable type ChatModel

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -233,7 +282,10 @@ private constructor(
     @RequiresEdt
     fun createFromState(project: Project, state: ChatState): AgentChatSession {
       val sessionId = createNewPanel(project) { it.server.chatNew() }
-      val chatSession = AgentChatSession(project, sessionId, state.internalId!!)
+      val selectedModel = state.model?.let { ChatModel.fromDisplayName(it) } ?: DEFAULT_MODEL
```
pkukielka: Please at least ass a TODO there.
This might be incorrect in some scenarios, you should not assume what default model is.
Either fetch it from agent or change agent API to not require it in the first place.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -49,10 +58,68 @@ class ChatPanel(project: Project, chatSession: ChatSession) :
     add(lowerPanel, BorderLayout.SOUTH)
   }
 
+  fun addModelDropdown(project: Project, sessionId: String, selectedModel: ChatModel?) {
+    CodyAgentService.applyAgentOnBackgroundThread(project) { agent ->
+      val activeAccountType = CodyAuthenticationManager.instance.getActiveAccount(project)
+      //    Display available model for Enterprise user without possibility to change
+      if (activeAccountType != null && !activeAccountType.isDotcomAccount()) {
```
pkukielka, line 67: Those two branches of logic are like 100% duplicate?
And the only difference is `modelDropdown.isEnabled = false`?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
Sa1to: Actually i will remove this default since selectedModel is not required in `AgentChatSession` constructor

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -49,10 +59,57 @@ class ChatPanel(project: Project, chatSession: ChatSession) :
...
+          }
+        }
+      }
+    }
+  }
+
+  private fun addAllAvailableModelsToTheDropdown(agent: CodyAgent, sessionId: String) {
+    val chatModels = agent.server.chatModels(ChatModelsParams(sessionId))
+    chatModels.thenApplyAsync { response ->
+      modelDropdown.setRenderer(CodyModelComboboxRenderer())
```
pkukielka: This shouldn't be on the EDT?

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
Sa1to: I am getting `Access is allowed from Event Dispatch Thread (EDT) only` when i try to execute it on background thread...

pkukielka (approved):
LGTM,

Please make sure you also updated TESTING.MD with all corner cases.

PR #526, opened 2024-02-07
Fix uri for autocomplete

mkondratek:
The recent changes (https://github.com/sourcegraph/jetbrains/pull/367) broke autocomplete. The uri was string before but not it is properly serialized 😅 Agent expect the param to be string (not an object). Let's keep the serialisation for now but just pass string as param.

## Test plan
- autocomplete works

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -45,6 +45,7 @@ import com.sourcegraph.utils.CodyFormatter
 import difflib.Delta
 import difflib.DiffUtils
 import difflib.Patch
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -199,14 +200,14 @@ class CodyAutocompleteManager {
     val params =
         if (lookupString.isNullOrEmpty())
             AutocompleteParams(
-                virtualFile.path,
+                Paths.get(virtualFile.path).toUri().path,
```
mkondratek, line 203: Probably redundant wrapping but I wanted to make sure I get the same thing as it was in the constructor (`Paths.get(filePath).toUri()`).

pkukielka (approved):
LGTM

PR #520, opened 2024-02-07
Fix: Context files are not showing up in the chat

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/328.

Requires https://github.com/sourcegraph/cody/pull/3068.

## Test plan 
1. Write to chat / Use commands
2. Verify their replies contain context files when applicable 

---

Note that for now there is no persistence for context files.

PR #535, opened 2024-02-07
fix: Missing context files

### LINKED ISSUE #328: Context files are not showing up in the chat

### LINKED ISSUE #57378: JetBrains: context files in chat are not clickable
The filenames in the "Read N files" list are not clickable

Image

It would be nice if the user could open these files by clicking on the items.

mkondratek:
Reattempt of https://github.com/sourcegraph/jetbrains/pull/520.

Fixes https://github.com/sourcegraph/sourcegraph/issues/57378.
Fixes https://github.com/sourcegraph/jetbrains/issues/328.

## Test plan 
- test chat, test chat after restart, test commands


src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -140,15 +148,35 @@ private constructor(
             extensionMessage.isMessageInProgress == false) {
           getCancellationToken().dispose()
         } else {
-          if (lastMessage?.text != null && extensionMessage.chatID != null) {
-            addAssistantResponseToChat(lastMessage.text, lastMessage.displayText)
+
+          if (extensionMessage.chatID != null) {
+            if (prevLastMessage != null) {
+              if (lastMessage?.contextFiles != messages.lastOrNull()?.contextFiles) {
+                val index = restoredMessagesCount + extensionMessage.messages.count() - 2
```
pkukielka: I was always confused by that `- 2` but I finally get it.
Won't it be the same (but IMHO more easier to understand) if instead you will use `messages.count() - 1`?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -173,12 +201,15 @@ private constructor(
...
+    val messageToUpdate = messages.getOrNull(index)
+    if (messageToUpdate != null) {
+      messages[index] = message
+    } else {
+      messages.add(message)
     }
-    messages.add(message)
-    chatPanel.addOrUpdateMessage(message)
+    chatPanel.addOrUpdateMessage(
+        message, index, shouldAddBlinkingCursor = message.actualMessage().isBlank())
```
pkukielka, line 212: That have one downside, it will start showing blinking cursor just when we get first blank response from the assistant, right?
If e.g. agent is restarting that could be a few sec.
Maybe it would be better to add it without any conditions later, under:

```kotlin
 chatSession.addMessageAtIndex(
           ChatMessage(
               Speaker.ASSISTANT,
               Source.CHAT,
               text = "",
               displayText = "",
           ),
           chatSession.messages.count())
```
I think here we know we surely need it, no?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -225,25 +256,41 @@ private constructor(
...
 
-      chatSession.addMessage(ChatMessage(Speaker.HUMAN, commandId.displayName))
+      chatSession.addMessageAtIndex(
+          ChatMessage(
+              Speaker.HUMAN,
+              source = commandId.source,
+              commandId.displayName,
+          ),
+          chatSession.messages.count())
+      chatSession.addMessageAtIndex(
```
pkukielka, line 267: Can we force `shouldAddBlinkingCursor = true` there?
It will guarantee that cursor started blinking even if there is no response from the agent yet, it will be more responsive in some scenarios (e.g restart of the agent, slow LLM, bad connectivity)

src/main/kotlin/com/sourcegraph/cody/ui/AccordionSection.kt
```diff
@@ -28,6 +28,8 @@ class AccordionSection(title: String) : JPanel() {
         contentPanel.isVisible = true
         toggleButton.text = createToggleButtonHTML(sectionTitle, false)
       }
+      revalidate()
```
pkukielka: What scenario was broken without that?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: `extensionMessage.messages` always contains the full transcript of a given session. 

Hence, 
`messages.count() - 1` is always the index of the last element
and 
`messages.count() - 2` is always the index of the element before the last element 

That's all :smiley:

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: ok, but I tested it with no interent connection and it does not show the cursor (removes it immedietaly). 

```
cancellationToken.onFinished {
        removeBlinkingCursor()
        getLastMessage()?.onPartFinished()
      }
    }
```

`cancellationToken` removes it when there is not internet

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -28,6 +28,7 @@ private constructor(
     private val project: Project,
     newSessionId: CompletableFuture<SessionId>,
     private val internalId: String = UUID.randomUUID().toString(),
+    private val restoredMessagesCount: Int = 0
```
pkukielka: It seems to be never updated, I guess it's not needed at all now?

pkukielka (approved):
LGTM except one comment I left.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -225,25 +262,45 @@ private constructor(
...
     fun createFromState(project: Project, state: ChatState): AgentChatSession {
       val sessionId = createNewPanel(project) { it.server.chatNew() }
-      val chatSession = AgentChatSession(project, sessionId, state.internalId!!)
-      for (message in state.messages) {
+      val chatSession =
+          AgentChatSession(
+              project,
+              sessionId,
+              state.internalId!!,
+              restoredMessagesCount = state.messages.count())
```
mkondratek: @pkukielka, pls take a look - the value is not updated, it's initialized here

exigow (approved):
Looks good! :+1: I tested this on different repositories. It also works well with @.

pkukielka (approved):
LGTM

Issue #439, opened 2024-01-30
bug: IntelliJ plugin filled up disk with agent files


gabtorre:
### Installation Information

Cody plugin version 5.2.20247 

IntelliJ IDEA 2023.3.2 (Ultimate Edition)
Build #IU-233.13135.103, built on December 20, 2023
Runtime version: 17.0.9+7-b1087.9 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 14.3

### Describe the bug

The plugin seemed stuck in a loop, leading to excessive disk usage. Additionally, it created large temporary copies of the agent file during this process, with many of these logs:

```
2024-01-30 09:34:58,888 [86661021]   INFO - #CodyAgent - extracting Cody agent binary to /var/folders/z4/qq9brd3x7y9gxs52nc_3x6sm0000gn/T/cody-agent8534747379058237572
2024-01-30 09:34:58,920 [86661053]   INFO - #CodyAgent - starting Cody agent /var/folders/z4/qq9brd3x7y9gxs52nc_3x6sm0000gn/T/cody-agent8534747379058237572
```

To mitigate the issue, the user cleared all the files associated with the plugin, restarted the macOS, and then reenabled the plugin. After these actions, the plugin behaves correctly, with only two copies of the agent created.

It seems like the plugin may not clean up after itself properly, as indicated by the excessive number of temporary copies created during the loop.

### Expected behavior

The plugin cleans up temporary files appropriately to prevent excessive disk usage.

### Additional context (logs, images, etc)

Additional context: https://sourcegraph.slack.com/archives/C02842Z6YAK/p1706567669821759

```
2024-01-30 09:34:53,884 [86656017]   INFO - #CodyAgent - extracting Cody agent binary to /var/folders/z4/qq9brd3x7y9gxs52nc_3x6sm0000gn/T/cody-agent15487046631569367214
2024-01-30 09:34:53,918 [86656051]   INFO - #CodyAgent - starting Cody agent /var/folders/z4/qq9brd3x7y9gxs52nc_3x6sm0000gn/T/cody-agent15487046631569367214
2024-01-30 09:34:55,094 [86657227]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ getRgPath: Did not find bundled `rg`.
[...]

PR #542, opened 2024-02-08
Delete agent binary file on exit

olafurpg:
Addresses https://github.com/sourcegraph/jetbrains/issues/439

Previously, the JetBrains plugin created a new temporary file every time we started a new agent process and only deleted this file when the IDE shut down. This buggy behavior has existed since Septemeber and we haven't received user complaints about it, likely because the plugin was conservative about restarting the agent process. Recent changes introduced retry logic where we could start a new agent process in a loop, and never delete the underlying binary. Some users have reported that this bug caused the plugin to take up unreasonable amounts of disk space. I'm sorry for users who were impacted by this behavior.

This PR addresses the problem with a new callback hook that deletes the agent binary after the agent process finishes running. This change means that we never have more than one agent binary sitting on the disk per IntelliJ project. This should prevent the catastrophic situation some users have reported.

## Test plan
I manually tested this change by adding a println to when the file gets deleted, started IntelliJ, triggered the "Restart Cody Agent" action and confirmed that the file got deleted.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
@@ -8,6 +8,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.SystemInfoRt
 import com.intellij.util.system.CpuArch
 import com.sourcegraph.cody.agent.protocol.*
+import com.sourcegraph.cody.vscode.CancellationToken
 import com.sourcegraph.config.ConfigUtil
 import java.io.*
 import java.net.Socket
@@ -128,6 +129,7 @@ private constructor(
       if (shouldConnectToDebugAgent()) {
         return connectToDebugAgent()
       }
+      val token = CancellationToken()
       val command: List<String> =
           if (System.getenv("CODY_DIR") != null) {
             val script = File(System.getenv("CODY_DIR"), "agent/dist/index.js")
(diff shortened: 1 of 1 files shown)


olafurpg:
Looking at the original issue, the current diff may not fix the disk issue because the error is happening in `Files.copy()`

```
024-01-30 08:17:01,833 [81983966]   WARN - #CodyAgent - Unable to start Cody agent
com.sourcegraph.cody.agent.CodyAgentException: failed to create agent binary
Caused by: java.nio.file.FileSystemException: /Users/foo/Library/Application Support/JetBrains/IntelliJIdea2023.3/plugins/Sourcegraph/agent/agent-macos-arm64 -> /var/folders/z4/qq9brd3x7y9gxs52nc_3x6sm0000gn/T/cody-agent7707303340069661988: No space left on device
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:100)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixCopyFile.copyFile(UnixCopyFile.java:256)
	at java.base/sun.nio.fs.UnixCopyFile.copy(UnixCopyFile.java:603)
	at java.base/sun.nio.fs.UnixFileSystemProvider.copy(UnixFileSystemProvider.java:257)
	at java.base/java.nio.file.Files.copy(Files.java:1305)
	at CodyAgent$Companion.agentBinary(CodyAgent.kt:180)
```

I pushed a commit that may address this.

pkukielka (approved):
LGTM

olafurpg:
I want to add that this PR does *not* fix the root bug where we are restarting the agent process in loop, which will continue to eat up CPU resources and make the computer unresponsive.

PR #552, opened 2024-02-08
Add line ranges to the context files

### LINKED ISSUE #339: No indicator in the chat what command I triggered after triggering explain/smell/test
We're missing this indicator here above the reply

CleanShot 2024-01-23 at 21 00 02@2x

mkondratek:
Merging queue 😅 
image

Fixes https://github.com/sourcegraph/jetbrains/issues/339.

## Test plan
- run a command / ask Cody about something related to your code
- review the context files

image

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
@@ -15,6 +15,7 @@ data class ContextFileFile(
     override val uri: URI,
     override val repoName: String?,
     override val revision: String?,
+    val range: Range? = null,
 ) : ContextFile() {
   override val type: String = "file"
 }
@@ -27,8 +28,8 @@ val contextFileDeserializer =
           val uri = context.deserialize<URI>(jsonObject["uri"], URI::class.java)
           val repoName = jsonObject["repoName"]?.asString
           val revision = jsonObject["revision"]?.asString
-
-          ContextFileFile(uri, repoName, revision)
+          val range = gsonMapper.fromJson(jsonObject["range"], Range::class.java)
+          ContextFileFile(uri, repoName, revision, range)
         }
 
         // TODO(beyang): should throw an exception here, but we don't because the context field is
@@ -36,6 +37,8 @@ val contextFileDeserializer =
         else -> null
       }
     }
+
+private val gsonMapper = GsonBuilder().create()
 
 val uriDeserializer =
     JsonDeserializer { jsonElement: JsonElement?, _: Type, _: JsonDeserializationContext ->
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
@@ -18,6 +18,22 @@ data class ContextFileFile(
     val range: Range? = null,
 ) : ContextFile() {
   override val type: String = "file"
+
+  fun getLinkActionText(projectPath: String?): String {
+    val intelliJRange = range?.intellijRange()
+    val localPath = uri.path.removePrefix(projectPath ?: "")
+
+    return buildString {
+      append("@$localPath")
+      if (intelliJRange != null) {
+        if (intelliJRange.first != intelliJRange.second) {
+          append(":${intelliJRange.first}-${intelliJRange.second}")
+        } else {
+          append(":${intelliJRange.first}")
+        }
+      }
+    }
+  }
 }
 
 val contextFileDeserializer =
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
@@ -6,9 +6,7 @@ import com.intellij.openapi.fileEditor.OpenFileDescriptor
 import com.intellij.openapi.project.DumbAwareAction
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.LocalFileSystem
-import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.ui.components.AnActionLink
-import com.intellij.util.concurrency.annotations.RequiresBackgroundThread
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.ui.JBInsets
 import com.intellij.util.ui.JBUI
@@ -23,7 +21,6 @@ import com.sourcegraph.cody.chat.ChatUIConstants.TEXT_MARGIN
 import com.sourcegraph.cody.ui.AccordionSection
 import java.awt.BorderLayout
 import java.awt.Insets
-import java.nio.file.Path
 import java.nio.file.Paths
 import javax.swing.JPanel
 import javax.swing.border.EmptyBorder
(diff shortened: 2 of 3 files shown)


pkukielka:
I'm wondering why it always report whole file as the context, even if you specifically select only a bit of the text. It does not seem valid.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -73,14 +82,32 @@ class ContextFilesPanel(
...
+      projectRelativeFilePath: String
+  ): String {
+    val startLine = contextFileFile?.range?.start?.line
+    val endLine = contextFileFile?.range?.end?.line
+    return buildString {
+      append("@$projectRelativeFilePath")
+      if (startLine != null && endLine != null) {
+        if (startLine != endLine) {
+          append(":$startLine-$endLine")
+        }
```
pkukielka, line 89:
```kotlin
else {
    append(":$startLine")
}
```

pkukielka: Actually, I think even more correct approach would be:

```kotlin

if (startLine != null && endLine != null && startLine != endLine) {
  append(":$startLine-$endLine")
} else if (startLine != null) {
  append(":$startLine")
}  else if (endLine != null) {
  append(":$endLine")
}
```

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
```diff
@@ -56,15 +64,11 @@ val uriDeserializer =
     }
 
 val uriSerializer = JsonSerializer { uri: URI, type, context ->
-  if (uri == null) {
```
pkukielka: Why that change? You seem like you are removing useful safeguard?

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -122,8 +122,8 @@ class ChatPanel(project: Project, chatSession: ChatSession) :
   ) {
     if (messagesPanel.componentCount == 1) {
       modelDropdown.isEnabled = false
+      promptPanel.updateEmptyTextAfterFirstMessage()
```
pkukielka: Why that change?
Name of the function does not reflect what it does anymore.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextFile.kt
mkondratek: whoops... good point. intellij suggestion mislead me 

image

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
mkondratek: why? `componentCount == 1` when we have only the hello message

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -54,48 +60,85 @@ class ContextFilesPanel(
...
 
+  private fun getActionTextForFileLinkAction(
+      contextFileFile: ContextFileFile?,
+      projectRelativeFilePath: String
+  ): String {
+    val (startLine, endLine) =
+        contextFileFile?.range?.let {
+          // We need to .plus(1) since the ranges use 0-based indexing
+          // but IntelliJ presents it as 1-based indexing.
+          Pair(it.start.line.plus(1), it.end.line.plus(1))
```
pkukielka: Cannot you use `Range` instead of pair?
You could even add method like `intellijRange` on the `Range` class so you won't need to do that mapping there, and just call  `intellijRange` where you need it.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -54,48 +60,85 @@ class ContextFilesPanel(
...
-        contextFiles
-            .map {
-              if (it.repoName != null) {
-                "${project.basePath}/${it.uri.path}"
-              } else {
-                Paths.get(it.uri.withFragment(null).withQuery(null)).absolutePathString()
-              }
-            }
-            .toSet()
+    val contextFileNames = mutableMapOf<String, ContextFile>()
```
pkukielka: Can you make `Path` a key?
It will be more clear from the interface what this do and looks like you have `Path` anyway and you convert it to String.
Or maybe that `absolutePathString` is important?

pkukielka: Either way a bit cleaner way to do it would be using `groupBy` + `minBy` I think.
But it's a nitpick.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
pkukielka: I'm blind. You actually fixed it, I somehow thought it was other way around.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -54,53 +61,91 @@ class ContextFilesPanel(
...
 
+  private fun getActionTextForFileLinkAction(
+      contextFileFile: ContextFileFile?,
+      projectRelativeFilePath: String
+  ): String {
+    val (startLine, endLine) = contextFileFile?.range?.intellijRange() ?: Pair(null, null)
+
+    return buildString {
+      append("@$projectRelativeFilePath")
+      if (startLine != null && endLine != null) {
```
pkukielka: You resolved my comment but problem I mentioned is not addressed I think.

mkondratek:
Related PR: https://github.com/sourcegraph/cody/pull/3209

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
mkondratek: fixed now

build.gradle.kts
```diff
@@ -199,7 +199,7 @@ tasks {
     return destinationDir
   }
 
-  val codyCommit = "0e908a782b4b23423a362f1f56d3745575a613f5"
+  val codyCommit = "f16ebbe747c4a2a9f00cf7cd46004414ff333a37"
```
mkondratek: ❗

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -39,68 +36,58 @@ class ContextFilesPanel(
...
-  @RequiresBackgroundThread
-  private fun updateFileList(contextFileNames: Set<String>) {
-    val filesAvailableInEditor =
-        contextFileNames
-            .map(Paths::get)
-            .mapNotNull(VirtualFileManager.getInstance()::findFileByNioPath)
-            .toList()
+  fun updateContentWith(contextFiles: List<ContextFile>?) {
+    val contextFileFiles =
+        contextFiles?.mapNotNull { it as? ContextFileFile }?.filter { it.repoName == null }
```
mkondratek, line 41: the handling for `it.repoName != null` is in the next PR: https://github.com/sourcegraph/jetbrains/pull/559

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -83,6 +85,40 @@ class SettingsMigration : Activity {
...
   }
 
+  private fun migrateLlms(project: Project) {
+    HistoryService.getInstance(project)
+        .state
+        .chats
+        .filter { it.llm == null }
+        .forEach {
+          val (model, provider, title) =
+              modelToProviderAndTitle.getOrDefault(it.model, Triple("", "Unknown", "Unknown"))
```
pkukielka: That is my only doubt. I think you should stick to the `null`s.
Then this code will work properly:

```kt
val modelFromState =
             state.llm?.let {
               ChatModelsResponse.ChatModelProvider(
                   default = it.model == null,
                   codyProOnly = false,
                   provider = it.provider ?: "Unknown",
                   title = it.title ?: "Unknown",
                   model = it.model ?: "")
             }
```

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -297,28 +293,48 @@ private constructor(
...
+      val agentChatSession = CompletableFuture<AgentChatSession>()
+      createNewPanel(project) { codyAgent ->
+        val innerSessionId = codyAgent.server.chatNew()
+        val modelFromState =
+            state.llm?.let {
+              ChatModelsResponse.ChatModelProvider(
+                  default = it.model == null,
+                  codyProOnly = false,
+                  provider = it.provider ?: "Unknown",
+                  title = it.title ?: "Unknown",
```
pkukielka: Maybe we should just say "Default"?
It will be less annoying for user than seeing "Unknown" (which looks like UI bug)

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
mkondratek: will be fixed in https://github.com/sourcegraph/jetbrains/pull/753

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: will be fixed in https://github.com/sourcegraph/jetbrains/pull/753

pkukielka (approved):
LGTM

exigow (approved):
Looks good! :+1: Clickable links works correctly even when some parts of code is visually collapsed.

PR #553, opened 2024-02-08
Mention @ in the placeholder message

mkondratek:
## Test plan
1. runIde
2. see Cody tool window

|   |   |
| --- | --- |
| image | image |

### DIFF
--- a/src/main/resources/CodyBundle.properties
+++ b/src/main/resources/CodyBundle.properties
@@ -106,3 +106,5 @@ popup.select-chat=Select Chat
 popup.remove-chat=Remove Chat
 popup.remove-all-chats=Remove All Chats
 
+PromptPanel.ask-cody=Message (use @ to include files)...
+
--- a/src/main/resources/CodyBundle.properties
+++ b/src/main/resources/CodyBundle.properties
@@ -106,5 +106,5 @@ popup.select-chat=Select Chat
 popup.remove-chat=Remove Chat
 popup.remove-all-chats=Remove All Chats
 
-PromptPanel.ask-cody=Message (use @ to include files)...
-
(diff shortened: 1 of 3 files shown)


taylorsperry:
Thanks, @mkondratek ! I'll defer to @danielmarquespt on design, but I think this is already an improvement. Is it possible to suggest the open file as soon as I type `@` (like VS Code does)? I feel like you type `@` and expect something to happen / don't know what to do next. 
Screenshot 2024-02-08 at 4 47 34 PM

mkondratek:
Taylor, it's fixed by #555 🚀

src/main/resources/CodyBundle.properties
```diff
@@ -108,3 +108,5 @@ popup.select-chat=Select Chat
 popup.remove-chat=Remove Chat
 popup.remove-all-chats=Remove All Chats
 
+PromptPanel.ask-cody=Message (use @ to include files)...
```
pkukielka: @danielmarquespt WDYT?

To me that `"Message"` alone looks very incomplete.
I would prefer something like fix of old and new:
`"Ask a question about this code (you can use @ to include files)..."`
Unless that is too wordy?

mkondratek: I just tried to make it as similar to VS Code as possible

mkondratek: VS Code: 
image

danielmarquespt: Let's go with:

`Message (type @ to include specific files as context)`

if it is a follow-up, (low priority)

`Follow-up message (type @ to include specific files as context)`

mkondratek:
I applied the fixes:

|.  |.  |
| --- | --- |
| image | image |

Sa1to (approved):
LGTM

PR #559, opened 2024-02-09
Add support for remote context files

mkondratek:
Merging queue 😅 
image

VS Code example:
image

## Test plan
1. chat about the code, commands explain text
2. test link navigation

## The look

Local files:

https://github.com/sourcegraph/jetbrains/assets/19799111/620a2b4c-60d6-47ea-8c63-f116e1cda986

Remote files:
image

Remote and local:
image

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
@@ -1,10 +1,13 @@
 package com.sourcegraph.cody.agent.protocol
 
-typealias IntelliJRange = Pair<Int, Int>
+typealias RangePair = Pair<Int, Int>
 
 data class Range(val start: Position, val end: Position) {
 
   // We need to .plus(1) since the ranges use 0-based indexing
   // but IntelliJ presents it as 1-based indexing.
-  fun intellijRange(): IntelliJRange = IntelliJRange(start.line.plus(1), end.line.plus(1))
+  fun intellijRange(): RangePair = RangePair(start.line.plus(1), end.line.plus(1))
+
+  // The link to Sourcegraph Search on the other hand looks like this:
+  fun toSearchRange(): Pair<Int, Int> = RangePair(start.line.plus(1), end.line)
 }
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
@@ -9,5 +9,5 @@ data class Range(val start: Position, val end: Position) {
   fun intellijRange(): RangePair = RangePair(start.line.plus(1), end.line.plus(1))
 
   // The link to Sourcegraph Search on the other hand looks like this:
-  fun toSearchRange(): Pair<Int, Int> = RangePair(start.line.plus(1), end.line)
+  fun toSearchRange(): RangePair = RangePair(start.line.plus(1), end.line)
 }
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
@@ -14,7 +14,7 @@ class ContextFileActionLink(
     anAction: AnAction
 ) : AnActionLink("", anAction) {
   private val localFileBackground = JBColor(Color(182, 210, 242), Color(56, 85, 112))
-  private val isReferringToLocalFile = contextFileFile.repoName == null
+  private val isReferringToLocalFile = contextFileFile.isLocal()
 
   init {
     text = contextFileFile.getLinkActionText(project.basePath)
@@ -34,9 +34,9 @@ class ContextFileActionLink(
   }
 
   private fun ContextFileFile.getLinkActionText(projectPath: String?): String {
-    val theRange = if (repoName == null) range?.intellijRange() else range?.toSearchRange()
+    val theRange = if (isLocal()) range?.intellijRange() else range?.toSearchRange()
     val path =
-        if (repoName == null) {
+        if (isLocal()) {
           "@${uri.path.removePrefix(projectPath ?: "")}"
         } else {
           val repoCommitFile = uri.path.split("@", "/-/blob/")
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
@@ -2,17 +2,35 @@ package com.sourcegraph.cody.chat.ui
 
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.project.Project
+import com.intellij.ui.JBColor
 import com.intellij.ui.components.AnActionLink
 import com.sourcegraph.cody.agent.protocol.ContextFileFile
+import java.awt.Color
+import java.awt.Graphics
 
 class ContextFileActionLink(
     project: Project,
     contextFileFile: ContextFileFile,
     anAction: AnAction
 ) : AnActionLink("", anAction) {
+  private val localFileBackground = JBColor(Color(182, 210, 242), Color(56, 85, 112))
+  private val isReferringToLocalFile = contextFileFile.repoName == null
+
   init {
     text = contextFileFile.getLinkActionText(project.basePath)
     toolTipText = contextFileFile.uri.path
+  }
+
+  override fun paintComponent(g: Graphics) {
+    if (isReferringToLocalFile) {
+      g.color = localFileBackground
+
+      val fm = g.fontMetrics
+      val rect = fm.getStringBounds(text, g)
+      val textWidth = rect.width.toInt()
+      g.fillRect(0, 0, textWidth, height)
+    }
+    super.paintComponent(g)
   }
 
   private fun ContextFileFile.getLinkActionText(projectPath: String?): String {
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
@@ -32,29 +32,4 @@ class ContextFileActionLink(
     }
     super.paintComponent(g)
   }
-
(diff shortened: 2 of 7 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -40,67 +43,148 @@ class ContextFilesPanel(
...
-            .map(Paths::get)
-            .mapNotNull(VirtualFileManager.getInstance()::findFileByNioPath)
+        localFiles
+            .map {
+              val path = Paths.get(it.key.displayText)
+              val findFileByNioPath = VirtualFileManager.getInstance().findFileByNioPath(path)
+              findFileByNioPath ?: return@map null
+              Pair(findFileByNioPath, it.value as? ContextFileFile)
+            }
+            .filterNotNull()
```
pkukielka: You always create `Pair` instance, so `.filterNotNull()` will never filter anything out?

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -40,67 +43,148 @@ class ContextFilesPanel(
...
   }
 
   companion object {
     private val logger = Logger.getInstance(ContextFilesPanel::class.java)
   }
 }
+
+sealed class TypedContextFile(val displayText: String)
+
+class LocalContextFile(displayText: String) : TypedContextFile(displayText)
```
pkukielka: Is display text so me kind path or uri?
If so, name should reflect that, `displayTest` sounds a bit like content to me.
I understand it is what we display, and it is maybe not a real uri in any sense, but still...

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
pkukielka: Maybe:

```kotlin
.mapNotNull {
    val path = Paths.get(it.key.displayText)
    val fileByNioPath = VirtualFileManager.getInstance().findFileByNioPath(path)
    val contextFile =  it.value as? ContextFileFile

    if (fileByNioPath != null && contextFile != null) Pair(fileByNioPath, contextFile) 
    else null
}
```

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -40,67 +43,148 @@ class ContextFilesPanel(
   }
 
   @RequiresBackgroundThread
-  private fun updateFileList(contextFileNames: Set<String>) {
+  private fun updateFileList(
+      localFiles: Map<LocalContextFile, ContextFile>,
+      remoteFiles: Map<RemoteContextFile, ContextFile>
```
pkukielka: I would say lets discuss this offline, I think we have too many different files abstractions already.
`ContextFile`, `ContextFileFile`, `LocalContextFile`, `RemoteContextFile`, `VirtualFile`, nio `Path`, `URI`s,  and what not.
We will get lost in that pretty quick, I think we need to do some simplifications there.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -40,67 +43,148 @@ class ContextFilesPanel(
...
+    val startLine = contextFileFile?.range?.start?.line
+    val endLine = contextFileFile?.range?.end?.line
+    return buildString {
+      if (typedContextFile is LocalContextFile) {
+        append("@")
+      }
+      append(typedContextFile)
+      if (startLine != null && endLine != null) {
+        if (startLine != endLine) {
+          append(":$startLine-$endLine")
```
pkukielka: What if `startLine == endLine`? Don't we want to append `:$startLine` in such case?

src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
```diff
@@ -57,7 +57,10 @@ class AddRepositoryDialog(
         DialogValidationUtils.custom(
             repoUrlInputField,
             CodyBundle.getString("context-panel.add-repo-dialog.error-no-repo")) {
-              val codebaseName = convertGitCloneURLToCodebaseNameOrError(repoUrlInputField.text)
+              val codebaseName =
+                  runCatching { convertGitCloneURLToCodebaseNameOrError(repoUrlInputField.text) }
```
mkondratek, line 61: unrel: it was spamming the logs while entering the remote

src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
```diff
@@ -69,7 +72,9 @@ class AddRepositoryDialog(
         DialogValidationUtils.custom(
             repoUrlInputField,
             CodyBundle.getString("context-panel.add-repo-dialog.error-repo-already-added")) {
-              val codebaseName = convertGitCloneURLToCodebaseNameOrError(repoUrlInputField.text)
+              val codebaseName =
+                  runCatching { convertGitCloneURLToCodebaseNameOrError(repoUrlInputField.text) }
```
mkondratek, line 76: unrel: it was spamming the logs while entering the remote

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.sourcegraph.cody.agent.protocol.ContextFileFile
+import java.awt.Color
+import java.awt.Graphics
+
+class ContextFileActionLink(
+    project: Project,
+    contextFileFile: ContextFileFile,
+    anAction: AnAction
+) : AnActionLink("", anAction) {
+  private val localFileBackground = JBColor(Color(182, 210, 242), Color(56, 85, 112))
```
mkondratek, line 16: I tried to find this color somewhere in IntelliJ but I have not seen it. I guess we must declare ourselves.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
```diff
@@ -1,10 +1,13 @@
...
 
 data class Range(val start: Position, val end: Position) {
 
   // We need to .plus(1) since the ranges use 0-based indexing
   // but IntelliJ presents it as 1-based indexing.
-  fun intellijRange(): IntelliJRange = IntelliJRange(start.line.plus(1), end.line.plus(1))
+  fun intellijRange(): RangePair = RangePair(start.line.plus(1), end.line.plus(1))
+
+  // The link to Sourcegraph Search on the other hand looks like this:
+  fun toSearchRange(): Pair<Int, Int> = RangePair(start.line.plus(1), end.line)
```
pkukielka: Should return type be also `RangePair`?
Also, I would change name to `toSourcegraphSearchRange()` or even `toSourcegraphRange()`... search is too common word to be meaningful without a context.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
```diff
@@ -0,0 +1,60 @@
...
+
+      val fm = g.fontMetrics
+      val rect = fm.getStringBounds(text, g)
+      val textWidth = rect.width.toInt()
+      g.fillRect(0, 0, textWidth, height)
+    }
+    super.paintComponent(g)
+  }
+
+  private fun ContextFileFile.getLinkActionText(projectPath: String?): String {
```
pkukielka: Any reason why we cannot move this to `ContextFileFile` (with a bit more generic name like `getLinkText`)?

pkukielka: Also, if you can move it there it would be great to also add some tests for it.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -55,37 +51,43 @@ class ContextFilesPanel(
...
   @RequiresEdt
   private fun createFileWithLinkPanel(contextFileFile: ContextFileFile): JPanel {
-    val anAction = OpenLocalContextFileAction(project, contextFileFile)
-    val actionText = contextFileFile.getLinkActionText(project.basePath)
-    val goToFile = AnActionLink(actionText, anAction)
-    goToFile.toolTipText = actionText
+    val anAction =
+        object : DumbAwareAction() {
+          override fun actionPerformed(anActionEvent: AnActionEvent) {
+            if (contextFileFile.repoName == null) {
```
pkukielka: I see this check used multiple times, maybe we can add simple utility method for it like `isLocalFile` or `isRemoteFile`?

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
mkondratek: for me it sounds more logical to keep this method here - closer to the scope of it's usage (`ContextFileActionLink` class). `ContextFileFile` file is responsible only for data type definition (and serialisation).

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
mkondratek: done

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
mkondratek: done

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
pkukielka: Either way please add the tests :)

mkondratek: done!

pkukielka (approved):
LGTM

PR #567, opened 2024-02-09
Simple fix for index out of bounds

### LINKED ISSUE #565: bug: IntelliiJ crash report
### Installation Information

On version `5.2.21709` (slack thread with customer: https://sourcegraph.slack.com/archives/C039TJ9TJ2Y/p1706912853962849?thread_ts=1706910668.266109&cid=C039TJ9TJ2Y)

Original issue was hard disk space filling up

### Describe the bug

Customer reports (in addition to hard disk filling up) that IntelliJ crashes "once in a while"
Strack trace attached
[stack_trace.txt](https://github.com/sourcegraph/jetbrains/files/14223987/stack_trace.txt)

### Expected behavior

IDE client should not crash when using Cody

_No response_

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/565

## Test plan

Truth to be told I'm totally unable to reproduce this issue artifically.
Hopefully fix is simple enough to speak for itself.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -157,10 +157,8 @@ class CodyAutocompleteManager {
     val caretPositionInLine = offset - editor.document.getLineStartOffset(lineNumber)
     val originalText = editor.document.getText(TextRange(offset - caretPositionInLine, offset))
 
-    val prefixStartPosition = maxOf(originalText.lastIndexOf("."), originalText.lastIndexOf(" "), 0)
-    if (!lookupString.isNullOrEmpty() &&
-        !lookupString.startsWith(
-            originalText.subSequence(prefixStartPosition + 1, originalText.length))) {
+    val originalTextTrimmed = originalText.takeLastWhile { c -> c != '.' && !c.isWhitespace() }
+    if (!lookupString.isNullOrEmpty() && !lookupString.startsWith(originalTextTrimmed)) {
       logger.debug("Skipping autocompletion for lookup element due to not matching prefix")
       return
     }


beyang (approved):

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -158,7 +158,8 @@ class CodyAutocompleteManager {
     val originalText = editor.document.getText(TextRange(offset - caretPositionInLine, offset))
 
     val prefixStartPosition = maxOf(originalText.lastIndexOf("."), originalText.lastIndexOf(" "), 0)
-    if (!lookupString.isNullOrEmpty() &&
+    if (prefixStartPosition < originalText.length &&
```
olafurpg: Based on the original stack trace, this happens with empty strings

```
@ "".subSequence(1, 0)
java.lang.StringIndexOutOfBoundsException: begin 1, end 0, length 0
```


```suggestion
    if (prefixStartPosition + 1 < originalText.length &&
```

olafurpg (commented):
I would merge but it's 9pm Friday here so don't want to make any rushed decision

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
pkukielka: I was crashing on this line `originalText.subSequence(prefixStartPosition + 1, originalText.length))`
when `originalText.length = 0` and `prefixStartPosition + 1 = 1` (so `prefixStartPosition = 0`).
Thus I think condition I added is valid and sufficient to prevent crash.
That said it can still result in producing empty subSequence and in such case `startsWith` will be always try no matter what `lookupString` is.
I would say let's double check and merge on Monday.

olafurpg: Agree the current diff fixes the case when `originalText.length === 0`. Still, the logic reads like the bug below
```java
if (index < text.length) {
  text.charAt(index + 1)
}
```
My brain is just wired to not allow that kind of code.

Either way, a much cleaner fix is to use the `startsWith()` overload from the Kotlin stdlib that accepts a `startIndex` parameter, and remove `.subSequence()`. The original code was written in Java where that overload doesn't exist

olafurpg:
LGTM 👍🏻

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
pkukielka: That would be great idea, but it `Returns true if a substring of this string starting at the specified offset startIndex starts with the specified prefix.`
So `startIndex` would affect position in `lookupString` while we want it to affect a prefix (`originalText`).

Truth to be told I see at least one more bug there and I do not fully understand what that code is supposed to do exactly, so maybe let's discuss offline.

Sa1to (approved):
LGTM

PR #593, opened 2024-02-12
Fix symf reindexing error notification

### LINKED ISSUE #381: GA: If search indexing fails, users are notified with guidance for debugging

exigow:
Fixes #381

## Test plan

a) Throw exception during symf indexing
b) ...or use `codyCommit=` https://github.com/sourcegraph/cody/commit/9f6160ac02107c535f3d68cd63cef5d7195bd10c

## Demo

report-symf


src/main/kotlin/com/sourcegraph/cody/context/ui/ReindexButton.kt
```diff
@@ -29,10 +35,7 @@ class ReindexButton(private val project: Project) :
                     val cmd = CommandExecuteParams("cody.search.index-update", emptyList())
                     agent.server.commandExecute(cmd).get()
                   } catch (e: Exception) {
-                    val errMsg = e.message ?: e.toString()
-                    Messages.showErrorDialog(
-                        CodyBundle.getString("context-panel.error-message").fmt(errMsg),
-                        CodyBundle.getString("context-panel.error-title"))
+                    showNotification(e.message ?: e.toString())
```
pkukielka: Can we make that message more user friendly?
It should tell how to report a bug, ideally link to GitHub issue template.

exigow: Unfortunately, the exception `message` contains the entire stack trace instead of user friendly text. 

But sure, maybe it's possible to use CodyErrorSubmitter here instead of instructing user how to submit GH issue. I will think about it.

pkukielka: > Unfortunately, the exception message contains the entire stack trace instead of user friendly text.

I'm aware of that :) 
What I meant is that we can add any test we want, and we should not just print stack trace without telling user what to do.

exigow: Looks like an uncaught exception is recognized as IDE error, so it is handled by our error submitter. IMO this is better UX.

(see: new PR description + changes in code)

pkukielka (approved):
LGTM

PR #612, opened 2024-02-12
Do not set model explicitly for enterprise 

mkondratek:
## Test plan
1. Having enterprise account
2. chat with Cody


pkukielka (approved):
LGTM

PR #621, opened 2024-02-13
Fix restoring chat session with custom llm

Sa1to:
## Test plan
1. Create chat session with non-default llm
2. Write a message
3. Change account and then go back to initial account
4. Open pervious chat from the history
5. Ask in chat what model is used

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
@@ -52,7 +52,7 @@ private constructor(
     }
   }
 
-  fun restoreAgentSession(agent: CodyAgent, model: String? = null) {
+  fun restoreAgentSession(agent: CodyAgent, model: ChatModel? = null) {
     val messagesToReload =
         messages
             .toList()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -61,16 +61,14 @@ private constructor(
...
 
-    val modelFromState = model?.let { ChatModel.fromDisplayNameNullable(it) }
-
     val selectedItem =
         chatPanel.modelDropdown.selectedItem?.let {
           ChatModel.fromDisplayNameNullable(it.toString())
         }
     val restoreParams =
         ChatRestoreParams(
             //        TODO: Change in the agent handling chat restore with null model
```
mkondratek, line 70: Can we remove this Todo now?

pkukielka: No. We need to pass `null` instead of `DEFAULT_MODEL.agentName` here.

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
Sa1to: Not yet

PR #678, opened 2024-02-15
Fix persisting context files

### LINKED ISSUE #636: JetBrains: Code context files are not preserved when re-launching the editor and opening previous chat from "Chat History".
### Installation Information

**Version:**
5.2.23857-nightly and 5.2.23711-nightly

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. User is already logged in to Cody with PRO subscription
3. Open any code file, select code snippet and go to Cody "Commands" and select "Explain Code"
4. Verify that At the top of the chat transcript, the arrow dropdown to display the code that Cody used as context.
5. Close the editor and relaunch it.
6. Go to Cody "Chat History"
7. Open the recent chat from step-3

**Observed behavior:**
Chat context files are missing from the transcript/conversation history.

### Expected behavior

It should keep showing the chat context used to generate that response.

**Video link:**

https://drive.google.com/file/d/1eY59lT1mi8YTfTT_Ou_ntjtCTYtbmcQD/view?usp=drive_link

Sa1to:
It fixes https://github.com/sourcegraph/jetbrains/issues/636
## Test plan

1. Send message to the chat asking about context of the repo
2. Change account and go back to the initial one
3. Open initial chat from the history
4. Check if you see context files in the chat

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
@@ -6,10 +6,7 @@ import com.jetbrains.rd.framework.base.deepClonePolymorphic
 import com.sourcegraph.cody.agent.protocol.ChatMessage
 import com.sourcegraph.cody.agent.protocol.Speaker
 import com.sourcegraph.cody.config.CodyAuthenticationManager
-import com.sourcegraph.cody.history.state.ChatState
-import com.sourcegraph.cody.history.state.EnhancedContextState
-import com.sourcegraph.cody.history.state.HistoryState
-import com.sourcegraph.cody.history.state.MessageState
+import com.sourcegraph.cody.history.state.*
 import java.time.LocalDateTime
 
 @State(name = "ChatHistory", storages = [Storage("cody_history.xml")])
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -93,6 +90,16 @@ class HistoryService(private val project: Project) :
                 Speaker.HUMAN -> MessageState.SpeakerState.HUMAN
                 Speaker.ASSISTANT -> MessageState.SpeakerState.ASSISTANT
               }
+          val contextFiles = chatMessage.contextFiles
+          if (!contextFiles.isNullOrEmpty()) {
+            contextFiles.forEach { contextFile ->
+              val element = ContextFile()
+              element.uri = contextFile.uri.toString()
+              if (!it.contextFiles.contains(element)) {
```
mkondratek, line 98: In general the approach is good. However I think we should be more precise about the saved files. 

1. Probably we should distinguish the local files and the files from the remote repo contexts.
2. We should consider persisting a given file only once - it this implementation the same file but different fragment (the lines, see example below). Maybe we should save only the files that are presented? 🤔 

Note that VS Code persists the context files with their proper links. 

```
{
  "uri": "https://sourcegraph.sourcegraph.com//github.com/sourcegraph/jetbrains@3761a12f72c5bcf8f777924d75f32db729c57788/-/blob/CONTRIBUTING.md?L49-56",
  "repoName": "github.com/sourcegraph/jetbrains",
  "revision": "3761a12f72c5bcf8f777924d75f32db729c57788",
  "range": {
    "start": {
      "line": 48,
      "character": 0
    },
    "end": {
      "line": 56,
      "character": 0
    }
  },
  "type": "file"
},
```

dominiccooney: Why do we have this code *at all* in JetBrains, instead of relying on the extension code's history? The extension shards chat history by account, preventing bugs like this, #1210 that I noticed yesterday, etc.

PR #748, opened 2024-02-20
Change model on dropdown change

mkondratek:
## Test plan
1. runIde
2. open chat
3. choose different models, write a message, verify persistence

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/LLMDropdown.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/LLMDropdown.kt
@@ -6,6 +6,8 @@ import com.intellij.openapi.ui.ComboBox
 import com.intellij.ui.MutableCollectionComboBoxModel
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.sourcegraph.cody.agent.CodyAgentService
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/ui/LLMDropdown.kt
```diff
@@ -95,4 +99,24 @@ class LLMDropdown(val project: Project, private val modelFromState: ChatModel?)
       toolTipText = CodyBundle.getString("LLMDropdown.disabled.text")
     }
   }
+
+  private fun setLlmForAgentSession(modelName: String) {
+    val sessionIdGetNow = sessionId.getNow(null) ?: return
```
pkukielka, line 104: I assume it's possible to change model only after `sessionId` was already obtained, until then sector is disabled, right?
Otherwise it could lead to a problems.

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/chat/ui/LLMDropdown.kt
mkondratek: yes, that's inconvenient but correct

PR #749, opened 2024-02-20
Do not call agent to obtain repo URL

pkukielka:
## Changes

`getRemoteRepoUrl` should be local function, not depending on agent.
It's user in a few critical places, also around a IDE startup time, so relaying on agent being ready there is potentially risky.

## Test plan

Added unit-tests + manual verification of the agent startup.

### DIFF
--- a/src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
+++ b/src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
@@ -1,55 +1,44 @@
 package com.sourcegraph.vcs
 
-import java.net.URL
-import java.util.regex.Pattern
+import java.net.URI
 
 fun convertGitCloneURLToCodebaseNameOrError(cloneURL: String): String {
 
   // Handle common Git SSH URL format
-  val sshUrlRegexMatchResult =
-      Regex("""^[\w-]+@([^:]+):([\w-]+)/([\w-]+)(\.git)?$""").find(cloneURL)
+  val sshUrlRegexMatchResult = Regex("""^[\w-]+@([^:]+):([\w-]+)/([\w-.]+)$""").find(cloneURL)
   if (sshUrlRegexMatchResult != null) {
     val (host, owner, repo) = sshUrlRegexMatchResult.destructured
-    return "$host/$owner/$repo"
+    return "$host/$owner/${repo.replace(".git$".toRegex(), "")}"
   }
-  val url = addSchemaIfNeededAndConvertURL(cloneURL) ?: return ""
+
+  val uri = URI(cloneURL)
 
   // Handle Azure DevOps URLs
-  if (url.host?.contains("dev.azure") == true && url.path.isNotEmpty()) {
-    return "${url.host}${url.path.replace("/_git", "")}"
+  if (uri.host?.contains("dev.azure") == true && !uri.path.isNullOrEmpty()) {
+    return "${uri.host}${uri.path.replace("/_git", "")}"
   }
 
   // Handle GitHub URLs
-  if (url.protocol?.startsWith("github") == true || url.toString().startsWith("github")) {
-    return "github.com/${url.path.replace(".git", "")}"
+  if (uri.scheme?.startsWith("github") == true) {
+    return "github.com/${uri.schemeSpecificPart.replace(".git", "")}"
   }
 
   // Handle GitLab URLs
-  if (url.protocol?.startsWith("gitlab") == true || url.toString().startsWith("gitlab")) {
-    return "gitlab.com/${url.path.replace(".git", "")}"
+  if (uri.scheme?.startsWith("gitlab") == true) {
+    return "gitlab.com/${uri.schemeSpecificPart.replace(".git", "")}"
   }
 
   // Handle HTTPS URLs
-  if (url.protocol?.startsWith("http") == true && url.host != null && url.path.isNotEmpty()) {
-    return "${url.host}${url.path.replace(".git", "")}"
+  if (uri.scheme?.startsWith("http") == true &&
+      !uri.host.isNullOrEmpty() &&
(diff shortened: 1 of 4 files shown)


src/test/kotlin/com/sourcegraph/vcs/ConvertUtilTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  fun `test conversion HTTP UR`() {
+    assertEquals(
+        "github.com/sourcegraph/sourcegraph",
+        convertGitCloneURLToCodebaseNameOrError("http://github.com/sourcegraph/sourcegraph"))
+  }
+
+  //  fun `test if returns null for invalid UR`() {
+  //    assertThrows(Exception::class.java) { convertGitCloneURLToCodebaseNameOrError("invalid") }
+  //  }
```
exigow: Commented-out code.

src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
```diff
@@ -1,55 +1,44 @@
...
-  val urlRegex =
-      Pattern.compile("^(https?://)?([^/?:]+)(:(\\d+))?((/[^/?#]+)*)?/?", Pattern.CASE_INSENSITIVE)
-
-  val matcher = urlRegex.matcher(cloneURL)
-  if (!matcher.matches()) return null
-  val extractedSchema = matcher.group(1)
-  val url = if (extractedSchema.isNullOrEmpty()) URL("http://$cloneURL") else URL(cloneURL)
-  return url
+  throw Exception("Cody could not extract repo name from clone URL $cloneURL")
 }
```
exigow, line 44: Maybe NIT, but it is possible to simplify this ugly if-return-...-if-return chain and use `when` similarly to:
```kotlin
fun convertGitCloneURLToCodebaseNameOrError(cloneURL: String): String =
  when {
    isGithub(cloneURL) -> "github.com/${uri.schemeSpecificPart.replace(".git", "")}"
    isHttps(cloneURL) -> "${uri.host}${uri.path?.replace(".git", "")}"
    ...
    else -> error("Cody could not extract repo name from clone URL $cloneURL")
  }
```

Also, the presence of comments like "Handle GitLab URLs" is IMO a sufficient argument to create such methods, like `isGithub`/`isHttps`/`isGeneric`/etc.

src/test/kotlin/com/sourcegraph/vcs/ConvertUtilTest.kt
exigow: BTW thanks for adding these tests! :+1:

exigow (approved):
Looks good :+1:

Added one minor comment.

src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
pkukielka: This method is conversion of VS Code method from there:
https://github.com/sourcegraph/cody/blob/main/lib/shared/src/utils.ts
I would rather keep them as similar as possible, so if someone add/change anything in the VS Code one and we will need to upgrade it will be easier to spot the diff.

exigow: Sure! I agree.

src/test/kotlin/com/sourcegraph/vcs/ConvertUtilTest.kt
exigow: `assertThrows(Exception::class.java) { convertGitCloneURLToCodebaseNameOrError("invalid") }` is failing.

src/test/kotlin/com/sourcegraph/vcs/ConvertUtilTest.kt
```diff
@@ -88,9 +89,9 @@ class ConvertUtilTest : TestCase() {
         convertGitCloneURLToCodebaseNameOrError("http://github.com/sourcegraph/sourcegraph"))
   }
 
-  //  fun `test if returns null for invalid UR`() {
-  //    assertThrows(Exception::class.java) { convertGitCloneURLToCodebaseNameOrError("invalid") }
-  //  }
+  fun `test if returns null for invalid UR`() {
```
mkondratek: nit: `URL` not `UR`

src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
```diff
@@ -1,55 +1,44 @@
 package com.sourcegraph.vcs
 
-import java.net.URL
-import java.util.regex.Pattern
+import java.net.URI
 
 fun convertGitCloneURLToCodebaseNameOrError(cloneURL: String): String {
```
mkondratek, line 5: Note: does `codebaseName` deserve its own type? It's `String` now, it's easy to confuse it with the clone url string.

mkondratek (approved):
+ we can remove `git/codebaseName` request from `CodyAgentServer`. 


pls see my comment above 🙏 it's an improvement anyway so lgtm

src/test/kotlin/com/sourcegraph/vcs/ConvertUtilTest.kt
pkukielka: Fixed.

src/test/kotlin/com/sourcegraph/vcs/ConvertUtilTest.kt
pkukielka: Fixed.

src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
pkukielka: I'm not doing this in this PR but generally I think it's good idea to change that when we will touch this code again.

PR #753, opened 2024-02-20
Simplify llm handling & keep the selected model as the default 

### LINKED ISSUE #388: PLG pro users can set one of the supported LLMs as their default for chat, so that new chats always open with their preferred LLM pre-selected

### LINKED ISSUE #691: JetBrains: There is a typo in LLM name for enterprise vs Pro user.
### Installation Information

Plugin version: `5.3.27`
IDE version: `IC-233.14015.106`

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. Login to Cody with Pro subscription
3. Remove the account
4. Login to Cody with Enterprise subscription

**Observed behavior:**
There is a typo in LLM name "Cloude 2 by Anthropic" is shown for Enterprise and "Cloude 2.0 by Anthropic" is shown for Pro 

### Expected behavior

The typo should be fixed for LLM name.

**Screenshots:**

Screenshot 2024-02-16 at 10 36 21 PM
Screenshot 2024-02-16 at 10 36 50 PM

mkondratek:
Merging queue 😅 
image

Fixes https://github.com/sourcegraph/jetbrains/issues/691.
Closes https://github.com/sourcegraph/jetbrains/issues/388.

## Test plan

👉  default model persistence:
1. Having Pro account 
2. switch model
3. open a new chat
4. selected model persists 

👉  other:
1. Model dropdown from TESTING.md


src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -297,28 +300,39 @@ private constructor(
...
-              MessageState.SpeakerState.HUMAN -> Speaker.HUMAN
-              MessageState.SpeakerState.ASSISTANT -> Speaker.ASSISTANT
-              else -> error("unrecognized speaker $speaker")
-            }
-        val chatMessage = ChatMessage(speaker = parsed, message.text)
-        chatSession.messages.add(chatMessage)
-        chatSession.chatPanel.addOrUpdateMessage(
-            chatMessage, index, shouldAddBlinkingCursor = false)
-      }
+      val agentChatSession = CompletableFuture<AgentChatSession>()
```
mkondratek, line 296: this is the worst part 😕 I am open to suggestions

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -58,11 +62,11 @@ private constructor(
             }
 
     val selectedItem =
-        chatPanel.llmDropdown.selectedItem?.let { ChatModel.fromDisplayNameNullable(it.toString()) }
+        chatPanel.llmDropdown.selectedItem?.let { (it as? ChatModelsResponse.ChatModelProvider) }
     val restoreParams =
         ChatRestoreParams(
             // TODO: Change in the agent handling chat restore with null model
-            model?.agentName ?: selectedItem?.agentName ?: DEFAULT_MODEL.agentName,
+            chatModelProvider?.model ?: selectedItem?.model ?: "", // fixme: empty string
```
pkukielka: fixme?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -297,28 +300,39 @@ private constructor(
...
-              else -> error("unrecognized speaker $speaker")
-            }
-        val chatMessage = ChatMessage(speaker = parsed, message.text)
-        chatSession.messages.add(chatMessage)
-        chatSession.chatPanel.addOrUpdateMessage(
-            chatMessage, index, shouldAddBlinkingCursor = false)
-      }
+      val agentChatSession = CompletableFuture<AgentChatSession>()
+      createNewPanel(project) { codyAgent ->
+        val innerSessionId = codyAgent.server.chatNew()
```
pkukielka, line 305: Maybe just inline that calls?
Now you have problem with tmp names which are kind of duplicated or artificial like `response`:

```
        val mySessionId = codyAgent.server.chatNew().completeOnTimeout(null, 4, TimeUnit.SECONDS).get()
         val chatModels = codyAgent.server.chatModels(ChatModelsParams(mySessionId)).completeOnTimeout(null, 4, TimeUnit.SECONDS).get()
      ```

File: src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt [hunk 4]
```diff
@@ -297,28 +300,39 @@ private constructor(
...
+        val mySessionId = innerSessionId.completeOnTimeout(null, 4, TimeUnit.SECONDS).get()
+        val chatModels = codyAgent.server.chatModels(ChatModelsParams(mySessionId))
+        val response = chatModels.completeOnTimeout(null, 4, TimeUnit.SECONDS).get()
+        val modelFromState =
+            response.models.singleOrNull { provider -> provider.model == state.model }
```
pkukielka: what is in the `state` on the very first start of the Cody (new user)?

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -297,28 +300,39 @@ private constructor(
...
+          val chatMessage = ChatMessage(speaker = parsed, message.text)
+          chatSession.messages.add(chatMessage)
+          chatSession.chatPanel.addOrUpdateMessage(
+              chatMessage, index, shouldAddBlinkingCursor = false)
+        }
 
-      CodyAgentService.withAgentRestartIfNeeded(project) { agent ->
-        chatSession.restoreAgentSession(agent, modelFromState)
-      }
+        CodyAgentService.withAgentRestartIfNeeded(project) { agent ->
```
pkukielka: You already have `codyAgent` available here, you don't need to use `CodyAgentService.withAgentRestartIfNeeded(project)` again.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -297,28 +300,39 @@ private constructor(
...
+            AgentChatSession(project, innerSessionId, state.internalId!!, modelFromState)
+        state.messages.forEachIndexed { index, message ->
+          val parsed =
+              when (val speaker = message.speaker) {
+                MessageState.SpeakerState.HUMAN -> Speaker.HUMAN
+                MessageState.SpeakerState.ASSISTANT -> Speaker.ASSISTANT
+                else -> error("unrecognized speaker $speaker")
+              }
+          val chatMessage = ChatMessage(speaker = parsed, message.text)
+          chatSession.messages.add(chatMessage)
```
pkukielka: That wa previously called on the EDT thread and now is on the background without `invokeLater`

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: I guess there is no such option. This is `createFromState` method.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: good point!

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: let's just pass `null` here. it's not worse but the code is simpler

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: changed. from this class pov it's like we are setting all messages at once now

build.gradle.kts
```diff
@@ -23,7 +23,7 @@ plugins {
   id("java")
   // Dependencies are locked at this version to work with JDK 11 on CI.
   id("org.jetbrains.kotlin.jvm") version "1.9.22"
-  id("org.jetbrains.intellij") version "1.17.0"
+  id("org.jetbrains.intellij") version "1.17.2"
```
mkondratek, line 26: 🤫

src/main/kotlin/com/sourcegraph/cody/ui/LLMComboBoxRenderer.kt
```diff
@@ -17,32 +18,34 @@ class LLMComboBoxRenderer(private val llmDropdown: LLMDropdown) : DefaultListCel
...
-    if (llmComboBoxItem.codyProOnly && isCurrentUserFree) {
+    if (chatModelProvider.codyProOnly && isCurrentUserFree) {
       textBadgePanel.add(JLabel(Icons.LLM.ProSticker), BorderLayout.EAST)
     }
 
     if (llmDropdown.isEnabled) {
       textBadgePanel.background = component.background
       textBadgePanel.foreground = component.foreground
     }
+    listOf(displayNameLabel, textBadgePanel, panel).forEach { it.isEnabled = llmDropdown.isEnabled }
```
mkondratek, line 48: this fixes the disabled component rendering in darcula theme

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -94,6 +113,24 @@ class ChatPanel(project: Project, chatSession: ChatSession, modelFromState: Chat
...
+            WebviewReceiveMessageParams(sessionId, WebviewMessage(command = "chatModel")))
+      } else {
+        agent.server.webviewReceiveMessage(
+            WebviewReceiveMessageParams(
+                sessionId, WebviewMessage(command = "chatModel", model = chatModelProvider.model)))
+      }
+    }
+
+    HistoryService.getInstance(project)
+        .updateChatLlmProvider(chatSession.getInternalId(), chatModelProvider)
```
mkondratek, line 134: instead of updating it every time the messages change let's update it on selected item change only

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSessionService.kt
```diff
@@ -34,10 +33,6 @@ class AgentChatSessionService(private val project: Project) {
   fun getSession(sessionId: SessionId): AgentChatSession? =
       chatSessions.find { it.hasSessionId(sessionId) }
 
-  fun restoreAllSessions(agent: CodyAgent) {
```
mkondratek, line 37: unused

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: looking at it now, it's not that bad

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -105,6 +101,8 @@ private constructor(
             displayText = "",
         )
     addMessageAtIndex(responsePlaceholder, index = messages.count(), shouldAddBlinkingCursor = true)
+
+    submitMessageToAgent(humanMessage, contextFiles)
```
mkondratek, line 105: I just moved it from line 94 - it makes more sense to call it after we add the placeholder

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -57,12 +61,10 @@ private constructor(
               if (acc.lastOrNull()?.speaker == msg.speaker) acc else acc.plus(msg)
             }
 
-    val selectedItem =
-        chatPanel.llmDropdown.selectedItem?.let { ChatModel.fromDisplayNameNullable(it.toString()) }
     val restoreParams =
         ChatRestoreParams(
             // TODO: Change in the agent handling chat restore with null model
-            model?.agentName ?: selectedItem?.agentName ?: DEFAULT_MODEL.agentName,
+            chatModelProvider?.model,
```
pkukielka, line 67: 🙇

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -297,28 +293,48 @@ private constructor(
...
+      createNewPanel(project) { codyAgent ->
+        val innerSessionId = codyAgent.server.chatNew()
+        val modelFromState =
+            state.llm?.let {
+              ChatModelsResponse.ChatModelProvider(
+                  default = false,
+                  codyProOnly = false,
+                  provider = it.provider ?: "Unknown",
+                  title = it.title ?: "Unknown",
+                  model = it.model ?: "")
```
pkukielka, line 306: I guess it's just a safeguard as we expect ot always get it, right?
Just to be extra safe we could also do:
model = it.model ?: it.provider ?: "Default"

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -297,28 +293,48 @@ private constructor(
...
-              MessageState.SpeakerState.HUMAN -> Speaker.HUMAN
-              MessageState.SpeakerState.ASSISTANT -> Speaker.ASSISTANT
-              else -> error("unrecognized speaker $speaker")
+      val agentChatSession = CompletableFuture<AgentChatSession>()
+      createNewPanel(project) { codyAgent ->
+        val innerSessionId = codyAgent.server.chatNew()
+        val modelFromState =
+            state.llm?.let {
+              ChatModelsResponse.ChatModelProvider(
+                  default = false,
```
pkukielka: You later check if:
`if (modelFromState?.model.isNullOrEmpty()) {` 

Maybe we can make it more clear here what is allowed state:

```kotlin
val modelFromState = 
  if (state.llm?.model == null ||
     state.llm?.it.title == null ||
     state.llm?.it.provider == null) {
    null
  } else {
    ChatModelsResponse.ChatModelProvider(
      default = false, codyProOnly = false, 
        state.llm.provider, state.llm.title, state.llm.model)
  }
```

And then this check later would be just `if (modelFromState != null)`
My point there is that "Unknown" states are invalid anyway.

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -83,6 +85,33 @@ class SettingsMigration : Activity {
...
   }
 
+  private fun migrateLlms(project: Project) {
+    HistoryService.getInstance(project)
+        .state
+        .chats
+        .filter { it.llm == null }
+        .forEach {
+          val (provider, title) =
+              modelToProviderAndTitle.getOrDefault(it.model, Pair("Unknown", "Unknown"))
```
pkukielka: I assume this is mostly used of displaying to user, right?
Then maybe instead of `Pair("Unknown", "Unknown")`:
```kotlin
val nameParts = it.model.split("/")
val provider = nameParts.firstOrNull() ?: "Unknown"
val model = nameParts.lastOrNull() ?: "Unknown"
Pair(provider, model)
```

It will be much more user friendly if some model will be unexpectedly added before we catch up in the code.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: it will not work 😢 

```
e: file:///Users/mkondratek/IdeaProjects/jetbrains/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt:317:27 Smart cast to 'LLMState' is impossible, because 'state.llm' is a mutable property that could have been changed by this time
```

mkondratek: the question is - what should we do if we cannot restore the model? 

I agree approach allows us to restore it anyway while we may not be able to determine the model. A user still can use it. Ideally, we could make such a chat READ-ONLY.

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
mkondratek: I thought we were storing the models as `modelID` but that's not true. it was `displayName` 😕 I must map display names - pls retake a look here

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -29,6 +30,7 @@ class SettingsMigration : Activity {
       migrateProjectSettings(project)
       migrateAccounts(project)
     }
+    RunOnceUtil.runOnceForProject(project, "CodyHistoryLlmMigration") { migrateLlms(project) }
```
exigow, line 33: Why this migration is in-between other migrations? I think logically this migration should be performed last.

mkondratek: It's the last of other `runOnceForProject`s. The later entries are `runOnceForApp`.

In general, I agree - we should append the migration e.g.:
```
1. A->B
2. B->C
```
The other way it would not work (assuming that dependency exists).

However `runOnceForProject` is not the same as `runOnceForApp` - those are different properties handled by different xml files. IMHO we are OK in this case.

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -209,7 +207,7 @@ private constructor(
         }
       }
       else -> {
-        logger.debug(String.format("CodyToolWindowContent: unknown message type: %s", message.type))
+        logger.debug(String.format("AgentChatSession: unknown message type: %s", message.type))
```
pkukielka: That in fact should not be needed as logger already log the class

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
exigow: Sure! Thanks for explanation.

exigow (approved):

PR #755, opened 2024-02-20
Do not create tmp binary file on every agent startup

pkukielka:
## Changes

Currently agent binary is always copied to tmp dir on every startup.
I do not see any reason to not start it directly from plugins dir.
On slower computers it will space some CPU and disk usage.
We also won't need to care about the cleanup.

## Test plan

Manual tests - start/restart agent multiple times in various conditions


mkondratek (approved):
I cannot see any contraindications

PR #761, opened 2024-02-20
Updated the protocol bindings for PR 3219

steveyegge:
The title field of ProtocolCommand is now a structured object. We are still committing protocol changes manually for now, and this matches the change in sourcegraph/cody/agent/bindings.

## Test plan

Tested manually in my branch, which is the only code that uses it.


steveyegge:
Don't want to wrestle with this merge anymore; this change can wait until my main PR is merged.

Issue #509, opened 2024-02-06
Cody for JetBrain is consuming high CPU when running Cody Commands i.e. `Generate Test`


jay-fibi:
### Installation Information

Version:
5.2.21709

### Describe the bug

Steps to reproduce:
1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. User is already logged in to Cody
3. Open any file and select a code snippet
4. Click on 'Cody' icon from right pane and go to 'Commands'
5. Click on 'Generate Test'
6. Open 'Activity Monitor' and search for IntelliJ and observe CPU usage during Cody action you have performed in step 4th and 5th

Observed behavior:
CPU usage is increased to ~50% during test generation.

### Expected behavior

CPU usage should not be spiked up very high.

### Additional context (logs, images, etc)

In the VS Code, if user perform same operation then CPU usage will not spiked up very high.

Video attachment:
https://drive.google.com/file/d/17ey32-g0WN7mX-O5jDrIp_5n2wk47xeo/view?usp=drive_link

taylorsperry:
Thanks, @jay-fibi. We're aware of this issue and are working on a fix!

exigow:
Looks like CPU usage increases when markdown appears in the chat, and it's not just about the situation where tokens are generated, but also when I simply scroll through the chat so this may be some kind of rendering issues. I'm looking for cause of these spikes.

taylorsperry:
@jay-fibi would you please confirm that you're no longer seeing this issue on `v5.2.23880-nightly`?

jay-fibi:
@taylorsperry I am still facing this issue with `v5.3.27-nightly`

exigow:
My debugging so far:

1. Baseline: CPU usage on idle IDE is low around ~few percent.
2. Running the Generate Test command (or anything that contains Markdown in response) in native resolution (2560x1440 px) causes CPU usage spike from 2-5% to 10-20%.
3. After increasing the resolution 2x (to 5120x2880 px, virtually via `xrandr --output DP-0 --mode 2560x1440 --fb 5120x2880 --rate 165 --scale 2x2 --dpi 256`) the CPU usage increases significantly from 2-10% to 20-80% :rocket::rocket::rocket: for an identical command and response. The resolution increase is intended to simulate the resolution on Macs with hi-res displays. This also proves that the problem is with rendering performance (more pixels to draw -> higher CPU usage).
[...]

PR #784, opened 2024-02-21
Remove excessive memory allocations during Chat History sorting

exigow:
Related to #509

This simple fix introduces cache for ChatState.updatedAt date. The getUpdatedTimeAt getter was causing memory allocations because each call invoked the LocalDateTime.parse method. Value updatedAt is stored internally as string. This date is used as a basis for sorting, so it's intensively used during the Chat History tree update (in multiple places).

## Test Plan

1. Start new conversation -> date in is properly initialized (today's date) and saved to XML
2. Continue old conversation -> date is properly updated, moved to top of the list, and saved to XML

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/history/state/ChatState.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/state/ChatState.kt
@@ -33,10 +33,8 @@ class ChatState : BaseState() {
   }
 
   fun getUpdatedTimeAt(): LocalDateTime {
-    if (updatedAt == null)
-      return LocalDateTime.now()
-    if (updatedAtDate == null)
-      updatedAtDate = LocalDateTime.parse(updatedAt, DATE_FORMAT)
+    if (updatedAt == null) return LocalDateTime.now()
+    if (updatedAtDate == null) updatedAtDate = LocalDateTime.parse(updatedAt, DATE_FORMAT)
     return updatedAtDate!!
   }
 
--- a/src/main/kotlin/com/sourcegraph/cody/history/state/ChatState.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/state/ChatState.kt
@@ -35,7 +35,7 @@ class ChatState : BaseState() {
   fun getUpdatedTimeAt(): LocalDateTime {
     if (updatedAt == null) return LocalDateTime.now()
     if (updatedAtDate == null) updatedAtDate = LocalDateTime.parse(updatedAt, DATE_FORMAT)
-    return updatedAtDate!!
+    return updatedAtDate ?: LocalDateTime.now()
   }
 
   companion object {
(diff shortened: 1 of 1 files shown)


exigow:
Proof w/ screenshots:

## Before

1. Memory allocations when I asked for long response (smth like `convert this X java code to kotlin`). Only HistoryTree is listening for these changes.

Selection_083

2. You can see on flamegraph that parse method is causing ~45% of all memory allocations of our code.

image

## After

(identical prompt, similar response)

image

image

The allocation is marginal (~4%, highlighted yellow bars).

src/main/kotlin/com/sourcegraph/cody/history/state/ChatState.kt
```diff
@@ -23,15 +23,19 @@ class ChatState : BaseState() {
...
 
   fun setUpdatedTimeAt(date: LocalDateTime) {
+    updatedAtDate = date
     updatedAt = date.format(DATE_FORMAT)
   }
 
   fun getUpdatedTimeAt(): LocalDateTime {
     if (updatedAt == null) return LocalDateTime.now()
-    return LocalDateTime.parse(updatedAt, DATE_FORMAT)
+    if (updatedAtDate == null) updatedAtDate = LocalDateTime.parse(updatedAt, DATE_FORMAT)
```
pkukielka, line 37:
```kotlin
  private var updatedAtDate: LocalDateTime = LocalDateTime.now()
  
  fun setUpdatedTimeAt(date: LocalDateTime) {
    updatedAt = date.format(DATE_FORMAT)
    updatedAtDate = date
  }

  fun getUpdatedTimeAt(): LocalDateTime {
      if (updatedAt == null) setUpdatedTimeAt(LocalDateTime.now())
      return updatedAtDate
  }
```

exigow: Thanks! However, the chat dates are not properly initialized and everything lands in Today after deser. I had a similar problem, and as a result, it ended up with this code with `!!`.

image

exigow: Maybe let's just replace `return updatedAtDate!!` -> `return updatedAtDate ?: LocalDateTime.now()`?

EDIT: pushed `return updatedAtDate ?: LocalDateTime.now()` and tested. Everything works.

pkukielka: Yea, I realised afterwards it initially will be incorrectly initialized and will possibly not get re-initialized.
Do you know if that `BaseState` classes are initialised by constructor or using setters when loaded for the first time?

exigow:
The issue is more visible if we have a large number of chats. Out of curiosity, I created 100 empty ones and it turns out that generating a new response (long response with code) allocates 20GB during sorting, which is several times more than the memory allocated by IDE itself.

image

pkukielka (approved):
LGTM

PR #804, opened 2024-02-22
Update tokens incrementally in Chat window

### LINKED ISSUE #886: Unfinished code blocks are parsed as paragraphs in Cody Chat
When Cody responds with an incomplete markdown code block ("```"), we may receive a message ("`" or "``") parsed as a paragraph.

image

This happens quite often.

exigow:
Related to #509
Fixes #886

This commit aims to reduce CPU usage when using the Chat window when Cody is responding.

Changes:

1. Bug #886 is fixed.
2. The chat is now updated only when the number of tokens grows. It sometimes happens that Cody responds few times with the identical text (in streaks, even 2-3 times) or responds in a different order. Such responses are now ignored, so the panel is not unnecessarily refreshed.
3. The code block visible in the chat (which is an underlying virtual editor) is now updated incrementally. Each response from Cody adds a few tokens at the end of the document instead of replace-all strategy.

## Test Plan

* Ask a question that will display a code block (e.g. "write hello world in 10 different languages").
* The chat looks clean and does not have excessive paragraphs with "``" symbols.

## Demo

Before:

Selection_094

After:

Selection_096

-----

~~WIP (it's working 80% of the time, i've added some details below)~~

~~ToDos/ToDiscuss:~~

* ~~Responses with new tokens are not always in a logical order or may be redundant (e.g., may go "backwards" or respond with identical response 2-3 times). This explains why the code is implemented this way, not otherwise, meaning a new token causes the entire UI to be overwritten. This also introduces interesting behaviour: when responses are not in logical order, tokens can disappear for a short time.~~
* ~~All the code has been written with this "feature" in mind (e.g. replace everything and don't worry about the logical order). This is root of the performance issues.~~

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
@@ -36,7 +36,7 @@ class SingleMessagePanel(
   }
 
   fun updateContentWith(text: String) {
-    val trimmedText = text.removeBlockSuffixAndTrim()
+    val trimmedText = text.trimEnd { c -> c == '`' || c.isWhitespace() }
     val isGrowing =
         trimmedText.contains(lastTrimmedText) && trimmedText.length > lastTrimmedText.length
     if (isGrowing) {
@@ -52,14 +52,6 @@ class SingleMessagePanel(
         addOrUpdateText(renderedHtml)
       }
     }
-  }
-
-  private fun String.removeBlockSuffixAndTrim(): String {
-    var lastIndex = this.length - 1
-    while (lastIndex >= 0 && (this[lastIndex] == '`' || this[lastIndex].isWhitespace())) {
-      lastIndex--
-    }
-    return this.substring(0, lastIndex + 1)
   }
 
   fun addOrUpdateCode(code: String, language: String?) {
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
@@ -28,23 +28,38 @@ class SingleMessagePanel(
     private val chatSession: ChatSession,
 ) : PanelWithGradientBorder(gradientWidth, chatMessage.speaker) {
   private var lastMessagePart: MessagePart? = null
+  private var lastTrimmedText = ""
 
   init {
     val markdownNodes: Node = markdownParser.parse(chatMessage.actualMessage())
     markdownNodes.accept(MessageContentCreatorFromMarkdownNodes(this, htmlRenderer))
   }
 
-  fun updateContentWith(message: ChatMessage) {
-    val markdownNodes = markdownParser.parse(message.actualMessage())
-    val lastMarkdownNode = markdownNodes.lastChild
-    if (lastMarkdownNode != null && lastMarkdownNode.isCodeBlock()) {
-      val (code, language) = lastMarkdownNode.extractCodeAndLanguage()
-      addOrUpdateCode(code, language)
-    } else {
-      val nodesAfterLastCodeBlock = markdownNodes.findNodeAfterLastCodeBlock()
-      val renderedHtml = htmlRenderer.render(nodesAfterLastCodeBlock)
-      addOrUpdateText(renderedHtml)
+  fun updateContentWith(text: String) {
+    val trimmedText = text.removeBlockSuffixAndTrim()
+    val isGrowing =
+        trimmedText.contains(lastTrimmedText) && trimmedText.length > lastTrimmedText.length
+    if (isGrowing) {
+      lastTrimmedText = trimmedText
+      val markdownNodes = markdownParser.parse(text)
+      val lastMarkdownNode = markdownNodes.lastChild
+      if (lastMarkdownNode != null && lastMarkdownNode.isCodeBlock()) {
+        val (code, language) = lastMarkdownNode.extractCodeAndLanguage()
+        addOrUpdateCode(code, language)
+      } else {
+        val nodesAfterLastCodeBlock = markdownNodes.findNodeAfterLastCodeBlock()
+        val renderedHtml = htmlRenderer.render(nodesAfterLastCodeBlock)
+        addOrUpdateText(renderedHtml)
+      }
     }
+  }
+
+  private fun String.removeBlockSuffixAndTrim(): String {
+    var lastIndex = this.length - 1
+    while (lastIndex >= 0 && (this[lastIndex] == '`' || this[lastIndex].isWhitespace())) {
+      lastIndex--
+    }
+    return this.substring(0, lastIndex + 1)
   }
 
   fun addOrUpdateCode(code: String, language: String?) {
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagesPanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagesPanel.kt
@@ -28,7 +28,7 @@ class MessagesPanel(private val project: Project, private val chatSession: ChatS
     if (messageToUpdate != null) {
       val singleMessagePanel = messageToUpdate.getComponent(0) as? SingleMessagePanel
       val contextFilesPanel = messageToUpdate.getComponent(1) as? ContextFilesPanel
-      singleMessagePanel?.updateContentWith(message)
+      singleMessagePanel?.updateContentWith(message.actualMessage())
       contextFilesPanel?.updateContentWith(message.contextFiles)
     } else {
       addChatMessageAsComponent(message)


src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
```diff
@@ -59,9 +61,21 @@ class CodeEditorPart(
...
   }
 
-  private fun updateText(project: Project, text: String) {
-    this.text = text
-    WriteCommandAction.runWriteCommandAction(
-        project, Computable { editor.document.replaceText(text, System.currentTimeMillis()) })
+  private fun updateText(project: Project, newText: String) {
+    val newIndex = newText.length - 1
+    if (newIndex >= lastIndex) {
+      val size = newIndex - lastIndex
```
pkukielka: I'm not sure if that is correct.
I think - but I'm not 100% sure - I have seen LLM responses when it was retracting few last tokens and replacing them with something else. So to be really sure we should also compare content, not just the length.

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
```diff
@@ -29,6 +29,8 @@ class CodeEditorPart(
     val attribution: AttributionButtonController
 ) : MessagePart {
 
+  private var lastIndex = 0
```
pkukielka: Do we really need that tmp variable? Cannot we use `editor.document.textLength` instead?

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
```diff
@@ -59,9 +61,21 @@ class CodeEditorPart(
...
+        val toAdd = newText.substring(lastIndex, newIndex)
+        WriteCommandAction.runWriteCommandAction(
+            project, Computable { editor.document.insertString(lastIndex, toAdd) })
+        lastIndex = newIndex
+      } else {
+        // todo sometimes difference to previous response may be 0 (responses have no value / don't
+        // need to refresh)
+      }
+    } else {
+      // todo this may happen (cody responses are not always ordered)
```
pkukielka: that sounds a bit strange, let's discuss offline.

exigow: Discussed. There is a different implementation now.

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
exigow: This code has been changed. Please take a look.

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
```diff
@@ -59,9 +60,15 @@ class CodeEditorPart(
...
+    val newIndex = newText.length - 1
+    if (newIndex >= lastIndex) {
+      val toAdd = text.substring(lastIndex, newIndex)
+      WriteCommandAction.runWriteCommandAction(
+          project, Computable { editor.document.insertString(lastIndex, toAdd) })
+      lastIndex = newIndex
+    }
   }
+
 }
```
exigow: Bad news: After checking how it looks in the profiler, there is no/minimal gain between incremental text addition vs previous, simple replace-all (at the document level). These changes are local, so we can discard them (?). Unfortunately, there was no other option than to simply implement & measure.

src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
```diff
@@ -26,22 +26,30 @@ class SingleMessagePanel(
...
+      val lastMarkdownNode = markdownNodes.lastChild
+      if (lastMarkdownNode != null && lastMarkdownNode.isCodeBlock()) {
+        val (code, language) = lastMarkdownNode.extractCodeAndLanguage()
+        val fixed  = code.removeIncompleteSuffix()
+        addOrUpdateCode(fixed, language)
+      } else {
+        val nodesAfterLastCodeBlock = markdownNodes.findNodeAfterLastCodeBlock()
+        val renderedHtml = htmlRenderer.render(nodesAfterLastCodeBlock)
+        addOrUpdateText(renderedHtml)
+      }
```
exigow, line 53: FYI: Here is the biggest gain (discarding redundant messages; according to profiler ~10-20% less cpu time for big responses).

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
pkukielka: My question remains valid.
I do not see why we cannot use `editor.document.textLength` and `editor.document.text` instead of tmp variables `lastIndex` and `text`.

src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
```diff
@@ -117,5 +125,13 @@ class SingleMessagePanel(
     private val markdownParser = Parser.builder().extensions(extensions).build()
     private val htmlRenderer =
         HtmlRenderer.builder().softbreak("<br />").extensions(extensions).build()
+
+    private fun String.removeIncompleteSuffix(): String = trimEnd().removeBlockSuffix()
+
+    private fun String.removeBlockSuffix(): String {
+      var lastIndex = this.length - 1
```
pkukielka: Why not just:
```this.trimEnd { c -> c == '`' || c.isWhitespace() }```
:)

BTW. Your code would fail for this string: ```"some string ` "```

**Review by pkukielka — CHANGES_REQUESTED** [review]
I think this implementation cannot work properly, let's discuss offline.

File: src/main/kotlin/com/sourcegraph/cody/history/state/ChatState.kt [hunk 7]
```diff
@@ -23,15 +23,19 @@ class ChatState : BaseState() {
...
 
   fun setUpdatedTimeAt(date: LocalDateTime) {
+    updatedAtDate = date
     updatedAt = date.format(DATE_FORMAT)
   }
 
   fun getUpdatedTimeAt(): LocalDateTime {
     if (updatedAt == null) return LocalDateTime.now()
-    return LocalDateTime.parse(updatedAt, DATE_FORMAT)
+    if (updatedAtDate == null) updatedAtDate = LocalDateTime.parse(updatedAt, DATE_FORMAT)
```
mkondratek: 👍

src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
```diff
@@ -28,23 +28,38 @@ class SingleMessagePanel(
...
-    val markdownNodes = markdownParser.parse(message.actualMessage())
-    val lastMarkdownNode = markdownNodes.lastChild
-    if (lastMarkdownNode != null && lastMarkdownNode.isCodeBlock()) {
-      val (code, language) = lastMarkdownNode.extractCodeAndLanguage()
-      addOrUpdateCode(code, language)
-    } else {
-      val nodesAfterLastCodeBlock = markdownNodes.findNodeAfterLastCodeBlock()
-      val renderedHtml = htmlRenderer.render(nodesAfterLastCodeBlock)
-      addOrUpdateText(renderedHtml)
+  fun updateContentWith(text: String) {
```
mkondratek, line 38: nit: It would be nice to have a comment over this method. The logic seems to be more complex than needed (I know it's needed today but tomorrow... 😅 )

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
exigow: Updated. Please take a look.

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
exigow: I've removed this code.

src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
exigow: Updated as suggested! Please take a look.

src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagePart.kt
exigow: Code removed.

src/main/kotlin/com/sourcegraph/cody/history/state/ChatState.kt
exigow: :D

I added it by mistake. Fixed.

PR #854, opened 2024-02-26
Remove unused CodyProJetBrains flag

pkukielka:
## Changes

Remove unused CodyProJetBrains flag which sometimes were causing problems on the EDT thread.

## Test plan

Manual tests according to QA testing guide, rate limit testing section

### DIFF
--- a/src/main/java/com/sourcegraph/cody/MyAccountTabUtils.kt
+++ b/src/main/java/com/sourcegraph/cody/MyAccountTabUtils.kt
@@ -7,11 +7,7 @@ import com.sourcegraph.cody.agent.CodyAgentServer
 import com.sourcegraph.cody.config.CodyAuthenticationManager
 import java.util.concurrent.CompletableFuture
 
-data class MyAccountTabPanelData(
-    val isDotcomAccount: Boolean,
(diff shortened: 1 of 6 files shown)


src/main/java/com/sourcegraph/cody/MyAccountTabUtils.kt
```diff
@@ -28,25 +26,12 @@ fun fetchMyAccountPanelData(
...
-              MyAccountTabPanelData(
-                  activeAccountType.isDotcomAccount(),
-                  codyProFeatureFlag = false,
-                  isCurrentUserPro = null))
-        }
+        val isCurrentUserPro = getIsCurrentUserPro(server) ?: false
+        result.complete(
+            MyAccountTabPanelData(
+                activeAccountType.isDotcomAccount(),
+                codyProFeatureFlag = true,
```
mkondratek: I guess we can remove it here too - it should be always true now.

pkukielka: Good point, fixed.

mkondratek (approved):

PR #897, opened 2024-02-28
Fix InvalidPathException during URI conversion

### LINKED ISSUE #609: bug: InvalidPathException
Users have reported an error around `InvalidPathException`. @pkukielka noted that because these reports are coming from just 3 users, it could be they are using network drives or containers for their workspaces, making their environments atypical (thus the `P1` label). This issue collects them in one spot:

https://github.com/sourcegraph/jetbrains/issues/342
https://github.com/sourcegraph/jetbrains/issues/355
https://github.com/sourcegraph/jetbrains/issues/466
https://github.com/sourcegraph/jetbrains/issues/519
https://github.com/sourcegraph/jetbrains/issues/525
https://github.com/sourcegraph/jetbrains/issues/527
https://github.com/sourcegraph/jetbrains/issues/528
https://github.com/sourcegraph/jetbrains/issues/529
https://github.com/sourcegraph/jetbrains/issues/530
https://github.com/sourcegraph/jetbrains/issues/531

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/609

## Changes

I replaced manual path conversion of `file: VirtualFile` from:

```kotlin
Paths.get(file .path).toUri().toString()
```

to:

```kotlin
file.url
```

During manual tests it seems to be giving proper results, and I would argue that more correct ones than before.
E.g. if accessing file inside a jar protocol is now `jar` instead of `file`.

## Test plan

Unfortunately I do not have a local reproduction, but I completely eliminate the code path which was problematic

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
@@ -14,7 +14,7 @@ private constructor(
 
     @JvmStatic
     @JvmOverloads
-    fun fromPath(
+    fun fromVirtualFile(
         file: VirtualFile,
         content: String? = null,
         selection: Range? = null
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
@@ -1,7 +1,7 @@
 package com.sourcegraph.cody.agent.protocol
 
+import com.intellij.openapi.vfs.VirtualFile
 import com.sourcegraph.cody.agent.protocol.util.Rfc3986UriEncoder
-import java.nio.file.Paths
 
 class TextDocument
 private constructor(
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
```diff
@@ -14,9 +14,12 @@ private constructor(
 
     @JvmStatic
     @JvmOverloads
-    fun fromPath(path: String, content: String? = null, selection: Range? = null): TextDocument {
-      val uri = Paths.get(path).toUri().toString()
-      val rfc3986Uri = Rfc3986UriEncoder.encode(uri)
+    fun fromPath(
```
mkondratek: nit: `s/fromPath/fromFile/`

src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
```diff
@@ -14,9 +14,12 @@ private constructor(
...
+        file: VirtualFile,
+        content: String? = null,
+        selection: Range? = null
+    ): TextDocument {
+      val rfc3986Uri = Rfc3986UriEncoder.encode(file.url)
```
mkondratek, line 22: to be verified on windows

src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
pkukielka: Done

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/agent/protocol/TextDocument.kt
pkukielka: I checked on windows, seems to be working just fine.

PR #939, opened 2024-03-01
Hotfix: Fix URL in error submitter

exigow:
I didn't notice that the URL points to a different, test repository. It should be okay now.


pkukielka (approved):
LGTM

Issue #762, opened 2024-02-20
JetBrains: An error occurred "java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException:" after deleting token and changing project


jay-fibi:
### Installation Information

Plugin version: ```5.3.27```
IDE version: ```PC-233.13763.11```

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. User is already logged in to Cody with Pro subscription
3. Delete the auth token from Sourcegraph web
4. Go to editor and run any command i.e. Generate Test
5. Switch to another project and run any command

**Observed behavior:**
An error occurred on performing this activity quickly and below stacktrace thrown,

```text
java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: accessing Sourcegraph GraphQL API: Error: HTTP status code 401: Invalid access token.
 (https://sourcegraph.com/.api/graphql?CurrentUserCodySubscription)

Error: accessing Sourcegraph GraphQL API: Error: HTTP status code 401: Invalid access token.
 (https://sourcegraph.com/.api/graphql?CurrentUserCodySubscription)
    at /snapshot/dist/agent.js
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /snapshot/dist/agent.js
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
	at com.sourcegraph.cody.initialization.EndOfTrialNotificationScheduler.lambda$1$lambda$0(EndOfTrialNotificationScheduler.kt:44)
	at com.sourcegraph.cody.agent.CodyAgentService$Companion.withAgentRestartIfNeeded$lambda$1(CodyAgentService.kt:134)
	at com.intellij.openapi.application.impl.ApplicationImpl$2.run(ApplicationImpl.java:249)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
[...]

Issue #289, opened 2024-01-12
Invalid access token reported


simonintech:
### Installation Information

IntelliJ IDEA 2023.3.1 (Community Edition)
Build #IC-233.11799.300, built on December 12, 2023
Runtime version: 17.0.9+7-b1087.7 x86_64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.2.1
GC: G1 Young Generation, G1 Old Generation
Memory: 4096M
Cores: 16
Metal Rendering is ON
Registry:
  ide.experimental.ui=true
Non-Bundled Plugins:
  com.sourcegraph.jetbrains (5.2.17749)
Kotlin: 233.11799.300-IJ

<img width="444" alt="image" src="https://github.com/sourcegraph/jetbrains/assets/23657075/5249ca9b-d15e-4217-8fca-78d9f9514705">

<img width="394" alt="image" src="https://github.com/sourcegraph/jetbrains/assets/23657075/1ab9ce51-8ff6-4319-9212-8a36d053f573">

### Describe the bug

Installed latest stable build received 'Cody has encountered an error' when using chat. IDEA log reported invalid token. Deleted both of my existing accounts - enterprise Sourcegraph and Sourcegraph.com. Restarted IDEA. Recreated enterprise account with new token. Encountered same issue - 'Cody has encountered an error'

### Expected behavior

Expected response from chat

### Additional context (logs, images, etc)

```
2024-01-12 10:20:25,777 [  74591]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - onChatUpdateMessageInProgress is null or message type is not transcript
2024-01-12 10:20:25,777 [  74591]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - webview/postMessage 43b0bf7a-030b-4254-a0eb-793f68e3a0dc: ExtensionMessage(type=history, messages=[ChatMessage(speaker=null, text=null, displayText=null, contextFiles=null, error=null)], isMessageInProgress=null, chatID=null, isTranscriptError=null, customPrompts=null, context=null, errors=null)
2024-01-12 10:20:25,778 [  74592]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ SimpleChatPanelProvider: getEnhancedContext > embeddings (start)
[...]

Issue #487, opened 2024-02-04
bug: java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: current user subscription data not found


vaskenkazanjian:
Plugin version: ```5.2.21709```
IDE version: ```IU-233.14015.106```
Exception: ```java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: current user subscription data not found```
Stacktrace:
```text
java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: current user subscription data not found

Error: current user subscription data not found
    at C:\snapshot\dist\agent.js
    at extractDataOrError (C:\snapshot\dist\agent.js)
    at C:\snapshot\dist\agent.js
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async C:\snapshot\dist\agent.js
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
	at com.sourcegraph.cody.initialization.EndOfTrialNotificationScheduler.lambda$1$lambda$0(EndOfTrialNotificationScheduler.kt:44)
	at com.sourcegraph.cody.agent.CodyAgentService$Companion$applyAgentOnBackgroundThread$1.invoke$lambda$0(CodyAgentService.kt:109)
	at com.intellij.openapi.application.impl.ApplicationImpl$2.run(ApplicationImpl.java:249)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:702)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:699)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1.run(Executors.java:699)
[...]

PR #956, opened 2024-03-04
Report all agent errors using the chat or status bar

pkukielka:
Fixes: Invalid access token reported
[#289](https://github.com/sourcegraph/jetbrains/issues/289)
[#762](https://github.com/sourcegraph/jetbrains/issues/762)

Now every chat or agent error is properly reported on the UI:

image

## Test plan

1. Create new access token
2. Login using that access token
3. Run some Cody actions, like chat question and autocomplete
4. Remove access token from SG instance
5. Run some Cody action and check if it properly reports an error
6. Switch account and check if error is gone
7. Switch account back and check if error is reported again


src/main/kotlin/com/sourcegraph/cody/statusbar/CodyAutocompleteStatus.kt
```diff
@@ -34,6 +35,12 @@ enum class CodyAutocompleteStatus : PresentableEnum, WithIcon {
 
     override val icon: Icon = Icons.StatusBar.CodyAutocompleteUnavailable
   },
+  AgentError {
+    override fun getPresentableText(): String =
+        "Cody encountered an error: ${CodyAgentService.agentError.get()}"
+
+    override val icon: Icon = Icons.StatusBar.CodyAutocompleteUnavailable
```
mkondratek: nit: `s/CodyAutocompleteUnavailable/CodyUnavailable`?

src/main/resources/CodyBundle.properties
```diff
@@ -43,6 +43,11 @@ chat.rate-limit-error.explain=\
     To ensure that Cody can stay operational for all Cody users, please come back tomorrow for more chats, commands, and autocompletes. \
     <a href=\"https://sourcegraph.com/docs/cody/core-concepts/cody-gateway#rate-limits-and-quotas\">Learn more.</a>\
   </html>
+chat.general-error=\
+  <html>\
+    &#9888; <b> Cody encountered an error when processing your message:</b><br>\
```
mkondratek, line 48: I wonder... maybe we should put a link to GitHub bug template here? 🤔

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -111,33 +113,53 @@ class CodyAgentService(project: Project) : Disposable {
   }
 
   companion object {
+    private val logger = Logger.getInstance(CodyAgent::class.java)
+
+    val agentError: AtomicReference<String?> = AtomicReference(null)
```
mkondratek, line 105: nit: `s/agentError/mostRecentlyReportedError/`?

mkondratek (changes requested):

src/main/resources/CodyBundle.properties
mkondratek: (users should be able to easily report cases like https://github.com/sourcegraph/jetbrains/issues/487)

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
pkukielka: All of those errors are `recently reported` as they are used in the status bar.
So really I do not think adding it only for this element (or at all) adds any information here.

src/main/resources/CodyBundle.properties
pkukielka: I'm on the edge there.
Example you gave is good one.
On the other hand things like missing token are not things we would like to be reported.

Maybe @danielmarquespt should look at this and advise, I can do follow-up PR to adjust the message.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
mkondratek: no strong feelings. my suggestion seems to better reflect the variable sense

src/main/resources/CodyBundle.properties
mkondratek: We should not leave the user with "Invalid access token." for sure 😄 
For that particular case we probably can suggest the user to re-authenticate. 
For a general case - I'd rather leave a link to bug report.

mkondratek (approved):
LGTM. let's improve the error handling in next PRs if needed

Issue #966, opened 2024-03-04
JetBrains: File will not be opened when clicking on file name from Context in response in Windows.


jay-fibi:
### Installation Information

Plugin version: `5.3.2999-nightly`
IDE version: `IC-233.14475.28`

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. User is already logged in to Cody with Free/Pro/Enterprise account
3. Open any file and go to "Commands" and select "Explain Code" and make sure that only single file should be shown under Context from response.
4. Close the opened file and click on file name from Context link

**Observed behavior:**
File will not be opened when clicked from Context in the response.

### Expected behavior

File should be opened correctly.

### Additional context

**Video link:**

https://drive.google.com/file/d/1sRxTdWh7PKrC9he3zvanURYc91n47BZ3/view?usp=drive_link

pkukielka:
Fixed in https://plugins.jetbrains.com/plugin/9682-sourcegraph-cody--code-search/versions/nightly/499547

Issue #964, opened 2024-03-04
JetBrains: Cody autocompletion is not working on Windows.


jay-fibi:
### Installation Information

Plugin version: ```5.3.2999-nightly```
IDE version: ```IC-233.14475.28```

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. User is already logged in to Cody with Free/Pro/Enterprise account
3. Open any Java file and type as below,
```
// print Hello World!
System.out. 
```
4. Place a cursor at the end of the `System.out.` line.
5. Trigger autocompletion with `Alt + /`

**Observed behavior:**
Cody autocompletion does not triggered.

### Expected behavior

Cody autocompletion should work and finish the remaining line of code as below,
```
// print Hello World!
System.out.println("Hello World!")
```

### Additional context

**Video link:**

https://drive.google.com/file/d/1T551u6MBUH1-5STDjQv7uf4l0nHX5fxB/view?usp=drive_link

pkukielka:
Fixed in https://plugins.jetbrains.com/plugin/9682-sourcegraph-cody--code-search/versions/nightly/499547

PR #981, opened 2024-03-05
Fix problems with URI conversions

pkukielka:
## Changes

Fixes two bugs:
https://github.com/sourcegraph/jetbrains/issues/966
https://github.com/sourcegraph/jetbrains/issues/964

## Test plan

Manual tests on both windows and MacOs + QA tests.

1. Check if completions works according to TESTING.MD
2. Check if context links works and navigates to proper file and line


mkondratek (approved):
lgtm

Issue #967, opened 2024-03-04
JetBrains: Unable to remove remote repo for one specific project.


jay-fibi:
### Installation Information

Plugin version: `5.3.2999-nightly`
IDE version: `IC-233.14475.28`

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. User is already logged in to Cody with Enterprise account and few remote repo are already added
3. Click on '-' button from Context window to remove remote repo.
4. Upon removal, click on "New Chat" button and verify the Context window

**Observed behavior:**
Remote repo are still there and not removed correctly.
Though, this behavior is only for this specific project and it is working fine with my other projects.

Here is the log snippet,
```
2024-03-05 02:14:43,987 [11890112]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ ChatPanelsManager:createWebviewPanel: 1
2024-03-05 02:14:43,987 [11890112]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ SimpleChatPanelProvider: postContextStatusToWebView []
2024-03-05 02:14:43,987 [11890112]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ SimpleChatPanelProvider: postContextStatusToWebView []
2024-03-05 02:14:43,987 [11890112]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ SimpleChatPanelProvider:registerWebviewPanel: registering webview panel
2024-03-05 02:14:43,988 [11890113]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ SimpleChatPanelProvider: postContextStatusToWebView []
2024-03-05 02:14:43,988 [11890113]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ SimpleChatPanelProvider: handleInitialized
2024-03-05 02:14:43,988 [11890113]   WARN - #com.sourcegraph.cody.agent.CodyAgentClient - Cody by Sourcegraph: █ SimpleChatPanelProvider: updateViewConfig  {
  "arch": "x64",
  "os": "win32",
  "extensionVersion": "1.6.0",
  "uiKindIsWeb": true,
  "debugEnable": true,
[...]

Issue #656, opened 2024-02-14
JetBrains: "Chat History" from yesterday is not available for enterprise account.


jay-fibi:
### Installation Information

Plugin version: `5.3.27-nightly`
IDE version: `IC-233.14015.106`

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains IntelliJ IDEA editor
2. User is already logged in to Cody with Enterprise subscription and have performed some Cody chat yesterday
3. Remove the account and used another account today
4. Again login to Enterprise account today

**Observed behavior:**
"Chat History" from yesterday is gone.

### Expected behavior

"Chat History" for Enterprise account should be available.

### Additional context (logs, images, etc)

**Screenshot:**

![Screenshot 2024-02-15 at 12 22 57 AM](https://uploads.linear.app/15b96a4d-ffc0-481a-ba9c-9d840020f7bc/9a02fc39-6d82-4ce5-8d73-bf63c8691588/435123e2-d8cf-4ed9-a418-e2ed04774bbf?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzE1Yjk2YTRkLWZmYzAtNDgxYS1iYTljLTlkODQwMDIwZjdiYy85YTAyZmMzOS02ZDgyLTRjZTUtOGQ3My1iZjYzYzg2OTE1ODgvNDM1MTIzZTItZDhjZi00ZWQ5LWE0MTgtZTJlZDA0Nzc0YmJmIiwiaWF0IjoxNzE3MTAxNTU4LCJleHAiOjMzMjg3NjYxNTU4fQ.du8ugOeCoukUzq8SSwojlRATKHU-xomlV200ajcX7tM "=406x")

jay-fibi:
Update:
It is loading for me now. Not sure what happened atm. Feel free to close it out.

taylorsperry:
Thanks for flagging. We'll monitor this!

mkondratek:
Our persistence seems to show some flakiness occasionally 😕

PR #983, opened 2024-03-06
WIP: Fix persistence flakiness

mkondratek:
Relates to https://github.com/sourcegraph/jetbrains/issues/656 and possibly to https://github.com/sourcegraph/jetbrains/issues/967.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
@@ -9,4 +9,6 @@ class HistoryState : BaseState() {
 
   @get:OptionTag(tag = "defaultEnhancedContext", nameAttribute = "")
   var defaultEnhancedContext: EnhancedContextState? by property()
+
+  fun doIncrementModificationCount() = super.incrementModificationCount()
 }
--- a/src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
@@ -1,6 +1,10 @@
 package com.sourcegraph.cody.history
 
-import com.intellij.openapi.components.*
+import com.intellij.openapi.components.Service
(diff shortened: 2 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
```diff
@@ -9,4 +9,6 @@ class HistoryState : BaseState() {
 
   @get:OptionTag(tag = "defaultEnhancedContext", nameAttribute = "")
   var defaultEnhancedContext: EnhancedContextState? by property()
+
+  fun doIncrementModificationCount() = super.incrementModificationCount()
```
mkondratek, line 13: https://github.com/JetBrains/intellij-community/blob/idea/233.14475.28/platform/projectModel-api/src/com/intellij/openapi/components/BaseState.kt

That file ☝️ mentions that it is our responsibility to call the method.

https://github.com/JetBrains/intellij-community/blob/idea/233.14475.28/platform/projectModel-api/src/com/intellij/openapi/components/PersistentStateComponentWithModificationTracker.java

Instead of above we could just try https://github.com/JetBrains/intellij-community/blob/idea/233.14475.28/platform/projectModel-api/src/com/intellij/openapi/components/PersistentStateComponent.java.

mkondratek: @pkukielka, do we have the steps to reproduce the flakiness issue?

mkondratek: I managed to reproduce it by running the IDE from `:runIde` (the Gradle task from IntelliJ) and killing the process from IntelliJ. I'd expect the same when we kill the process from Activity Monitor. I do not think those cases are important for us - killing a process extraordinary way to exit the IDE. 

(but maybe there are more common scenarios to reproduce the issue?)

mkondratek:
closing this one - we have not received more related reports recently. most likely the issue appears only when we are killing the process in dev runs

PR #989, opened 2024-03-06
Fix url case in add repository

mkondratek:
### Remote urls letter case
1. Try to add the same remote repo url but with upper/mixed-casing

expected: dialog reports that the repo already exists


pkukielka (approved):
LGTM

PR #1005, opened 2024-03-07
Fix formatting for single line completions

### LINKED ISSUE #987: bug: Code is formatted incorrectly after code completion
### Installation Information

Standard installation, all versions

### Describe the bug

image
https://twitter.com/danaketh/status/1765117259234603500

### Expected behavior

Could should always be formatted correctly

### Additional context

IntelliJ formatting have separate action which we could just call after the completion is done.
We try to make sure we insert completion at the right position regarding surrounding text, but I guess it might not be 100% robust. Also it won't help for adjusting e.g. whole completion block to the project coding standard, and using IJ formatter would help there.

pkukielka:
Fixes  https://github.com/sourcegraph/jetbrains/issues/987

## Changes

Fixed formatting (previously non-existing) for single line completions, simplified the code, and added the tests.

## Test plan

Automatic tests for formatter were added.
Additionally manual QA was done using `TESTING.MD` section `Single-line autocomplete' and `Multi-line autocomplete`

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,7 +13,7 @@ platformType=IC
 platformVersion=2022.1
 # Plugin Dependencies -> https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html
 # Example: platformPlugins = com.intellij.java, com.jetbrains.php:203.4449.22
-platformPlugins=Git4Idea,PerforceDirectPlugin
+platformPlugins=Git4Idea,PerforceDirectPlugin,java
 # Java language level used to compile sources and to generate the files for - Java 11 is required for 2020.3 <= x < 2022.2
 javaVersion=11
 # Gradle Releases -> https://github.com/gradle/gradle/releases
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -45,13 +45,13 @@ import com.sourcegraph.utils.CodyFormatter
 import difflib.Delta
 import difflib.DiffUtils
 import difflib.Patch
+import org.eclipse.lsp4j.jsonrpc.ResponseErrorException
 import java.util.concurrent.CancellationException
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionException
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicReference
 import java.util.stream.Collectors
-import org.eclipse.lsp4j.jsonrpc.ResponseErrorException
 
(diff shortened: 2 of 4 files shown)


src/test/kotlin/utils/CodyFormatterTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    TestCase.assertEquals(
+        """|public static int fib(int n) {
+           |        if (n <= 1) {
+           |            return n;
+           |        }
+           |        return fib(n - 1) + fib(n - 2);
+           |    }"""
+            .trimMargin(),
+        formatText(
+            "public static int fib(int n) { if (n <= 1) { return n; } return fib(n-1) + fib(n-2);  }",
```
valerybugakov: 💜

gradle.properties
```diff
@@ -13,7 +13,7 @@ platformType=IC
 platformVersion=2022.1
 # Plugin Dependencies -> https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html
 # Example: platformPlugins = com.intellij.java, com.jetbrains.php:203.4449.22
-platformPlugins=Git4Idea,PerforceDirectPlugin
+platformPlugins=Git4Idea,PerforceDirectPlugin,java
```
mkondratek, line 16: I wonder how limiting is it? Will Cody work in CLion now?

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -371,26 +354,31 @@ class CodyAutocompleteManager {
...
           offset,
           true,
           false,
           Int.MAX_VALUE,
-          CodyAutocompleteBlockElementRenderer(multilineInsertText, items, editor))
+          CodyAutocompleteBlockElementRenderer(completionText, items, editor))
+    } else {
+      inlayModel.addInlineElement(
+          offset,
+          true,
```
mkondratek: can you name trues, falses, etc? 🙏

gradle.properties
pkukielka: It does not matter for generated plugin as long as it's not part of plugin.xml.

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
pkukielka: Fixed.

mkondratek (approved):
The formatting seems to work correctly. 

I see a few issues to are probably not related to this PR directly:
- `fun bubbleSort(array: MutableList<Int>) {` in Kotlin does not work (while the similar code in Java produced the proper complection)
- multiline suggestions seem to appear very rare 
- it's hard to tell that Cody is "thinking" on the autocompletions - we indicator is still broken, we should fix it

PR #1038, opened 2024-03-11
Revert "Move text/descriptions from plugin.xml to the bundle (#988)"

mkondratek:
For some reason action groups does not work after the bundling 😕 let's revert.

## Test plan
1. Open `Settings | Tools | Sourcegraph & Cody`
2. Click ➕ sign (`Add` account) 

Expected:
two options appear


pkukielka (approved):
LGTM

PR #1079, opened 2024-03-13
Simplify `receiveMessage`'s message update handling

mkondratek:
This PR simplifies the logic of handling of the incoming transcript messages. Instead of updating only last two messages let's update the changed messages.

## Test plan
1. chat works, context files appear, restore works


src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -182,22 +181,15 @@ private constructor(
           getCancellationToken().dispose()
         } else {
 
-          if (extensionMessage.chatID != null) {
-            if (prevLastMessage != null) {
-              if (lastMessage?.contextFiles != messages.lastOrNull()?.contextFiles) {
-                val index = extensionMessage.messages.count() - 2
+          if (extensionMessage.chatID != null && extensionMessage.messages != null) {
```
pkukielka, line 184: Thanks, it's much easier to understand now and I think also less error prone!

pkukielka (approved):
LGTM

PR #1080, opened 2024-03-13
Fix enhanced remote context when multiple repos are added

### LINKED ISSUE #1081: bug: Adding multiple remote repos does not work as expected
### Installation Information

Any cody version with support for remote repos

### Describe the bug

When adding multiple repositories only context from last one is properly included

### Expected behavior

All remote repositories should be used

### Additional context

_No response_

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/1081

## Changes

Previously when adding multiple repositories last one added was effectively overriding al the previous ones.
That was leading to lacking context and surprising behaviour depending on the order of the repos added.

## Test plan

Manual tests according to QA testing guide but with multiple repositories added, and enabling/disabling single repositories.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
@@ -61,11 +61,10 @@ class AddRepositoryDialog(
               val codebaseName =
                   runCatching { convertGitCloneURLToCodebaseNameOrError(text) }.getOrNull()
               codebaseName ?: return@custom false
-              val repo =
-                  RemoteRepoUtils.getRepository(project, codebaseName)
-                      .completeOnTimeout(null, 2, TimeUnit.SECONDS)
-                      .get()
-              repo != null
+              RemoteRepoUtils.getRepositories(project, listOf(codebaseName))
+                  .completeOnTimeout(null, 2, TimeUnit.SECONDS)
+                  .get()
+                  .isNotEmpty()
             }
 
     fun validateRepoNotAddedYet() =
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
```diff
@@ -61,11 +61,10 @@ class AddRepositoryDialog(
...
               codebaseName ?: return@custom false
-              val repo =
-                  RemoteRepoUtils.getRepository(project, codebaseName)
-                      .completeOnTimeout(null, 2, TimeUnit.SECONDS)
-                      .get()
-              repo != null
+              RemoteRepoUtils.getRepositories(project, listOf(codebaseName))
+                  .completeOnTimeout(null, 2, TimeUnit.SECONDS)
+                  .get()
+                  .isNotEmpty()
```
mkondratek: nit: I guess it should be a singleton, right?

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -92,10 +92,10 @@ class EnhancedContextPanel(private val project: Project, private val chatSession
       } else {
         CodyAgentCodebase.getInstance(project).getUrl().thenApply { repoUrl ->
           val codebaseName = convertGitCloneURLToCodebaseNameOrError(repoUrl)
-          RemoteRepoUtils.getRepository(project, codebaseName)
+          RemoteRepoUtils.getRepositories(project, listOf(codebaseName))
               .completeOnTimeout(null, 15, TimeUnit.SECONDS)
-              .thenApply { repo ->
-                if (repo != null) {
+              .thenApply { repos ->
+                if (repos.isNotEmpty()) {
```
mkondratek: nit: what if it's `null`?

mkondratek (approved):
tested on 2023.3 - works 🚀

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
pkukielka: Fixed

src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
pkukielka: Changes

PR #1086, opened 2024-03-14
Add instructions for updating cody version

pkukielka:
## Changes

Added a new doc section:

----------------------------------
## Updating Cody version

To upgrade Cody to the newer version you need to update the `cody.commit` hash in `gradle.properties`.
After doing that:

1. Run `git diff --name-status OLD_HASH..NEW_HASH -- agent`  
   It will print you a list of files changed between updates.  
   Changes in tests or bindings might require adjustments in the jetbrains codebase, so please inspect those carefully.
2. Check if the main features are working before sending it for the full QA.  
   Please pay special attention to areas with potential changes indicated by point 1):
    * autocomplete
    * chat
    * chat history
    * context fixes in chat responses
    * @-files in chat prompt
    * enhanced remote context
    * commands from context menu and commands tab
    * login / account switch

   You don't need to cover all corner cases and do a full QA, but make sure basic features still work.  
   Cody version updates tends to break them completely or not at all, so it should allow us to catch most issues early.

----------------------------------

## Test plan
None, just a doc change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,6 +90,28 @@ Take the steps below _before_ [running JetBrains plugin with agent](#developing-
       Note: After [#56254](https://github.com/sourcegraph/sourcegraph/issues/56254) is resolved this step is not needed
       anymore.
 
+## Updating Cody version
+
+To upgrade Cody to the newer version you need to update the `cody.commit` hash in `gradle.properties`.
+After doing that:
+
+1. Run `git diff --name-status OLD_HASH..NEW_HASH -- agent`  
+   It will print you a list of files changed between updates.  
+   Changes in tests or bindings might require adjustments in the jetbrains codebase, so please inspect those carefully.
+2. Check if the main features are working before sending it for the full QA.  
+   Please pay special attention to areas with potential changes indicated by point 1):
+    * autocomplete
+    * chat
+    * chat history
+    * context fixes in chat responses
+    * @-files in chat prompt
+    * enhanced remote context
+    * commands from context menu and commands tab
+    * login / account switch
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -90,6 +90,28 @@ Take the steps below _before_ [running JetBrains plugin with agent](#developing-
       Note: After [#56254](https://github.com/sourcegraph/sourcegraph/issues/56254) is resolved this step is not needed
       anymore.
 
+## Updating Cody version
+
+To upgrade Cody to the newer version you need to update the `cody.commit` hash in `gradle.properties`.
+After doing that:
+
+1. Run `git diff --name-status OLD_HASH..NEW_HASH -- agent`
```
mkondratek: Can you make this instruction more precise? I tried to run it:

```
mkondratek@[pro 14] VL/D/1138  Mikołaj Kondratek scripts % git diff --name-status c8f5355a88783cce6bf1ac0088c73917e2a29d94..632887f9f0cf598579bb0b74855e6761b659d2a0 -- agent
fatal: Invalid revision range c8f5355a88783cce6bf1ac0088c73917e2a29d94..632887f9f0cf598579bb0b74855e6761b659d2a0
```

As you see, it fails 😕

mkondratek (approved):
Ideally, we could have a script that does the update (fetches the latest main commit hash from the agent, sets it, runs diff). Anyway, this PR is an improvement

CONTRIBUTING.md
pkukielka: it should be ran on the Cody repo, I update the instructions.

PR #1093, opened 2024-03-15
Improve cody error reporting to users, fix minor bugs

pkukielka:
## Changes
* Errors during chat operation are reported in chat and not as IJ exceptions
* Better error handling added (more cases covered)
* Some cleanup/unification of error handling code
* Fixed blinking cursor which was disappearing just after context files shows up

image

Screenshot shows Cody first having issues to start, and then hitting error rate limit.

## Test plan

Manual QA according to testing guide

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
@@ -188,32 +188,38 @@ private constructor(
   }
 
   private fun handleChatError(chatError: ChatError) {
-    getCancellationToken().abort()
-    CodyAgentService.setAgentError(project, chatError.message)
+    try {
+      CodyAgentService.setAgentError(project, chatError.message)
 
-    val rateLimitError = chatError.toRateLimitError()
-    val errorMessage =
-        if (rateLimitError != null) {
-          RateLimitStateManager.reportForChat(project, rateLimitError)
-          when {
-            rateLimitError.upgradeIsAvailable ->
-                CodyBundle.getString("chat.rate-limit-error.upgrade")
-            else -> CodyBundle.getString("chat.rate-limit-error.explain")
-          }
-        } else CodyBundle.getString("chat.general-error").fmt(chatError.message, "")
+      val rateLimitError = chatError.toRateLimitError()
+      val errorMessage =
+          if (rateLimitError != null) {
+            RateLimitStateManager.reportForChat(project, rateLimitError)
+            when {
+              rateLimitError.upgradeIsAvailable ->
+                  CodyBundle.getString("chat.rate-limit-error.upgrade")
+              else -> CodyBundle.getString("chat.rate-limit-error.explain")
+            }
+          } else CodyBundle.getString("chat.general-error").fmt(chatError.message, "")
 
-    addErrorMessageAsAssistantMessage(errorMessage)
+      addErrorMessageAsAssistantMessage(errorMessage)
+    } finally {
+      getCancellationToken().abort()
+    }
   }
 
   private fun handleException(e: Exception) {
-    getCancellationToken().abort()
-    CodyAgentService.setAgentError(project, e)
+    try {
+      CodyAgentService.setAgentError(project, e)
 
-    val message = ((e.cause as? CodyAgentException) ?: e).message ?: e.toString()
-    val errorReportLink = CodyErrorSubmitter().getEncodedUrl(e.getThrowableText(), message)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/ui/MessagesPanel.kt
```diff
@@ -34,7 +35,7 @@ class MessagesPanel(private val project: Project, private val chatSession: ChatS
       addChatMessageAsComponent(message)
     }
 
-    if (shouldAddBlinkingCursor) {
+    if (message.speaker == Speaker.ASSISTANT && message.actualMessage().isBlank()) {
```
mkondratek, line 38: nice, the blinking indicator is not visible exactly when it should 🚀

src/main/kotlin/com/sourcegraph/cody/error/CodyErrorSubmitter.kt
```diff
@@ -39,8 +34,18 @@ class CodyErrorSubmitter : ErrorReportSubmitter() {
     return true
   }
 
-  private fun getTitle(event: IdeaLoggingEvent): String {
-    val title = trimPostfix(event.throwableText.lines().first(), 128)
+  public fun getEncodedUrl(throwableText: String, additionalInfo: String? = null): String {
+    return "https://github.com/sourcegraph/jetbrains/issues/new" +
```
mkondratek, line 38: tested - seems to work properly

src/main/resources/CodyBundle.properties
```diff
@@ -45,9 +45,11 @@ chat.rate-limit-error.explain=\
   </html>
 chat.general-error=\
   <html>\
-    &#9888; <b> Cody encountered an error when processing your message:</b><br>\
-    <div marginwidth=15" marginheight="10">{0}</div>\
+   <b> Cody encountered an error when processing your message:</b><br>\
+    <div marginwidth=10" marginheight="10">&#9888;  {0}</div>\
```
mkondratek, line 49: does it work on windows? 🪟

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -105,103 +109,116 @@ private constructor(
...
+    val errorReportLink = CodyErrorSubmitter().getEncodedUrl(e.getThrowableText(), message)
+    val errorReportMsg = CodyBundle.getString("chat.general-error.report").fmt(errorReportLink)
+    addErrorMessageAsAssistantMessage(
+        CodyBundle.getString("chat.general-error").fmt(message, errorReportMsg))
+  }
+
+  private fun addErrorMessageAsAssistantMessage(chatMessage: String) {
     UIUtil.invokeLaterIfNeeded {
-      addMessageAtIndex(ChatMessage(Speaker.ASSISTANT, stringMessage), index)
+      addMessageAtIndex(ChatMessage(Speaker.ASSISTANT, chatMessage), messages.count() - 1)
```
mkondratek, line 227: hmmm... 🤔 `messages.count() - 1` can change... I wonder if we are aware of it's modifications and threading. previously the index was calculated in the pooled thread. Now, it's edt

pkukielka: I think it's a valid concern but highly unlikely.
I will move  `getCancellationToken().abort()` after call of the `addErrorMessageAsAssistantMessage` - that way user surely won't be able to start sending new message until we are fully done there.

src/main/resources/CodyBundle.properties
pkukielka: Yes it does :)

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -105,103 +109,122 @@ private constructor(
...
-    }
+    CodyAgentService.withAgentRestartIfNeeded(
+        project,
+        callback = { agent ->
+          val message =
+              WebviewMessage(
+                  command = "submit",
+                  text = humanMessage.actualMessage(),
+                  submitType = "user",
+                  addEnhancedContext = chatPanel.isEnhancedContextEnabled(),
```
mkondratek, line 134: It's not for this PR but I just realized - there is something strange in our design. Our model logic (`submitMessageToAgent`) seems to depend on the view (the chat panel).

mkondratek (approved):
lgtm

PR #1096, opened 2024-03-15
Add export action

### LINKED ISSUE #405: Export chat history to JSON & CSV
Mimic behavior in VSC added by [this PR](https://github.com/sourcegraph/cody/pull/829).

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/405.

- [x] verify against sourcegraph/cody/pull/3287 or later 
- [x] `chatRestore` instead of `createFromState`
- [x] restore a single chat
- [ ] update cody commit
- [x] test on windows

Related PR to the agent: https://github.com/sourcegraph/cody/pull/3487

| --- | --- |
| --- | --- |
| image | image |

## Test plan
1. tbd


src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,62 @@
...
+      CodyAgentService.withAgent(project) { agent ->
+        val result: ChatHistoryResponse? =
+            agent.server
+                .chatExport(IdParam(connectionId))
+                .completeOnTimeout(null, 15, TimeUnit.SECONDS)
+                .get()
+        if (result != null) {
+          onSuccess.invoke(result)
+        } else {
+          throw Error("Request timed out") // todo: handle it
```
mkondratek: @pkukielka, this is a scenario when we time out. Do you have any suggestions how we should handle that? 🙏

mkondratek:
We found one bug btw - our restored chats have `lastInteractionTimestamp` as the chat id we assign to the chats (e.g. `0c912af5-f2a9-4d60-b8b0-3a75e472a1da`). While VS Code generates and export chats with a "proper" timestamp (e.g. `Tue, 27 Feb 2024 23:55:50 GMT`). 

(Issue: https://github.com/sourcegraph/jetbrains/issues/1152)

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,61 @@
...
+    val accountId = CodyAuthenticationManager.instance.getActiveAccount(project)?.id
+    val chats =
+        HistoryService.getInstance(project)
+            .state
+            .chats
+            .filter { it.accountId == accountId }
+            .filter { it.messages.isNotEmpty() }
+
+    chats.forEachIndexed { index, chatState ->
+      AgentChatSessionService.getInstance(project).getOrCreateFromState(chatState)
```
pkukielka: TBH I do not like that approach although I know it's easiest from the perspective of the code reuse.
We are creating there a lot more than we need, including all UI panels, fetching LLM models, and what not, while all we really want is to call `agent.server.chatRestore`. Maybe we should refactor that bit of logic out and reuse (it will be very close to existing `AgentChatSession::restoreAgentSession` method, no)?

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
pkukielka: We are timing out because of `.completeOnTimeout(null, 15, TimeUnit.SECONDS)` or something else? I do not think we should have restrictive timeout there. If we are concerned about some arbitrary big number of chat being exported (e.g. > 1000) we can check that at the very start and notify user that e.g. only last 1000 chats will be exported.

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
```diff
@@ -0,0 +1,82 @@
...
+            project,
+            onSuccess = { chatHistory ->
+              val json = gson.toJson(chatHistory)
+              invokeLater {
+                WriteAction.run<RuntimeException> {
+                  saveTextToFile(json.toByteArray(), result.file)
+                }
+              }
+            },
+            onFinished = { isRunning = false })
```
pkukielka: `isRunning` is accessed from multiple threads? Please mark it as `volatile`.

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    val result = saveFileDialog.save(outputDir, fileName) ?: return
+
+    ExportChatsBackgroundable(
+            project,
+            onSuccess = { chatHistory ->
+              val json = gson.toJson(chatHistory)
+              invokeLater {
+                WriteAction.run<RuntimeException> {
+                  saveTextToFile(json.toByteArray(), result.file)
```
pkukielka: Use `getVirtualFile` instead of `file`, so you don't need to convert it back and forth risking conversion issues.

mkondratek: good point - thanks

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
mkondratek: it's always AWT

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
mkondratek: I have not encountered the timeout. I can imagine that someone can lose their connection during the export.

If we are considering some constraints we may need to find some more sophisticated way to determine the limit. One can have a single chat with 2000 messages while the other can have 1000 chats with only two messages each (human question, assistant response).

mkondratek: I just tried with a chat that has **over 5000 messages** (the number of assistant messages) - it fails 🔴 

```
2024-03-21 16:42:32,131 [ 342766]   WARN - #c.i.d.PerformanceWatcherImpl - UI was frozen for 5191ms, details saved to /Users/mkondratek/IdeaProjects/jetbrains/build/idea-sandbox/system/log/threadDumps-freeze-20240321-164231-IC-221.5080.210-5sec
java.lang.Throwable: Invocation timed out (1sec)
	at java.desktop/sun.lwawt.macosx.LWCToolkit.invokeAndWait(LWCToolkit.java:818)
	at java.desktop/sun.lwawt.macosx.LWCToolkit.invokeAndWait(LWCToolkit.java:664)
	at java.desktop/sun.lwawt.macosx.CAccessibility.invokeAndWait(CAccessibility.java:114)
	at java.desktop/sun.lwawt.macosx.CAccessibility.invokeAndWait(CAccessibility.java:108)
	at java.desktop/sun.lwawt.macosx.CAccessibility.getFocusOwner(CAccessibility.java:568)
java.lang.OutOfMemoryError: Java heap space
java.lang.OutOfMemoryError: Java heap space
java.lang.OutOfMemoryError: Java heap space
java.lang.OutOfMemoryError: Java heap space
java.lang.Throwable: Invocation timed out (1sec)
	at java.desktop/sun.lwawt.macosx.LWCToolkit.invokeAndWait(LWCToolkit.java:818)
	at java.desktop/sun.lwawt.macosx.LWCToolkit.invokeAndWait(LWCToolkit.java:664)
	at java.desktop/sun.lwawt.macosx.CAccessibility.invokeAndWait(CAccessibility.java:114)
	at java.desktop/sun.lwawt.macosx.CAccessibility.invokeAndWait(CAccessibility.java:108)
	at java.desktop/sun.lwawt.macosx.CAccessibility.getFocusOwner(CAccessibility.java:568)
Exception in thread "HttpClient-1-SelectorManager" java.lang.OutOfMemoryError: Java heap space
```

I tried with ~1500 messages (the number of assistant messages) - it works (I was able to export) but IntelliJ complains about running low on memory.

I tried with 1000 messages (the number of assistant messages) - it works - no complaints.

mkondratek:
I ve just found some designs (https://github.com/sourcegraph/jetbrains/issues/405) 😅 

image
image

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
```diff
@@ -74,8 +91,10 @@ interface CodyAgentServer {
   @JsonRequest("chat/submitMessage")
   fun chatSubmitMessage(params: ChatSubmitMessageParams): CompletableFuture<ExtensionMessage>
 
-  @JsonRequest("chat/models")
-  fun chatModels(params: ChatModelsParams): CompletableFuture<ChatModelsResponse>
+  @JsonRequest("chat/models") fun chatModels(params: IdParam): CompletableFuture<ChatModelsResponse>
+
+  @JsonRequest("chat/export")
+  fun chatExport(params: IdParam): CompletableFuture<ChatHistoryResponse?>
```
dominiccooney: This does not take an ID any more on the extension side.

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
dominiccooney: What's the purpose of reviving the chat here? `chat/export` below gives you all the chats in local history for the authed account.

mkondratek: Some account may not be available in the local history in the agent. We need to restore them from our persistency state so the agent is aware of them.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
mkondratek: thanks

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,74 @@
...
+    val chats =
+        HistoryService.getInstance(project)
+            .state
+            .chats
+            .filter { it.accountId == accountId }
+            .filter { it.messages.isNotEmpty() }
+            .filter { it.internalId != null }
+            .filter { chat -> if (internalId != null) chat.internalId == internalId else true }
+
+    AgentChatSession.createNew(project) { _ ->
```
pkukielka: Why do you even need that call? You are not using connectionId generated by it in any way.

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,74 @@
...
+      CodyAgentService.withAgent(project) { agent ->
+        chats.forEachIndexed { index, chatState ->
+          restoreChatSession(agent, chatState)
+          indicator.fraction = ((index + 1.0) / (chats.size + 1.0))
+          if (indicator.isCanceled) {
+            return@withAgent
+          }
+        }
+
+        val result = agent.server.chatExport().completeOnTimeout(null, 15, TimeUnit.SECONDS).get()
```
pkukielka: I do not like that random timeout values in the code.
Maybe we can use one default in the agent itself, or if we need a few tiers just have one object which defines them.

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,74 @@
...
+        if (indicator.isCanceled) {
+          return@withAgent
+        }
+
+        if (result != null) {
```
pkukielka, line 45: Could be one if to decrease amount of indentation levels

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,74 @@
...
+          if (internalId != null) {
+            val singleChatHistory = result.find { it.chatID == internalId }
+            if (singleChatHistory != null) {
+              onSuccess.invoke(singleChatHistory)
+            } else {
+              // todo: handle error
```
pkukielka: TODO in the code

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
```diff
@@ -0,0 +1,73 @@
...
+            onSuccess = { chatHistory ->
+              val json = gson.toJson(chatHistory)
+              invokeLater {
+                WriteAction.run<RuntimeException> {
+                  result.getVirtualFile(true)?.setBinaryContent(json.toByteArray())
+                  VirtualFileManager.getInstance().syncRefresh()
+                }
+              }
+            },
+            onFinished = { isRunning = false })
```
pkukielka: I do not like passing state variable as general purpose callback.
We already have domain purpose class exactly for what you need there: `CancellationToken`

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
mkondratek: whoops - that are some leftover - good point, thanks

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
mkondratek: I do not understand what do you mean by `default`. I tried `.getNow(null)` but it gives `null`.

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
mkondratek: | timeout | error |
| --- | --- |
| image | image |

@danielmarquespt, can you pls take a look? 🙏

mkondratek: Note that it is a "tool window" notification (the balloon points to the tool window tab - `Cody`, it's less aggressive than the usual notifications)

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,96 @@
...
+            .filter { it.accountId == accountId }
+            .filter { it.messages.isNotEmpty() }
+            .filter { it.internalId != null }
+            .filter { chat -> if (internalId != null) chat.internalId == internalId else true }
+
+    chats.forEachIndexed { index, chatState ->
+      restoreChatSession(agent, chatState)
+      indicator.fraction = ((index + 1.0) / (chats.size + 1.0))
+
+      Thread.sleep(3000)
```
pkukielka: What is that? :O

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,96 @@
...
+      if (internalId != null) {
+        val singleChatHistory = result.find { it.chatID == internalId }
+        if (singleChatHistory != null) {
+          onSuccess.invoke(singleChatHistory)
+        } else {
+          logger.warn("export failed: singleChatHistory is null")
+
+          val notification =
+              Notification(
+                  "Sourcegraph Cody",
```
pkukielka, line 61: That probably should be defined in the messages bundle.

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -0,0 +1,96 @@
...
+import com.sourcegraph.cody.initialization.EndOfTrialNotificationScheduler
+import com.sourcegraph.cody.vscode.CancellationToken
+import java.util.concurrent.TimeUnit
+
+class ExportChatsBackgroundable(
+    project: Project,
+    private val agent: CodyAgent,
+    private val internalId: String?,
+    private val onSuccess: (Any) -> Unit,
+    private val cancellationToken: CancellationToken
```
pkukielka, line 23: Thank you for changing it to use `CancellationToken`!

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
```diff
@@ -0,0 +1,85 @@
...
+                }
+              },
+              cancellationToken = token)
+          .queue()
+    }
+  }
+
+  companion object {
+    val gson: Gson = GsonBuilder().create()
+    var isRunning = AtomicReference(false)
```
pkukielka, line 77: I see what you mean talking about `AtomicReference`. 
It seems that if you need to check it externally there really is no easier way (although that probably could be just volatile boolean).

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
mkondratek: whoops, sorry - that was for testing only

src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
mkondratek: This is more a key than a message to be bundled (it is referenced by `plugin.xml`). Let me extract all notification group keys in the other PR: https://github.com/sourcegraph/jetbrains/pull/1204

mkondratek:
tested on windows - works

image

pkukielka (approved):
LGTM

Issue #550, opened 2024-02-08
JetBrains: Switching between Pro to Free will not get reflected on JetBrains IDE and requires restart


jay-fibi:
### Installation Information

Version:
5.2.21709

### Describe the bug

**Steps to reproduce:**

1. Have Cody plugin installed in your JetBrains PyCharm editor
2. User is already logged in to Cody with PRO subscription
3. Go to [Sourcegraph cody subscription] (https://sourcegraph.com/cody/subscription) page
4. Click on "Cancel subscription" and select "Switch to free"
5. Go to PyCharm editor and check "Cody > Subscription" tab

**Observed behavior:**
It should automatically change the current subscription

### Expected behavior

Subscription keep showing "Cody Pro" as Current tier.

### Additional context (logs, images, etc)

**Note:**
This change gets reflected immediately on VS Code editor.

**Video link:**
https://drive.google.com/file/d/1Ly7iFYYm70pxq0XsB_r5DTtvbzHiXy1X/view?usp=drive_link

taylorsperry:
When a user upgrades Free to Pro, there's a notice on the subscription panel that the user needs to restart the IDE for the upgrade to take effect. We should at least include the same notice when a user *downgrades* from Pro to Free. 

The ideal scenario, of course, is that they don't need to restart at all, but I suspect that's a larger effort (in which case it's fine to address after Feb 15).

Issue #454, opened 2024-02-01
Exception `java.io.IOException: Stream Closed` after account change with ongoing conversation


exigow:
### Steps to reproduce:

1. Ask Cody and wait for response.
2. Go to settings and select different account - exception occurs.

NOTE: Account switch was from "free" to "enterprise", but this is probably not related.

```
024-02-01 13:10:05,192 [  35624]   WARN - o.e.l.j.RemoteEndpoint - Failed to send notification message.
org.eclipse.lsp4j.jsonrpc.JsonRpcException: java.io.IOException: Stream Closed
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageConsumer.consume(StreamMessageConsumer.java:72)
	at org.eclipse.lsp4j.jsonrpc.TracingMessageConsumer.consume(TracingMessageConsumer.java:114)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.notify(RemoteEndpoint.java:126)
	at org.eclipse.lsp4j.jsonrpc.services.EndpointProxy.invoke(EndpointProxy.java:88)
	at com.sun.proxy.$Proxy133.configurationDidChange(Unknown Source)
	at com.sourcegraph.cody.config.notification.AccountSettingChangeListener$1.afterAction$lambda$0(AccountSettingChangeListener.kt:27)
	at com.sourcegraph.cody.agent.CodyAgentService$Companion$applyAgentOnBackgroundThread$1.invoke$lambda$0(CodyAgentService.kt:105)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:295)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:668)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:665)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
[...]

PR #1132, opened 2024-03-19
Add logging of user tier in the telemetry

pkukielka:
Fixes:
https://github.com/sourcegraph/jetbrains/issues/550
https://github.com/sourcegraph/cody/issues/2467

## Changes

1. Added `CodyAccount::getActiveAccountTier` method with result caching.
2. Added tier logging in `GraphQlLogger`
3. Refactored account change notification logic to be concentrated in one place (`CodyAccount::setActiveAccount`)
4. Added periodic and on focus change querying of the account state (see `CodyAuthenticationManager::init`)

## Test plan

Manual verification if proper entries are in the GraphQL


src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
```diff
@@ -63,35 +65,48 @@ class MyAccountTabPanel : JPanel() {
     }
   }
 
-  private fun createCenterPanel() = panel {
-    val getIsCurrentUserPro = isCurrentUserPro
+  private fun createCenterPanel(isPro: Boolean?) = panel {
     val tier =
-        if (getIsCurrentUserPro == null) CodyBundle.getString("my-account-tab.loading-label")
-        else if (getIsCurrentUserPro) CodyBundle.getString("my-account-tab.cody-pro-label")
+        if (isPro == null) CodyBundle.getString("my-account-tab.loading-label")
```
mkondratek: nit: I prefer descriptive names over short names - in this case `isCurrentUserPro` would look better imho

src/main/kotlin/com/sourcegraph/cody/config/CodyAccount.kt
```diff
@@ -22,14 +25,34 @@ data class CodyAccount(
...
-  fun isDotcomAccount() = server.url.lowercase().startsWith(ConfigUtil.DOTCOM_URL)
+  fun isDotcomAccount(): Boolean = server.url.lowercase().startsWith(ConfigUtil.DOTCOM_URL)
+
+  fun isEnterpriseAccount(): Boolean = isDotcomAccount().not()
+
+  fun isFreeUser(project: Project): CompletableFuture<Boolean> =
+      if (isDotcomAccount()) isProUser(project).thenApply { it.not() }
+      else CompletableFuture.completedFuture(false)
+
+  fun isProUser(project: Project): CompletableFuture<Boolean> {
```
mkondratek: maybe `deriveIsCurrentUserPro`? to be more precise and reflect the fact that it's not an instant quick "get" but a call to agent

mkondratek: oh, I just realised - we are caching the pro state now... 🤔 

When a user upgrades to pro they are informed that they need to restart the IDE to see the changes. What when a user downgrades to free during the usage?

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
```diff
@@ -63,35 +65,48 @@ class MyAccountTabPanel : JPanel() {
...
+        else if (isPro) CodyBundle.getString("my-account-tab.cody-pro-label")
         else CodyBundle.getString("my-account-tab.cody-free-label")
     row { label("<html>Current tier: <b>$tier</b><html/>") }
     row {
-      if (getIsCurrentUserPro != null && !getIsCurrentUserPro) {
+      if (isPro != null && !isPro) {
```
mkondratek: nit: isn't that simply `isPro == false`?

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
```diff
@@ -63,35 +65,48 @@ class MyAccountTabPanel : JPanel() {
...
         val upgradeButton =
             button("Upgrade") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/subscription") }
         upgradeButton.component.putClientProperty(DarculaButtonUI.DEFAULT_STYLE_KEY, true)
       }
       button("Check Usage") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/manage") }
     }
-    if (getIsCurrentUserPro != null && !getIsCurrentUserPro) {
+    if (isPro != null && !isPro) {
```
mkondratek: nit: isn't that simply `isPro == false`?

src/main/java/com/sourcegraph/cody/MyAccountTabUtils.kt
```diff
@@ -1,57 +0,0 @@
...
-    }
-
-    return result
-  }
-
-  return CompletableFuture.completedFuture(null)
-}
-
-@RequiresBackgroundThread
-private fun ensureUserIdMatchInAgent(jetbrainsUserId: String, server: CodyAgentServer) {
```
mkondratek, line 42: so this is redundant? nice 🚀

src/main/kotlin/com/sourcegraph/cody/config/CodyAccount.kt
```diff
@@ -22,14 +25,34 @@ data class CodyAccount(
...
 ) : ServerAccount() {
 
-  fun getAccountType(): AccountType {
-    if (isDotcomAccount()) {
-      return AccountType.DOTCOM
-    }
-    return AccountType.ENTERPRISE
+  @Volatile private var isProUser: Boolean? = null
+
+  fun getAccountType(project: Project): String {
```
mkondratek: I don't like that change. The difference b/w enterprise and free/pro is different than the difference b/w free and pro. We can call the first category "account type" and the second a "tier". And my feeling is that we are adjusting the model to the telemetry requirements (which feel bad).

If we really need to consider enterprise/free/pro as a same category maybe let's introduce an enum for them. Is `AccountType` still needed or should we deprecate it? 

Also, if we do the agent call here the method name should be altered to reflect the hidden process - "derive" instead of "get"?

src/main/java/com/sourcegraph/cody/MyAccountTabUtils.kt
mkondratek: tested - indeed it seems to work

src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
```diff
@@ -44,9 +43,9 @@ class LlmDropdown(
     data.models.forEach(::addItem)
     data.models.find { it.default }?.let { this.selectedItem = it }
 
-    val activeAccountType = CodyAuthenticationManager.instance.getActiveAccount(project)
-    isEnabled =
-        !didSendFirstMessage && !(activeAccountType.isEnterpriseAccount() || model.size <= 1)
+    val isEnterpriseAccount =
+        CodyAuthenticationManager.instance.getActiveAccount(project)?.isEnterpriseAccount() ?: false
+    isEnabled = !didSendFirstMessage && !isEnterpriseAccount && model.size > 1
```
mkondratek, line 49: hah - it looks much better now 😅 

image

src/main/java/com/sourcegraph/telemetry/GraphQlLogger.java
```diff
@@ -14,7 +15,9 @@ public class GraphQlLogger {
   public static CompletableFuture<Boolean> logInstallEvent(@NotNull Project project) {
     CodyApplicationSettings codyApplicationSettings = CodyApplicationSettings.getInstance();
     if (codyApplicationSettings.getAnonymousUserId() != null && !project.isDisposed()) {
-      var event = createEvent(ConfigUtil.getServerPath(project), "CodyInstalled", new JsonObject());
+      var event =
+          createEvent(
+              project, ConfigUtil.getServerPath(project), "CodyInstalled", new JsonObject());
```
mkondratek, line 20: btw, can we have an enum class for the event name? 🙏 it would be easier to navigate b/w then and see which are already in the code in a one place

src/main/java/com/sourcegraph/telemetry/GraphQlLogger.java
```diff
@@ -65,9 +75,15 @@ private static Event createEvent(
 
   @NotNull
   private static JsonObject addGlobalEventParameters(
```
mkondratek, line 77: ah, now I see why we need to cache the tier... 🕵️

mkondratek (changes requested):
The issue: https://github.com/sourcegraph/cody/issues/2467

---

comments, comments everywhere 🗯️

src/main/java/com/sourcegraph/telemetry/GraphQlLogger.java
mkondratek, line 1: I just noted new additional entries in the logs:

```
2024-03-19 23:48:44,484 [  20502]   WARN - #c.s.c.a.CodyAgentClient - Cody by Sourcegraph: █ GraphQLTelemetryExporter: telemetry: failed to evaluate server version: Error: accessing Sourcegraph GraphQL API: Error: HTTP status code 401: Private mode requires authentication.
 (https://sg02.sourcegraphcloud.com/.api/graphql?SiteProductVersion)
2024-03-19 23:48:44,867 [  20885]   WARN - #c.s.c.a.CodyAgentClient - Cody by Sourcegraph: █ GraphQLTelemetryExporter: Error exporting telemetry events: Error: accessing Sourcegraph GraphQL API: Error: HTTP status code 401: Private mode requires authentication.
```

Steps:
1. `runIde`
2. switch accounts from the settings

I tested it against v5.4.5 - no `GraphQLTelemetryExporter` entries there in the same scenario

mkondratek: I think it's a false alarm - sorry for confusion. 

The issue (https://github.com/sourcegraph/jetbrains/issues/454#issuecomment-2008334284) seems to be there already. The two mentioned lines of logs seem to appear when switching to `sg02` instance account (in particular).

src/main/kotlin/com/sourcegraph/cody/config/CodyAccount.kt
pkukielka: I can move this code the the logger as it's not user anywhere else so far.

src/main/kotlin/com/sourcegraph/cody/config/CodyAccount.kt
pkukielka: > What when a user downgrades to free during the usage?\

Good question :)
https://github.com/sourcegraph/jetbrains/issues/550
I will address that and some other improvements in a follow-up PR.

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
pkukielka: Fixed

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
pkukielka: Fixed

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -17,14 +29,85 @@ internal class CodyAuthData(val account: CodyAccount, login: String, token: Stri
...
+  DOTCOM_FREE("free"),
+  DOTCOM_PRO("pro"),
+  ENTERPRISE("enterprise")
+}
+
 /** Entry point for interactions with Sourcegraph authentication subsystem */
-@Service
-class CodyAuthenticationManager internal constructor() {
+@Service(Service.Level.PROJECT)
+class CodyAuthenticationManager(val project: Project) : Disposable {
```
mkondratek, line 40: good point about revamping it into a project-level service 👍

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeContext.kt
```diff
@@ -2,5 +2,8 @@ package com.sourcegraph.cody.config.notification
 
 class AccountSettingChangeContext(
     val serverUrlChanged: Boolean = false,
-    val accessTokenChanged: Boolean = false
-)
+    val accessTokenChanged: Boolean = false,
+    val accountTierChanged: Boolean = false
```
mkondratek, line 6: 💟

src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
```diff
@@ -5,24 +5,20 @@ import com.intellij.openapi.ui.ComboBox
...
-)
-
 class LlmDropdown(
     private val project: Project,
     private val onSetSelectedItem: (ChatModelsResponse.ChatModelProvider) -> Unit,
     private val chatModelProviderFromState: ChatModelsResponse.ChatModelProvider?,
 ) : ComboBox<ChatModelsResponse.ChatModelProvider>(MutableCollectionComboBoxModel()) {
 
   private var didSendFirstMessage: Boolean = false
+  var isCurrentUserFree = true
```
mkondratek, line 21: iirc that is the place you mentioned - the scenario is that a user has an open chat with no message in and changes their account tier - that change will not be reflected in the dropdown. is that right?

can you please put a comment on this field? this place looks a bit extraordinary compared to the rest of the codebase after your changes

src/main/kotlin/com/sourcegraph/cody/auth/ui/AccountsPanelFactoryExtensions.kt
```diff
@@ -152,8 +158,6 @@ fun <A : Account, Cred> Row.customAccountsPanel(
     }
     val newTokensMap = accountsModel.accounts.associateWith { null }
     accountManager.updateAccounts(newTokensMap)
-    val activeAccount = accountsModel.activeAccount
-    activeAccountHolder.account = activeAccount
```
mkondratek, line 156: iiuc this logic (setting the active account in the holder) is executed anyway by `CodyAuthenticationManager.getInstance(project).setActiveAccount(activeAccount)`

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeContext.kt
mkondratek: but it's unused 🤔

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
```diff
@@ -21,12 +20,7 @@ class AccountSettingChangeListener(project: Project) : ChangeListener(project) {
...
 
-            // Notify Cody Agent about config changes.
-            if (ConfigUtil.isCodyEnabled()) {
-              CodyAgentService.withAgentRestartIfNeeded(project) { agent ->
-                agent.server.configurationDidChange(ConfigUtil.getAgentConfiguration(project))
-              }
-              CodyAgentService.getInstance(project).restartAgent(project)
+            if (ConfigUtil.isCodyEnabled() && context.accountChanged()) {
               CodyToolWindowContent.executeOnInstanceIfNotDisposed(project) {
                 removeAllChatSessions()
```
mkondratek, line 25: hmm, if I change the token it makes `accountChanged == true` but that is not necessarily mean the account changed - I mean - I can change my own token to another token of the same account

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
pkukielka: Fixed

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeListener.kt
pkukielka: Maybe we should adjust the name. `accountConfigurationChanged`?

src/main/kotlin/com/sourcegraph/cody/auth/ui/AccountsPanelFactoryExtensions.kt
pkukielka: Correct

src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
pkukielka: I think we should just fix it, I kind of forgot about it afterwards.

src/main/kotlin/com/sourcegraph/cody/config/notification/AccountSettingChangeContext.kt
pkukielka: It was added mostly for clarity so e.g. in debugger when you will notice `AccountSettingChangeContext` it will be clear why it was emitted. With just two other properties set to false it could look like a bug.

PR #1158, opened 2024-03-21
Removed experimental user setting for inline edits

steveyegge:
We now have an environment-variable based feature flag, so the setting is no longer needed nor helpful.

## Test plan

Locally tested, and all unit tests pass. This work is helping speed up delivery of the integration tests.


pkukielka (approved):
LGTM

PR #1178, opened 2024-03-25
Refactor restore logic

mkondratek:
This is a minor refactor needed for https://github.com/sourcegraph/jetbrains/pull/1096.

It extracts the chat restore logic so it is doable to send the restore request without restoring the full session (with panels, etc).

## Test plan
1. Restore works properly

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
@@ -64,29 +64,6 @@ private constructor(
     fetchLlms()
   }
 
-  fun restoreAgentSession(
-      agent: CodyAgent,
-      chatModelProvider: ChatModelsResponse.ChatModelProvider? = null
-  ) {
-    val messagesToReload =
-        messages
-            .toList()
-            .dropWhile { it.speaker == Speaker.ASSISTANT }
-            .fold(emptyList<ChatMessage>()) { acc, msg ->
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
```diff
@@ -339,48 +324,54 @@ private constructor(
...
-        ApplicationManager.getApplication().invokeLater {
-          chatSession.chatPanel.addAllMessages(chatMessages)
-        }
-
-        if (modelFromState?.model.isNullOrEmpty()) {
-          chatSession.restoreAgentSession(codyAgent)
-        } else {
-          chatSession.restoreAgentSession(codyAgent, modelFromState)
-        }
+        val dummyConnectionId = codyAgent.server.chatNew()
```
pkukielka, line 368: Why it's `dummyConnectionId` ant not just `connectionId`?

pkukielka (approved):
LGTM except one question about variable name

src/main/kotlin/com/sourcegraph/cody/chat/AgentChatSession.kt
mkondratek: this is not the one we use eventually. pls take a look:

```
fun updateFromState(agent: CodyAgent, state: ChatState) {
    val newConnectionId = restoreChatSession(agent, state)
    connectionId.getAndSet(newConnectionId)
    fetchLlms()
  }
```

PR #1198, opened 2024-03-27
Add repo dialog will accept repo names which don't contain a /

### LINKED ISSUE #1194: Multi-repo validation code shouldn't require `/` characters
Prospect kicking off a trial this week isn't able to use multi-repo because their repos don't include `/` characters. This issue won't be relevant once we ship https://github.com/sourcegraph/jetbrains/issues/982, but we need to fix this for them in the meantime. [Thread for context](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1711477931175429).

dominiccooney:
This speculatively asks the remote to match the raw user input *and* the user input with the clone URL to repo name heuristic applied to it.

VSCode does not have this issue because repos are picked from a list retrieved from the remote, instead of input manually.

Fixes #1194

## Test plan

Tested manually:

- Breaking in `validateRepoExists` and check that we do fetch repo IDs for repo names without `/`, eg `foo`
- Check remotes like `ssh@github.com:sourcegraph/jetbrains` still work
- Check duplicate checks are still insensitive to exact input syntax, eg `ssh@github.com:sourcegraph/jetbrains` is detected as a duplicate of `github:sourcegraph/jetbrains` and `https://github.com/sourcegraph/jetbrains.git`, etc.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/context/ui/AddRepositoryDialog.kt
@@ -36,8 +36,34 @@ class AddRepositoryDialog(
     setValidationDelay(100)
   }
 
+  private var validatedInput: String? = null
+  private var validatedRepoName: CodebaseName? = null
+
+  /**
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
LGTM

PR #1204, opened 2024-03-27
Extract NotificationGroups.kt

mkondratek:
## Test plan
1. nothing specific - it's just a minor refactor

### DIFF
--- a/src/main/kotlin/com/sourcegraph/common/NotificationGroups.kt
+++ b/src/main/kotlin/com/sourcegraph/common/NotificationGroups.kt
@@ -1,0 +1,9 @@
+package com.sourcegraph.common
+
+object NotificationGroups {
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/sourcegraph/common/NotificationGroups.kt
```diff
@@ -0,0 +1,9 @@
+package com.sourcegraph.common
+
+object NotificationGroups {
```
pkukielka, line 3: We already have a message bundle, why there we want to use separate object?

mkondratek: these are not messages - those are notification group IDs

pkukielka (approved):
LGTM

PR #1207, opened 2024-03-27
Switch to `chatID`-agnostic `chat/models` API

mkondratek:
Cody part: https://github.com/sourcegraph/cody/pull/3588

## Test plan
1. LLM selection works
2. Chat restore works (proper llm is selected and used)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -24,4 +24,4 @@ gradleVersion=8.1.1
 kotlin.stdlib.default.dependency=false
 # See https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_configure
 cody.autocomplete.enableFormatting=true
-cody.commit=8f8218cfb76689e195d3d80fbd6adb14458f3396
+cody.commit=89722171622c5e6df5356b1dd4fad7ad9afe18d1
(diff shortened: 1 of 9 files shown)


pkukielka (approved):
LGTM

dominiccooney (approved):
LGTM too.

The bit about the `codyProOnly = false` when restoring the chat session I don't get... I realize you're just moving this in this PR, but it seems dodgy. Likewise `default = it.model == null ... model = it.model ?: ""` What happens when these restored chats are updated? The round trip will change `model` from `null` to `""` and `default` to `false` etc.

mkondratek:
> LGTM too.
> 
> The bit about the `codyProOnly = false` when restoring the chat session I don't get... I realize you're just moving this in this PR, but it seems dodgy. Likewise `default = it.model == null ... model = it.model ?: ""` What happens when these restored chats are updated? The round trip will change `model` from `null` to `""` and `default` to `false` etc.

@dominiccooney, after you set a chat and send a first human message. It is fixed afterwards. To do the restore all we need is the model (`ChatModelProvider::model`). The rest of the fields are either redundant (`codyProOnly`) or needed for UI only (like `title` and `provider` - it's used only in the presentation).

**I agree this part of code is not perfect** - I will rethink it and discuss with Peter.

PR #1229, opened 2024-03-28
Add toggle action for enabling inline edits

pkukielka:
## Changes

I added action which allow user to enable inline edits.
To do it:
1. Hit `shift shift`
2. Start typing "Enable Inline Edits"
3. When "Enable Inline Edits" action will be visible and selected hit `Enter` to toggle it on
4. Restart IDE

You need to do it only once, your choice wll be saved for later.
You can disable it the same way.

## Test plan

Manual verification that it works, it's internal action

image

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/config/CodyApplicationSettings.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/CodyApplicationSettings.kt
@@ -22,6 +22,7 @@ data class CodyApplicationSettings(
     var isOnboardingGuidanceDismissed: Boolean = false,
     var shouldAcceptNonTrustedCertificatesAutomatically: Boolean = false,
     var shouldCheckForUpdates: Boolean = true,
+    var isInlineEditionEnabled: Boolean = false
 ) : PersistentStateComponent<CodyApplicationSettings> {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/config/CodyApplicationSettings.kt
```diff
@@ -41,6 +42,8 @@ data class CodyApplicationSettings(
     this.isOnboardingGuidanceDismissed = state.isOnboardingGuidanceDismissed
     this.shouldAcceptNonTrustedCertificatesAutomatically =
         state.shouldAcceptNonTrustedCertificatesAutomatically
+    this.shouldCheckForUpdates = state.shouldCheckForUpdates
```
mkondratek, line 45: was it a bug? 😅

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/config/CodyApplicationSettings.kt
pkukielka: I think so :) Maybe the reason updater was not working? ;D

PR #1232, opened 2024-03-28
Simplify `next-release.sh`

mkondratek:
## Test plan
1. N/A

### DIFF
--- a/scripts/next-release.sh
+++ b/scripts/next-release.sh
@@ -13,35 +13,10 @@ if [ "$#" -ne 1 ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


dominiccooney (commented):
Why change this now?

mkondratek:
> Why change this now?

@dominiccooney, we improved the versioning recently by "disconnecting" it from the sourcegraph relases. Now we realised we can go further. The former buildnum versioning does not look good, cannot be controlled, and is very complicated. Now, with a new release script created by Peter we can easily create new releases and bump major/minor/patch one by one.

dominiccooney (approved):
Ok, sounds fine

PR #1263, opened 2024-04-02
Set default LLM to the last one used

pkukielka:
## Changes

User LLM selection is now saved and used for subsequest chats/edits.
TODO: new API for commands is needed to suport the same.

## Test plan

Manual tests according to QA guide

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/EditCommandPrompt.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/EditCommandPrompt.kt
@@ -48,7 +48,7 @@ class EditCommandPrompt(val controller: FixupService, val editor: Editor, val di
       LlmDropdown(
           modelUsage = ModelUsage.EDIT,
           project = controller.project,
-          onSetSelectedItem = controller::setCurrentModel,
+          onSetSelectedItem = {},
           chatModelProviderFromState = null)
 
   // History navigation helper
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -53,10 +50,6 @@ class FixupService(val project: Project) : Disposable {
     return true
   }
 
-  fun setCurrentModel(model: ChatModelsResponse.ChatModelProvider) {
```
pkukielka, line 56: Value never used.

src/main/kotlin/com/sourcegraph/cody/edit/EditCommandPrompt.kt
```diff
@@ -48,7 +48,7 @@ class EditCommandPrompt(val controller: FixupService, val editor: Editor, val di
       LlmDropdown(
           modelUsage = ModelUsage.EDIT,
           project = controller.project,
-          onSetSelectedItem = controller::setCurrentModel,
+          onSetSelectedItem = {},
```
mkondratek, line 51: this callback should persist the default for the inline edits, right?

mkondratek (approved):
lgtm

src/main/kotlin/com/sourcegraph/cody/edit/EditCommandPrompt.kt
pkukielka: No, @danielmarquespt said the default between edits and chats should be the same.
We are already saving the default in `LlmDropdown::setSelectedItem`

PR #1274, opened 2024-04-02
Fix restore after export

### LINKED ISSUE #1247: JetBrains: Chats stop loading when opening from history after exporting all chats.
### Cody Version

v5.5.1-nightly

### IDE Information

IntelliJ IDEA 2023.3.4 (Community Edition)
Build #IC-233.14475.28, built on February 13, 2024
Runtime version: 17.0.10+1-b1087.17 amd64
Windows 11.0
Memory: 2048M
Cores: 12
Registry:
ide.experimental.ui=true
Non-Bundled Plugins:
PythonCore (233.14475.9)
com.sourcegraph.jetbrains (5.5.1-nightly)
Kotlin: 233.14475.28-IJ

### Describe the bug

Platform: Windows

**Steps to reproduce:**

1. Have Cody plugin installed in your JB PyCharm editor
2. Login to Cody with Free/Pro/Enterprise user
3. Open Cody chat window and navigate to "Chat History"
4. Click on "Export All Chats As JSON" button and export the chats
5. Select any chat from the chat history

### Observed behavior:
Chat will not be loaded from chat history after exporting it.

### Expected behavior

Chat should be loaded correctly from chat history.

### Additional context

**Video link:**

https://drive.google.com/file/d/1Tsrr_Uyaz6InvrYrOHFmz-twPLNDH0IS/view?usp=drive_link

mkondratek:
Fixes #1247.

It could be a temporary solution to the original problem. The ideal solution is to make chat restore idempotent.

## Test plan
1. Export a chat.
2. Restore the same chat.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
@@ -17,6 +17,7 @@ import com.sourcegraph.cody.initialization.EndOfTrialNotificationScheduler
 import com.sourcegraph.cody.vscode.CancellationToken
 import com.sourcegraph.common.CodyBundle
 import com.sourcegraph.common.NotificationGroups
+import java.util.UUID
 import java.util.concurrent.TimeUnit
 
 class ExportChatsBackgroundable(
@@ -40,6 +41,7 @@ class ExportChatsBackgroundable(
             .filter { it.internalId != null }
             .filter { chat -> if (internalId != null) chat.internalId == internalId else true }
 
+    var dummyInternalId = UUID.randomUUID().toString()
     chats.forEachIndexed { index, chatState ->
       val chatMessages =
           chatState.messages.map { message ->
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/ExportChatsBackgroundable.kt
```diff
@@ -40,6 +41,7 @@ class ExportChatsBackgroundable(
             .filter { it.internalId != null }
             .filter { chat -> if (internalId != null) chat.internalId == internalId else true }
 
+    var dummyInternalId = UUID.randomUUID().toString()
```
pkukielka, line 44: Does it also work for restore all?
Or, in other words, we just need one `dummyInternalId` or one per restored chat?

mkondratek: we need one per chat and that's what I do here (see line 58)

pkukielka (approved):
LGTM

PR #1275, opened 2024-04-03
Change release script to determine next version based on tags

pkukielka:
## Changes

1. Do not force push to release tags/branches (if you need to changes something after release use patch release)
2. Do not use official GH releases to determine versions, as we no longer create those for the QA/nightly

## Test plan

N/A


mkondratek (approved):
lgtm

PR #1280, opened 2024-04-03
add claude 3 as free default

taylorsperry:
Adds Claude 3 Sonnet as the default model for Free users (and new Pro users). 

## Test plan

Passing CI

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -467,7 +467,7 @@ To open the context menu:
 
 1. Login to Cody Free account
 2. Create new chat
-3. User doesn't see model dropdown
+3. User sees model dropdown, but non-default LLM is disabled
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -466,6 +468,7 @@ To open the context menu:
 1. Login to Cody Free account
 2. Create new chat
 3. User doesn't see model dropdown
```
mkondratek: I think it is not true for a while already. We do display the dropdown with CTA for the Pro items. If I am right, you can fix that btw

mkondratek (approved):

TESTING.md
taylorsperry: Nice catch, updated!

PR #1283, opened 2024-04-04
Update TESTING.md

aramaraju:
## Test plan

Added test plans for verifying JBs client works for Enterprise customers, specifically Cody Gateway and AWS Bedrock deployments

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -477,20 +477,48 @@ To open the context menu:
 2. Create new chat
 3. User doesn't see model dropdown
 
-### Enterprise account
+### Cody Enterprise - Cody Gateway Instance
 
 #### Chat
 
-1. Login to Cody Enterprise account
-2. Create new chat
-3. User should see the default model in the dropdown but is unable to change it
+1. Login to the Cody Enterprise Gateway Instance
+2. Create a new chat
+3. User should not be able to change the LLM model on the chat window
+4. Ask Cody a question in the new chat window. 
 
 #### Commands
 
-1. Login to Cody Enterprise account
+1. Login to the Cody Enterprise Cody Gateway Instance
 2. Go to commands panel
 3. Trigger command
-4. Command should be executed with default model
+4. Command should be executed
+
+#### Completions
+
+1. Login to the Cody Enterprise Cody Gateway Instance
+2. Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go).
+ 
+
+### Cody Enterprise - AWS Bedrock Instance 
+
+#### Chat
+
+1. Login to the Cody Enterprise AWS Bedrock Instance
+2. Create a new chat
+3. User should not be able to change the LLM model on the chat window
+4. Ask Cody a question in the new chat window. 
+
+#### Commands
+
+1. Login to the Cody Enterprise AWS Bedrock Instance
+2. Go to commands panel
+3. Trigger command
+4. Command should be executed
+
+#### Completions
+
+1. Login to the Cody Enterprise AWS Bedrock Instance
+2. Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go).
 
 ## Other
 
@@ -532,14 +560,3 @@ Repeat the above starting from different themes.
 #### Expected behaviour
 
 Changing theme should lead to full repaint of the colours according to the current theme.
-
-
-### Claude 3 Enterprise Testing
-#### Cody Gateway Instance
-- [ ] Ask Cody a question in a new chat window. The question should include a request for Cody to generate code with the site-admin configuration is set to **Claude 3 Sonnet** for chatModel and **Claude 3 Haiku** for completionModel
-- [ ] Ask Cody a question in a new chat window. The question should include a request for Cody to generate code with the site-admin configuration is set to **Claude 3 Opus** for chatModel and **Claude 3 Haiku** for completionModel
-- [ ] Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go).
-
-#### AWS Bedrock Instance
-- [ ] Ask Cody a question in a new chat window. The question should include a request for Cody to generate code with the site-admin configuration is set to **Claude 3 Sonnet** for chatModel and **Claude 3 Haiku** for completionModel
-- [ ] Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go).
--- a/TESTING.md
+++ b/TESTING.md
@@ -484,7 +484,7 @@ To open the context menu:
 1. Login to the Cody Enterprise Gateway Instance
 2. Create a new chat
 3. User should not be able to change the LLM model on the chat window
-4. Ask Cody a question in the new chat window. 
+4. Ask Cody a question in the new chat window, Cody should respond. 
 
 #### Commands
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -35,6 +35,7 @@
     - [ ] [Automatic repository recognition](#automatic-repository-recognition)
     - [ ] [Persistent custom repository](#persistent-custom-repository)
     - [ ] [Colour scheme changes](#colour-scheme-changes)
+    - [ ] [Claude 3 Enterprise Testing](#claude-3-enterprise-testing)
 - Context
     - [ ] [PLG / SYMF / Local Keyword search](#local-keyword-search)
 - Model dropdown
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -531,3 +532,14 @@ Repeat the above starting from different themes.
 #### Expected behaviour
 
 Changing theme should lead to full repaint of the colours according to the current theme.
+
+
+### Claude 3 Enterprise Testing
+#### Cody Gateway Instance
+- [ ] Ask Cody a question in a new chat window. The question should include a request for Cody to generate code with the site-admin configuration is set to **Claude 3 Sonnet** for chatModel and **Claude 3 Haiku** for completionModel
```
pkukielka: I do not understand how tester can know/ensure this?
I'm talking about `with the site-admin configuration is set` part specifically.
User is not able to set this. So some specific sg instance should be used?
This need to be clear and actionable.

TESTING.md
```diff
@@ -531,3 +532,14 @@ Repeat the above starting from different themes.
...
+- [ ] Ask Cody a question in a new chat window. The question should include a request for Cody to generate code with the site-admin configuration is set to **Claude 3 Opus** for chatModel and **Claude 3 Haiku** for completionModel
+- [ ] Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go).
+
+#### AWS Bedrock Instance
```
pkukielka: Those steps are the same as **Cody Gateway Instance** testing steps.

TESTING.md
aramaraju: Is it better/easier if we just say 'Ask Cody a question in a new chat window with `Instance_name`?

dominiccooney: @aramaraju Just want to clarify, `Instance_name` is one we'd set up with chat and completions configured with those models? That sounds good to me... and we should put instructions on how to get credentials, the exact endpoint to log in to. This will keep the test turnaround time as fast as possible, increase confidence that we're covering what we want to test, and make the results useful because anyone can repro.

TESTING.md
dominiccooney: @pkukielka what's the best approach here? Should we be like "in each of these logins: ..., verify the following: ..."? Duplicate the steps? (We may want them to diverge soon enough anyway...)

TESTING.md
aramaraju: @dominiccooney Yes, that is my suggestion, specifically the exact endpoint. We can provide the instance as I'm inferring from Piotr's comment that the QA team doesn't actually configure changes on the instance. We can also find a way to provide the credentials for them in a safe place

dominiccooney: I'll reiterate, that sounds good to me.

kalanchan: I don't think we need to QA the part around different models, our QA testers should assume the environment we give them have the default configurations they need, unless we expand the scope to have them do configuration as well - which I don't recommend at the moment because it means we need to onboard them with elevated permissions to run certain workflows

This QA step is actually something a Sourcegraph team should own, as it touches configuration of the instance. As a client team, I think we should focus on the feature test cases instead. I think eventually we will want to expand the QA scope to include Sourcegraph configuration as well, but at the moment we can't.

TESTING.md
kalanchan: Ah this is where it gets tricky, so admins can choose either sonnet OR opus as their default model 🤔 We don't have a way for the QA team to switch configurations, SG02 is defaulted to sonnet at the moment

aramaraju: Simply created two subcases for Enterprise accounts, one for each instance provided. I have also removed the model validation as the team doesn't have the configuration, and will simply test against whatever configuration is on the instance

TESTING.md
```diff
@@ -476,20 +477,48 @@ To open the context menu:
...
 
 #### Chat
 
-1. Login to Cody Enterprise account
-2. Create new chat
-3. User should see the default model in the dropdown but is unable to change it
+1. Login to the Cody Enterprise Gateway Instance
+2. Create a new chat
+3. User should not be able to change the LLM model on the chat window
+4. Ask Cody a question in the new chat window.
```
dominiccooney: Let's elaborate this.
```suggestion
4. Ask Cody a question in the new chat window, Cody should respond. 
```

TESTING.md
```diff
@@ -476,20 +477,48 @@ To open the context menu:
...
+1. Login to the Cody Enterprise Cody Gateway Instance
 2. Go to commands panel
 3. Trigger command
-4. Command should be executed with default model
+4. Command should be executed
+
+#### Completions
+
+1. Login to the Cody Enterprise Cody Gateway Instance
+2. Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go).
```
dominiccooney, line 499:
```suggestion
2. Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go). A completion should appear.
```

@aramaraju

- Document the keyboard shortcut to press or the name of the command so the testers can look up the keyboard shortcut.
- This test should specify having a particular repo cloned and open, and mention specific files to open. It will make testing easier and make reproducing any bugs much easier.

@pkukielka does the tester's setup include other autocomplete providers? If so we need to ensure the completions are coming from Cody specifically somehow.

TESTING.md
```diff
@@ -476,20 +477,48 @@ To open the context menu:
...
 
-### Enterprise account
+### Cody Enterprise - Cody Gateway Instance
 
 #### Chat
 
-1. Login to Cody Enterprise account
-2. Create new chat
-3. User should see the default model in the dropdown but is unable to change it
+1. Login to the Cody Enterprise Gateway Instance
```
dominiccooney, line 484: Where's the credentials for this, if I wanted to perform these tests myself? Ditto below for "Cody Enterprise Cody Gateway Instance". Is that a *different* instance?

TESTING.md
```diff
@@ -476,20 +477,48 @@ To open the context menu:
...
+
+1. Login to the Cody Enterprise AWS Bedrock Instance
+2. Go to commands panel
+3. Trigger command
+4. Command should be executed
+
+#### Completions
+
+1. Login to the Cody Enterprise AWS Bedrock Instance
+2. Generate a completion using the manual-trigger key binding on any primary languages (Javascript, Typescript, TypescriptReact, Python, Go).
```
dominiccooney, line 521: This would be better if you removed the duplication and said something like: There are these kinds of instances (list them) and then had one set of instructions, like "For each kind of instance: ..."

TESTING.md
dominiccooney: @kalanchan 

> I don't think we need to QA the part around different models, our QA testers should assume the environment we give them have the default configurations they need, unless we expand the scope to have them do configuration as well

Everyone is in agreement that testers should not be having to edit admin policy.

Question for @kalanchan / @aramaraju ...  the updated diff now mentions different instances. Can we mention where the creds for these instances are in this file? Obviously the details of the creds won't be accessible to open source contributors, etc. But we need to drop a breadcrumb so teammates and contractors know where to get those creds, endpoint URLs, etc.

dominiccooney (changes requested):
TL;DR:

- Pointer to where to get creds.
- Remove duplication by instructing testers to do a set of steps for each kind of instance.
- More specificity in what those steps are.

TESTING.md
aramaraju: How are they getting the credentials today? I thought they were already shared outside of this doc. @kalanchan consolidated them into [1Password](https://start.1password.com/open/i?a=HEDEDSLHPBFGRBTKAKJWE23XX4&v=dnrhbauihkhjs5ag6vszsme45a&i=hd2jstpktlmzhodycq36phzo5e&h=team-sourcegraph.1password.com)

TESTING.md
aramaraju: One thing to note is that we are evaluating the LLM dropdown for Enterprise customers using Cody Gateway, which might not work as elegantly if combine the instance work. I agree combining them reads better, but I'm not sure if the team is going through a check list and having it split up by instances is easier in that case. If we do consolidate, shouldn't this also apply to the entire document? Do the following with these instances: Free, Pro, Enterprise 1, and 2? Will require a rewrite of the whole thing and I don't know why we didn't do that in the first place.

TESTING.md
kalanchan: "where can I get the credentials" kind of information should live in the handbook, but we can also add a link to 1password here. For the sake of completeness, let's just add it here 😄

TESTING.md
kalanchan: `opt \` is the manual trigger hotkey for autocomplete. 

The QA testers have experience with this, I don't think we need to assume they have another completion provider. The Cody logo also "spins/loads" when it triggers a completion, so theres indication from the UI that it's coming from Cody

PR #1291, opened 2024-04-06
Integration Test framework and an initial test for Document Code

steveyegge:
This PR introduces the new JB integration test framework, based on the built-in JB testing facilities.

# Overview

We only have one test so far, `DocumentCodeTest`, but:

- It tests end-to-end communication between JetBrains and the Cody Agent over the jsonrpc protocol
- It is fully asynchronous and can wait arbitrarily long for tests to complete
- It can be configured to hit a production LLM; currently there is a stub mock LLM interaction
- It has two separate validations:
  - one protocol roundtrip (`getFoldingRanges`)
  - one async client-side operation that awaits the appearance of the first codelens

This serves as a good reference example of using the test framework. 

Note that one facility of the test framework is that we use the built-in JetBrains pubsub support to signal to subscribers (the tests) that certain async checkpoints have been reached, which the tests can await.

`FixupSession` and some other classes are instrumented with these notifications, and the tests subscribe to them.

## Caveats

- There is only one test and it's incomplete
- I have not yet mocked out the LLM interaction for a full Document Code task/session
- Even after that, there are many tests left to write.

## Test plan

This is a test, so it's part of the test plan for previous PRs. 

I have tried not to break anything, so I've verified that on this branch, when using the `stevey/jetbrains-tests' branch of `sourcegraph/cody`:

- It's all still behind the `CODY_JETBRAINS_FEATURES` feature flag
- Cody works
- Integration test passes in both cli (`gradlew intTest`) and from within the IDE


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -97,10 +97,12 @@ private constructor(
         val conn = startAgentProcess()
         val client = CodyAgentClient()
         client.onSetConfigFeatures = project.service<CurrentConfigFeatures>()
+        logger.warn("Starting json-rpc Launcher")
```
pkukielka: Shouldn't all that logs be on the `info` level?

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -156,12 +159,27 @@ private constructor(
         processBuilder.environment()["CODY_LOG_EVENT_MODE"] = "connected-instance-only"
       }
 
+      if (ConfigUtil.isIntegrationTestModeEnabled()) {
+        // N.B. Do not set CODY_TESTING=true -- that is for Agent-side tests.
+        val testToken = System.getenv("CODY_INTEGRATION_TEST_TOKEN")
```
pkukielka: I think we should reuse tokens and polly.js recording infrastructure we already have in the agent.
This way we won't need to confugure/understand/setup two independent testing environments, and all token management could be taken care of on the agent side.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -47,24 +47,19 @@ class CodyAgentService(project: Project) : Disposable {
         FixupService.getInstance(project).getSessionForTask(task)?.update(task)
       }
 
-      agent.client.onEditTaskDidDelete = Consumer { task ->
```
pkukielka: Looks like you reverted my change to the callbacks/lifecycle managenet.
Was that intentional?
Initialising it in `FixupService` won't work in some cases, e.g. after agent restart.
If that was not intentional, please revert those changes.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -25,6 +27,71 @@ class FixupService(val project: Project) : Disposable {
   // The last text the user typed in without saving it, for continuity.
   private var lastPrompt: String = ""
 
+  init {
```
pkukielka: Please revert those changes or let's discuss offline.

CONTRIBUTING.md
```diff
@@ -381,3 +394,22 @@ see the configuration dropdown at the top.
...
 
+# Integration Testing
+
+Run the integration tests at the command line with:
+
+```
+./gradlew integrationTest
+```
+
+If you pass in an access token, it will use the default production LLM
```
dominiccooney, line 431: Let's make this more specific. "Pass in an access token" is confusing to me, because it looks like you use CODY_INTEGRATION_TEST_TOKEN to do that passing. But I don't think that is intended.

Where's the mock LLM? Does it care what token it is sent?

build.gradle.kts
```diff
@@ -401,4 +415,40 @@ tasks {
...
+    testClassesDirs = sourceSets["integrationTest"].output.classesDirs
+    classpath = sourceSets["integrationTest"].runtimeClasspath
+
+    include { it.file.hasParentNamed("integrationTest") }
+
+    useJUnit()
+
+    systemProperty("cody.integration.testing", "true")
+    systemProperty(
+        "idea.test.execution.policy", // For now, should be used by all tests
```
dominiccooney, line 542: Can you elaborate on the comment... What is this? Why? When would we not use it?

gradle.properties
```diff
@@ -18,6 +18,7 @@ platformPlugins=Git4Idea,PerforceDirectPlugin,java
 javaVersion=11
 # Gradle Releases -> https://github.com/gradle/gradle/releases
 gradleVersion=8.1.1
+kotlinVersion=1.9.22 // Also change this in build.gradle.kts plugins section
```
dominiccooney, line 21: noob question from me about Kotlin versioning... IntelliJ Platform also has some Kotlin pieces. What's the versioning constraints between our build and the IntelliJ hosts?

According to https://plugins.jetbrains.com/docs/intellij/using-kotlin.html#kotlin-standard-library , 2024.1 bundles 1.9.22. But I noticed "thing" that is our debug target, Main/Duke Icon, is 2022.something.

Does this mean we need to `kotlin.stdlib.default.dependency=true`? What does that mean if we're using IntelliJ's kotlin-isms (I was just looking at using their Kotlin Coroutine integration for the background task autocompleting repo names...)

src/integrationTest/kotlin/com/sourcegraph/cody/edit/DocumentCodeTest.kt
```diff
@@ -0,0 +1,152 @@
...
+
+class DocumentCodeTest : BasePlatformTestCase() {
+
+  override fun setUp() {
+    super.setUp()
+    configureFixture()
+  }
+
+  override fun tearDown() {
+    // TODO: Notify the Agent that all documents were closed.
```
dominiccooney: Very tempting for isolation to restart agent.

src/integrationTest/kotlin/com/sourcegraph/cody/edit/DocumentCodeTest.kt
```diff
@@ -0,0 +1,152 @@
...
+    super.tearDown()
+  }
+
+  fun testGetsWorkingGroupLens() {
+    // Do this before starting the edit operation, in order to be subscribed before
+    // the synchronous notification is sent, because it is not buffered or queued anywhere.
+    val foldingRangeFuture = subscribeToTopic(CodyInlineEditActionNotifier.TOPIC_FOLDING_RANGES)
+
+    val editor = myFixture.editor
+    assertFalse(editor.inlayModel.hasBlockElements())
```
dominiccooney: Hypothetical question, but is an assertion like this *after* a `runInEdtAndWait` likely to assert what you want in the JVM memory model (or at least do it in practice...)

src/integrationTest/kotlin/com/sourcegraph/cody/edit/DocumentCodeTest.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    // Execute the action and await the working group lens.
+    runInEdtAndWait { EditorTestUtil.executeAction(editor, "cody.documentCodeAction") }
+
+    val context = waitForTopic(CodyInlineEditActionNotifier.TOPIC_DISPLAY_WORKING_GROUP)
+    assertNotNull("Timed out waiting for working group lens", context)
+
+    // The inlay should be up.
+    assertTrue("Lens group inlay should be displayed", editor.inlayModel.hasBlockElements())
```
dominiccooney: ...oh I see my question about unsynchronized access was not theoretical. Is `runInEdtAndWait` a fence? But how could it be that if the EditorImpl's inlay model, block elements collection, etc. all unsynchronized, not-volatile stuff?

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
dominiccooney: Some test conditions are hard to recreate with production backends. VSCode uses both kinds to good effect.

dominiccooney (commented):
Some comments inline. Didn't make it through every last bit of this—dinner time.

@pkukielka would love to talk about how much agent restarting is a thing. On the one hand, I see all the code for it. On the other hand, I see state split between TypeScript and Kotlin and I'm skeptical that about restarting being very reliable. Keen to understand what I have missed.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
steveyegge: I added a few extra logging statements at key lifecycle stages of starting the Agent, because I found it tremendously helpful in narrowing down where it was failing when the Agent is pooping out. Which is often. I'll review them all, though, to see which ones we should really keep.

build.gradle.kts
```diff
@@ -401,4 +418,68 @@ tasks {
...
+
+    dependsOn("buildCody")
+  }
+
+  register<Test>("integrationTest") {
+    description = "Runs the integration tests."
+    applySharedConfiguration(sharedIntegrationTestConfig)
+  }
+
+  // Make sure to set CODY_INTEGRATION_TEST_TOKEN when using this task.
```
dominiccooney, line 553:
```
    onlyIf {
        System.env['CODY_INTEGRATION_TEST_TOKEN'] != null
    }
```

or let's conditionally print an error or something.

Ditto below.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
dominiccooney: @steveyegge pinging this comment ^

dominiccooney (changes requested):
Putting this back in your queue. The diff looks odd to me—there's a bunch of inline edits changes, and @pkukielka mentioned one revert? in here.

If the edits changes are particularly conflicty, we could also land everything except the edits changes + edits test and drop a trivial hello world type test in there so we can get hacking on other integration tests in parallel with the edit test landing. But if it is close then let's not toil over splitting it up.

dominiccooney:
This has conflicts with the actions refactoring in ff37c213728f45e2933b655b472e62bd95044583. Very tempting to split this up so we can write other tests while closing the loop on edit tests.

CONTRIBUTING.md
```diff
@@ -407,3 +420,22 @@ see the configuration dropdown at the top.
...
+rather than a mock LLM, which can be useful when updating the test if
+the protocol changes.
+
+```
+CODY_INTEGRATION_TEST_TOKEN=sgp_asdfasdfasdfasdfasdfasdfasdf
```
dominiccooney, line 436: There's a script in sourcegraph/cody, agent/scripts/export-cody-http-recording-tokens.sh . Can we standardize on using those tokens? Everyone has access to them and there are different variants for rate limited tokens, etc. if we need them down the road.

build.gradle.kts
```diff
@@ -426,4 +444,132 @@ tasks {
...
+          // For now, should be used by all tests
+          "idea.test.execution.policy" to "com.sourcegraph.cody.test.NonEdtIdeaTestExecutionPolicy",
+          "test.resources.dir" to project.file("src/integrationTest/resources").absolutePath)
+
+  val integrationTestEnvVars =
+      mapOf(
+          "CODY_JETBRAINS_FEATURES" to "cody.feature.inline-edits=true",
+          "CODY_RECORDING_MODE" to "replay",
+          "CODY_RECORDING_DIRECTORY" to "recordings",
+          "CODY_RECORD_IF_MISSING" to "false") // Polly needs this to record at all.
```
dominiccooney, line 491: ...weird!

build.gradle.kts
```diff
@@ -426,4 +444,132 @@ tasks {
...
+
+  // Common configuration for integration tests.
+  // TODO: This doesn't actually work.
```
dominiccooney, line 494: Can you elaborate?

dominiccooney (approved):
It will be so great to get these going, can't wait to write some.

A few minor comments inline.

This needs rebasing to pick up the changes where 'edit' and 'workspace edit' produce a boolean.

PR #1304, opened 2024-04-10
Reorganised the entity hierarchy of HistoryService.

### LINKED ISSUE #1210: bug: Switching accounts does not switch chats, context repos
### Cody Version

Main b33ee8da1a58527560ea9b05a5976e11421ccfd8

### IDE Information

IntelliJ IDEA 2022.1 (Community Edition)
Build #IC-221.5080.210, built on April 12, 2022
Runtime version: 11.0.14.1+1-b2043.25 aarch64
macOS 14.2.1
Memory: 752M
Cores: 12
Non-Bundled Plugins:
    com.sourcegraph.jetbrains (6.0-localbuild)

Kotlin: 221-1.6.20-release-285-IJ5080.210

### Describe the bug

In VSCode, the chats and chat history are related to the signed-in account.

In JetBrains, when I switch accounts, I'm still looking at the same chat. Repos selected from one instance still appear in the context list, but those repos won't exist/may have different IDs/etc. on another instance.

Loom demonstrating: https://www.loom.com/share/02a4773c4ee84782966696e9e36e5928?sid=8bb32b22-2b1b-4fae-b919-4ee1cb2baba2

### Expected behavior

When I switch accounts, I should get chat history and chosen repos related to the signed-in account.

### Additional context

VSCode also makes it much easier to see what account you're signed in to, and whether you're on the pro plan. This is the first thing you see when you click "Account" in VSCode:

Screenshot 2024-03-28 at 15 16 55

odisseus:
Fixes https://github.com/sourcegraph/jetbrains/issues/1210

The chats are now grouped by the account ID, and each account has (optionally) its own enhanced context.

## Test plan
 Added a new test

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
@@ -16,7 +16,7 @@ import com.sourcegraph.cody.CodyToolWindowFactory
 import com.sourcegraph.cody.api.SourcegraphApiRequestExecutor
 import com.sourcegraph.cody.api.SourcegraphApiRequests
 import com.sourcegraph.cody.history.HistoryService
-import com.sourcegraph.cody.history.state.AccountHistoryState
+import com.sourcegraph.cody.history.state.AccountData
 import com.sourcegraph.cody.history.state.ChatState
 import com.sourcegraph.cody.history.state.EnhancedContextState
 import com.sourcegraph.cody.history.state.LLMState
@@ -447,10 +447,8 @@ class SettingsMigration : Activity {
         val accountId = oldState.accountId
         if (accountId == null) return@forEach
         val newState =
-            historyState.accountHistories.find { it.accountId == accountId }
-                ?: (AccountHistoryState.create(accountId).also {
-                  historyState.accountHistories += it
-                })
+            historyState.accountData.find { it.accountId == accountId }
+                ?: (AccountData.create(accountId).also { historyState.accountData += it })
         newState.defaultLlm = historyState.defaultLlm
         newState.defaultEnhancedContext = historyState.defaultEnhancedContext
         if (newState.chats.find { it.internalId == oldState.internalId } == null) {
--- a/src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
@@ -16,6 +16,7 @@ import com.sourcegraph.cody.CodyToolWindowFactory
 import com.sourcegraph.cody.api.SourcegraphApiRequestExecutor
 import com.sourcegraph.cody.api.SourcegraphApiRequests
 import com.sourcegraph.cody.history.HistoryService
+import com.sourcegraph.cody.history.state.AccountHistoryState
 import com.sourcegraph.cody.history.state.ChatState
 import com.sourcegraph.cody.history.state.EnhancedContextState
 import com.sourcegraph.cody.history.state.LLMState
@@ -43,6 +44,9 @@ class SettingsMigration : Activity {
     RunOnceUtil.runOnceForProject(project, "CodyHistoryLlmMigration") { migrateLlms(project) }
     RunOnceUtil.runOnceForProject(project, "CodyConvertUrlToCodebaseName") {
       migrateUrlsToCodebaseNames(project)
+    }
+    RunOnceUtil.runOnceForProject(project, "CodyAccountHistoryMigration") {
+      organiseChatsByAccount(project)
     }
     RunOnceUtil.runOnceForApp("CodyApplicationSettingsMigration") { migrateApplicationSettings() }
     RunOnceUtil.runOnceForApp("ToggleCodyToolWindowAfterMigration") {
@@ -434,6 +438,25 @@ class SettingsMigration : Activity {
               .distinctBy { it.codebaseName }
               .toMutableList()
       enhancedContextState.remoteRepositories = remoteRepositories
+    }
+
+    fun organiseChatsByAccount(project: Project) {
+      val historyService = HistoryService.getInstance(project)
+      val historyState = historyService.state
+      historyState.chats.forEach { oldState ->
+        val accountId = oldState.accountId
+        if (accountId == null) return@forEach
+        val newState =
+            historyState.accountHistories.find { it.accountId == accountId }
+                ?: (AccountHistoryState.create(accountId).also {
+                  historyState.accountHistories += it
+                })
+        newState.defaultLlm = historyState.defaultLlm
+        newState.defaultEnhancedContext = historyState.defaultEnhancedContext
+        if (newState.chats.find { it.internalId == oldState.internalId } == null) {
+          newState.chats.add(oldState)
+        }
+      }
     }
   }
 }
--- a/src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
@@ -80,6 +80,10 @@ constructor(protected val project: Project, protected val chatSession: ChatSessi
 
   /** Gets the chat session's enhanced context state. */
   protected fun getContextState(): EnhancedContextState? {
+    if (CodyAuthenticationManager.getInstance(project).getActiveAccount() == null) {
+      // There is no active account, so there is no enhanced context either
+      return null
+    }
     val historyService = HistoryService.getInstance(project)
     return historyService.getContextReadOnly(chatSession.getInternalId())
         ?: historyService.getDefaultContextReadOnly()
--- a/src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
@@ -9,7 +9,7 @@ import com.intellij.openapi.project.Project
 import com.sourcegraph.cody.agent.protocol.ChatMessage
 import com.sourcegraph.cody.agent.protocol.Speaker
 import com.sourcegraph.cody.config.CodyAuthenticationManager
-import com.sourcegraph.cody.history.state.AccountHistoryState
+import com.sourcegraph.cody.history.state.AccountData
 import com.sourcegraph.cody.history.state.ChatState
 import com.sourcegraph.cody.history.state.EnhancedContextState
 import com.sourcegraph.cody.history.state.HistoryState
@@ -56,34 +56,31 @@ class HistoryService(private val project: Project) :
     if (contextState != null) {
       val found = getOrCreateChat(internalId)
       found.enhancedContext = EnhancedContextState()
-      found.enhancedContext?.copyFrom(contextState)
+      found.enhancedContext!!.copyFrom(contextState)
     }
   }
 
   @Synchronized
   fun updateDefaultContextState(contextState: EnhancedContextState) {
-    getOrCreateActiveAccountHistory().defaultEnhancedContext = EnhancedContextState()
-    getOrCreateActiveAccountHistory().defaultEnhancedContext?.copyFrom(contextState)
+    getOrCreateActiveAccountEntry().defaultEnhancedContext = EnhancedContextState()
+    getOrCreateActiveAccountEntry().defaultEnhancedContext!!.copyFrom(contextState)
   }
 
   @Synchronized
   fun getContextReadOnly(internalId: String): EnhancedContextState? {
     return copyEnhancedContextState(
-        getOrCreateActiveAccountHistory()
-            .chats
-            .find { it.internalId == internalId }
-            ?.enhancedContext)
+        getOrCreateActiveAccountEntry().chats.find { it.internalId == internalId }?.enhancedContext)
   }
 
   @Synchronized
   fun getDefaultContextReadOnly(): EnhancedContextState? {
-    return copyEnhancedContextState(getOrCreateActiveAccountHistory().defaultEnhancedContext)
+    return copyEnhancedContextState(getOrCreateActiveAccountEntry().defaultEnhancedContext)
   }
 
   @Synchronized
   fun getDefaultLlm(): LLMState? {
     val account = CodyAuthenticationManager.getInstance(project).getActiveAccount()
-    val llm = account?.let { findHistory(it.id) }?.defaultLlm
+    val llm = account?.let { findEntry(it.id) }?.defaultLlm
     if (llm == null) return null
     return LLMState().also { it.copyFrom(llm) }
   }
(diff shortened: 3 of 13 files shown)


src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
```diff
@@ -5,10 +5,8 @@ import com.intellij.util.xmlb.annotations.OptionTag
 
 class HistoryState : BaseState() {
 
-  @get:OptionTag(tag = "chats", nameAttribute = "") var chats: MutableList<ChatState> by list()
+  @get:OptionTag(tag = "accountHistories", nameAttribute = "")
+  var accountHistories: MutableList<AccountHistoryState> by list()
 
   @get:OptionTag(tag = "defaultLlm", nameAttribute = "") var defaultLlm: LLMState? by property()
```
pkukielka: Default LLM should also be per account, so we need to move it to `AccountHistoryState`

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -122,12 +136,16 @@ class HistoryService(private val project: Project) :
...
-    val newChat = ChatState.create(activeAccountId, internalId)
-    state.chats += newChat
-    return newChat
+    val accountHistory = getOrCreateActiveAccountHistory()
+    val found = accountHistory.chats.find { it.internalId == internalId }
+    return found ?: ChatState.create(internalId).also { accountHistory.chats += it }
+  }
+
+  private fun getOrCreateActiveAccountHistory(): AccountHistoryState {
+    val activeAccountId = CodyAuthenticationManager.getInstance(project).getActiveAccount()!!.id
```
pkukielka: Please try to never use `!!`.
Either `getActiveAccount` should not return `null` (and then we should change it) or `null` is a valid state (or marker for error) and then we should handle it properly.

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -99,32 +99,30 @@ class SettingsMigration : Activity {
   }
 
   private fun migrateLlms(project: Project) {
```
pkukielka: I think that won't work. Past migrations run only once.
If you need to migrate something now you need to add completely new function.
Also this way migrations are linear, changing past migrations is a bit more risky as we can accidentally mess up something in the chain of changes.

So:
1) Please move your migrations to separate methods called after all others
2) Add some test for it

src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
odisseus: Done

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
odisseus: Fixed

src/main/kotlin/com/sourcegraph/cody/history/state/AccountHistoryState.kt
```diff
@@ -0,0 +1,24 @@
+package com.sourcegraph.cody.history.state
+
+import com.intellij.openapi.components.BaseState
+import com.intellij.util.xmlb.annotations.OptionTag
+
+class AccountHistoryState : BaseState() {
```
mkondratek: nice 🚀

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -111,8 +113,10 @@ class EnhancedContextPanel(private val project: Project, private val chatSession
   private fun getContextState(): EnhancedContextState? {
     val historyService = HistoryService.getInstance(project)
 
-    return historyService.getContextReadOnly(chatSession.getInternalId())
-        ?: historyService.getDefaultContextReadOnly()
+    return CodyAuthenticationManager.getInstance(project).getActiveAccount()?.let { _ ->
```
mkondratek: why do we need `CodyAuthenticationManager.getInstance(project).getActiveAccount()`?

mkondratek: oh, this is probably just a null check. if it is so please rewrite it as a regular if-statement to avoid the confusion in the future

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -40,8 +40,10 @@ class EnhancedContextPanel(private val project: Project, private val chatSession
   private val enhancedContextNode =
       ContextTreeRootNode(CodyBundle.getString("context-panel.tree.node-chat-context")) { isChecked
         ->
-        isEnhancedContextEnabled.set(isChecked)
-        updateContextState { it.isEnabled = isChecked }
+        if (CodyAuthenticationManager.getInstance(project).getActiveAccount() != null) {
```
mkondratek: and can you please put some comment here? the code in the if-statement body seems to be unrelated to the if statement condition

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -54,6 +58,29 @@ class SettingsMigration : Activity {
...
+              })
+      newState.defaultLlm = historyState.defaultLlm
+      newState.defaultEnhancedContext = historyState.defaultEnhancedContext
+      if (newState.chats.find { it.internalId == oldState.internalId } == null) {
+        newState.chats.add(oldState)
+      }
+    }
+
+    // TODO is this required in this case?
+    CodyToolWindowContent.executeOnInstanceIfNotDisposed(project) { refreshChatHistoryPanel() }
```
mkondratek: We do not need it here. `CodyToolWindowContent` and `refreshChatHistoryPanel` refer the UI.

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -9,11 +9,7 @@ import com.intellij.openapi.project.Project
 import com.sourcegraph.cody.agent.protocol.ChatMessage
 import com.sourcegraph.cody.agent.protocol.Speaker
 import com.sourcegraph.cody.config.CodyAuthenticationManager
-import com.sourcegraph.cody.history.state.ChatState
-import com.sourcegraph.cody.history.state.EnhancedContextState
-import com.sourcegraph.cody.history.state.HistoryState
-import com.sourcegraph.cody.history.state.LLMState
-import com.sourcegraph.cody.history.state.MessageState
+import com.sourcegraph.cody.history.state.*
```
mkondratek: Personally I do not like wildcard imports (and [here](https://stackoverflow.com/a/147461) is why). 

You can change your IDE settings (and [here](https://stackoverflow.com/a/3348855) is how).

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -61,45 +57,51 @@ class HistoryService(private val project: Project) :
 
   @Synchronized
   fun updateDefaultContextState(contextState: EnhancedContextState) {
-    state.defaultEnhancedContext = EnhancedContextState()
-    state.defaultEnhancedContext?.copyFrom(contextState)
+    getOrCreateActiveAccountHistory().defaultEnhancedContext = EnhancedContextState()
```
mkondratek: This PR is big enough already but we should consider renaming "History" to "State". It's a bit strange that `updateDefaultContextState` modifies the "history".

src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
```diff
@@ -5,10 +5,18 @@ import com.intellij.util.xmlb.annotations.OptionTag
 
 class HistoryState : BaseState() {
 
-  @get:OptionTag(tag = "chats", nameAttribute = "") var chats: MutableList<ChatState> by list()
+  @get:OptionTag(tag = "accountHistories", nameAttribute = "")
+  var accountHistories: MutableList<AccountHistoryState> by list()
```
mkondratek: Before we introduce more strange naming... Are we sure about `accountHistories`? IMHO the word "history" is redundant and misleading. And has been introduced probably in the context of Chat History panel.

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -61,45 +57,51 @@ class HistoryService(private val project: Project) :
...
   @Synchronized
-  fun getDefaultLlm(): LLMState? {
-    val defaultLlm = LLMState()
-    defaultLlm.copyFrom(state.defaultLlm ?: return null)
-    return defaultLlm
-  }
+  fun getDefaultLlm(): LLMState? =
+      CodyAuthenticationManager.getInstance(project).getActiveAccount()?.let { account ->
+        findHistory(account.id)?.defaultLlm?.let { llm -> LLMState().also { it.copyFrom(llm) } }
+      }
```
mkondratek: a lot of nesting here 😕 can you make it more readable?

mkondratek (dismissed):

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -122,12 +124,26 @@ class HistoryService(private val project: Project) :
...
   private fun getOrCreateChat(internalId: String): ChatState {
-    val found = state.chats.find { it.internalId == internalId }
-    if (found != null) return found
-    val activeAccountId = CodyAuthenticationManager.getInstance(project).getActiveAccount()?.id
-    val newChat = ChatState.create(activeAccountId, internalId)
-    state.chats += newChat
-    return newChat
+    val accountHistory = getOrCreateActiveAccountHistory()
+    val found = accountHistory.chats.find { it.internalId == internalId }
+    return found ?: ChatState.create(internalId).also { accountHistory.chats += it }
```
pkukielka: Why do we need to use `ChatState.create` in this method?

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -54,6 +58,29 @@ class SettingsMigration : Activity {
     }
   }
 
+  private fun migrateHistoryToActiveAccount(project: Project) {
```
pkukielka: Please add test for that method.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
odisseus: Done

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
odisseus: Removed

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
odisseus: Done

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
odisseus: To be honest, I don't understand why this method wasn't written as a constructor. But given that it already exists, I don't see any problem with using it either.

**P.S.** This seems to be some kind of a pattern, but I don't understand its purpose. I'm going to try refactoring all such functions into constructors.

**P.P.S.** #1325

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
odisseus: Done

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
odisseus: Does this structure really contain the whole state associated with an account? If not, the new name wouldn't be a good fit either.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
odisseus: Done

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
mkondratek: true ☝️

odisseus: Done

src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
odisseus: How about a separate PR to change the existing names? It may be worth doing that _before_ merging this PR.

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
odisseus: Done

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
```diff
@@ -44,6 +45,9 @@ class SettingsMigration : Activity {
     RunOnceUtil.runOnceForProject(project, "CodyConvertUrlToCodebaseName") {
       migrateUrlsToCodebaseNames(project)
     }
+    RunOnceUtil.runOnceForProject(project, "CodyAccountHistoryMigration") {
```
pkukielka, line 48: Shouldn't that migration be run as last one?

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -80,6 +80,10 @@ constructor(protected val project: Project, protected val chatSession: ChatSessi
 
   /** Gets the chat session's enhanced context state. */
   protected fun getContextState(): EnhancedContextState? {
+    if (CodyAuthenticationManager.getInstance(project).getActiveAccount() == null) {
```
pkukielka, line 83: Why is that addition needed? Is that a real scenario which happen during your testing?
One could argue that even if there is no active account there is always default context, so that comment is not really true.

src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
pkukielka: We cannot freely change that names, because they are used to seriallize the history.
When they are out and someone started to use it they is written in the stone.

Although, I think it should be possible to use some mapping, possibly with that `nameAttribute` property from above.

pkukielka: I would vote for just name `AccountData`.
`AccountState` sounds more like info about account and in fact, we have class which is called `AccountState` already,

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/history/state/HistoryState.kt
odisseus: Done

src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
odisseus: Renamed to `getOrCreateActiveAccountEntry`.

src/main/kotlin/com/sourcegraph/cody/config/SettingsMigration.kt
odisseus: This order seems to be good.

PR #1307, opened 2024-04-10
Inline Edit: Add Show Diff

### LINKED ISSUE #1108: Inline Edits: Show Diff
### Installation Info:

IntelliJ IDEA 2022.1 (Community Edition)
Build #IC-221.5080.210, built on April 11, 2022
Runtime version: 11.0.14.1+1-b2043.25 aarch64
macOS 14.2.1
Memory: 752M
Cores: 10
Non-Bundled Plugins:
    com.sourcegraph.jetbrains (6.0-localbuild)

Kotlin: 221-1.6.20-release-285-IJ5080.210

### Description:

The VS Code extension shows a Code Lens, "Show Diff", in the "Accept/Retry/Undo/Diff" lens group after inserting a doc string or code edit sequence. This brings up a native diff viewer widget showing (just) the changes made by the AI for this particular fixup task.

We need this in JetBrains too. The Code Lens is already implemented. The diff widget or functionality is not documented in the JetBrains Plugin Development Guide, so it will take a little digging to figure out how to do it. I estimate it at a few days of work.

This issue is a placeholder for filling out all the actual detail when we do the Inline Edits PRD.

### Design Tasks:

```[tasklist]
### Design Tasks
- [ ] Add a draft title or issue reference here
```

### Engineering Tasks:

```[tasklist]
### Engineering Tasks
- [ ] Implement the show-diff widget
- [ ] Notifications/telemetry
```

### Related Docs:

- [PRD: Your-Title]()

mkondratek:
Resolves https://github.com/sourcegraph/jetbrains/issues/1108. 

## Test plan
1. Run Inline Edit
2. Click Show Diff 

## Demo

https://github.com/sourcegraph/jetbrains/assets/19799111/5bccf425-781b-43e2-adc0-d6b306079ff2


src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
```diff
@@ -46,6 +47,10 @@ abstract class FixupSession(
     val project: Project,
     val document: Document
 ) : Disposable {
+  protected val documentBefore: Document =
```
dominiccooney: This isn't how Cody VSCode's diff view works. Diffing off this will include all your other edits you make in the file while Cody is working.

Cody VSCode [(source)](https://sourcegraph.com/github.com/sourcegraph/cody@c69aaf7/-/blob/vscode/src/non-stop/codelenses/provider.ts?L288-295) takes the file when you click "diff", re-inserts the original text we sent to the LLM in the range we're tracking for the replacement, and diffs that.

dominiccooney (changes requested):

src/main/kotlin/com/sourcegraph/cody/edit/FixupSession.kt
mkondratek: @dominiccooney, I did it the same way now. can you please re-review?

src/main/kotlin/com/sourcegraph/cody/edit/EditShowDiffAction.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  override fun isAvailable(e: AnActionEvent): Boolean {
+    e.dataContext.getData(DOCUMENT_BEFORE_DATA_KEY) ?: return false
+    e.dataContext.getData(SELECTION_RANGE_DATA_KEY) ?: return false
+    return true
+  }
+
+  override fun getDiffRequestChain(e: AnActionEvent): DiffRequestChain {
+    val project = e.project
+    val selectionRange = e.dataContext.getData(SELECTION_RANGE_DATA_KEY)!!
```
dominiccooney: I don't think this is right. You should write a more detailed test plan, which includes editing (increasing and decreasting the line and character count) before, in, (and after, but that will probably work) the range:

- When the fixup is started, but before it is applied
- After the fixup is applied

And then showing the diff.

FixupTask on the extension side has a "live" range which tracks edits. But it appears that the JetBrains side is caching this range.

dominiccooney (changes requested):
I think this is still not correct, some feedback inline.

src/main/kotlin/com/sourcegraph/cody/edit/EditShowDiffAction.kt
mkondratek: I think it's working as expected now. There is a problem with changes when the fixup is started but not yet finished ("Cody is working..."). I does not work in that case but that case does not work for us at all yet.

src/main/kotlin/com/sourcegraph/cody/edit/BaseFixupSession.kt
```diff
@@ -0,0 +1,28 @@
+package com.sourcegraph.cody.edit
+
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.editor.Document
+import com.intellij.openapi.editor.RangeMarker
+
+open class BaseFixupSession(val document: Document) {
```
pkukielka: Maybe different name if it's just for passing around markers?

src/main/kotlin/com/sourcegraph/cody/edit/DiffSession.kt
```diff
@@ -0,0 +1,49 @@
+package com.sourcegraph.cody.edit
+
+import com.intellij.openapi.command.WriteCommandAction
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.editor.Document
+import com.intellij.openapi.project.Project
+
+class DiffSession(
```
pkukielka: That looks fine, but I'm wondering if we cannot get the same for free from the agent.
CC @dominiccooney

dominiccooney: I've got mixed feelings about this.

When it works, oh, it is perfection. Amazing.

On the other hand, a couple of times I got a very scrambled edit. Which makes me think leaning on the extension's range tracking will be better.

But, this has the actions and invoking the diff viewer, which we need in either case, so let's roll forward with this. We can adapt the source of the diff document later if it comes to that.

dominiccooney (approved):
Approving to unblock.

Let's write down a manual test plan so when the testers get this they can really hammer on it. Including editing behind, in, and after the region Cody is editing.

PR #1325, opened 2024-04-16
Refactored some static factory methods into secondary constructors.

odisseus:
If the factory method just accepts a different list of arguments, it should be a constructor.

## Test plan

No change in functionality – no new tests needed?


src/main/kotlin/com/sourcegraph/cody/history/state/AccountData.kt
```diff
@@ -3,7 +3,10 @@ package com.sourcegraph.cody.history.state
 import com.intellij.openapi.components.BaseState
 import com.intellij.util.xmlb.annotations.OptionTag
 
-class AccountData : BaseState() {
+class AccountData() : BaseState() {
+  constructor(accountId: String) : this() {
```
mkondratek, line 7: wow, I was not aware we can do this in Kotlin

mkondratek (approved):
lgtm

PR #1326, opened 2024-04-16
Rework listeners and notification code

### LINKED ISSUE #1321: bug: IntelliJ plugin sends too many file open requests to the agent
### Cody Version

All

### IDE Information

All

### Describe the bug

IntelliJ send `didOpen` notifications to the agent on every editor change, like selection or cursor change.
That is incorrect and causes some problems, like reparsing of too many files. by tree-sitter. 

### Expected behavior

We need to use proper callbacks

### Additional context

_No response_

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/1321

## Changes

* Streamline and cleanup listeners/notifications code
* Make sure we do not send `didOpen` notifications when it is not necessary 
 
## Test plan

Full manual QA for completions and edit actions will be necessary.


src/main/kotlin/com/sourcegraph/cody/initialization/PostStartupActivity.kt
```diff
@@ -27,7 +32,19 @@ class PostStartupActivity : StartupActivity.DumbAware {
...
+    val multicaster = EditorFactory.getInstance().eventMulticaster
+    if (multicaster is EditorEventMulticasterEx) {
+      try {
+        val disposable = CodyAgentService.getInstance(project)
+        multicaster.addFocusChangeListener(CodyFocusChangeListener(), disposable)
+        multicaster.addCaretListener(CodyCaretListener(), disposable)
+        multicaster.addSelectionListener(CodySelectionListener(), disposable)
+        multicaster.addDocumentListener(CodyDocumentListener(project), disposable)
+      } catch (e: Exception) {
+        // Ignore exception https://github.com/sourcegraph/sourcegraph/issues/56032
```
dominiccooney: This looks dodgy:

- Any one of these could throw, and you will neglect to process the rest. So if one throws, will they *all* throw? If not, let's wrap each one individually so failing on line 41 doesn't skip 42-44. The change linked to that issue only wraps one `add*Listener` so it doesn't have this problem.
- This catches *all* exceptions. The linked issue doesn't describe what's going on... it notes a cause but not the exception wrapping it. Is there any way we can filter the exception and re-throw ones unrelated to sourcegraph/sourcegraph#56032 ? We could walk the cause chain to look for that Throwable, etc. and rethrow the rest.
- Let's move as much code as we can out from the exception swallower... `CodyAgentService.getInstance(...)`, instantiating the listeners if adding them is failing, or alternatively adding the listeners if instantiating them is failing.

src/main/kotlin/com/sourcegraph/cody/listeners/CodyCaretListener.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    val editor = e.editor
+    val project = editor.project
+    val textDocument = ProtocolTextDocument.fromEditor(editor)
+    if (project != null && textDocument != null) {
+      CodyAgentService.withAgent(project) { agent: CodyAgent ->
+        agent.server.textDocumentDidFocus(textDocument)
+      }
+    }
```
dominiccooney: This is fine, I wonder if we want `CodyAgentService.maybeWithEditorDocument(editor) { ...` that takes `agent` and `textDocument`?

src/main/kotlin/com/sourcegraph/cody/listeners/CodyCaretListener.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    CodyAutocompleteManager.instance.clearAutocompleteSuggestions(e.editor)
+    CodyAutocompleteManager.instance.triggerAutocomplete(
```
dominiccooney, line 33: Surprised that IntelliJ platform doesn't handle this. Isn't there[ a "show suggestions as you type" setting](https://www.jetbrains.com/help/idea/auto-completing-code.html) where you can turn off automatic autocomplete and rely on manual triggers? Does this call honor that setting?

dominiccooney (approved):
Looks good, please minimize the scope of the exception swallowing wiring up event handlers as much as practical.

src/main/kotlin/com/sourcegraph/cody/listeners/CodyCaretListener.kt
pkukielka: That is Cody completion, which is completely separate than IntelliJ suggestions and have different UX as well (e.g. can have multiline suggestions). It have also it's own setting to enable/disable it. I do not think we want to mix those two.

Or do I misunderstood your question?

dominiccooney: > That is Cody completion, which is completely separate than IntelliJ suggestions

Super! Thanks for clarifying.

src/main/kotlin/com/sourcegraph/cody/listeners/CodyCaretListener.kt
pkukielka: I see where you are coming from as I do not like that boilerplate as well.
But I do not like to pollute  `CodyAgentService` too much with various utility functions either.
Let me try to rework this a bit!

src/main/kotlin/com/sourcegraph/cody/initialization/PostStartupActivity.kt
pkukielka: Good call on flagging this.
I think we should not hide any errors there, as any of those listeners failing to be attached will handicap Cody, and we want to know if that happens.
I think previous catch could be due to (long gone) issues with initialization order as I see it was called from the `CodyAgent` itself. In `plugin.xml` there was commented out code where we tried to attach focus listener through configuration, so maybe at some point we had both ways in code and they were conflicting.

I'm going to remove that try/catch completely and give it a solid testing round.

@olafurpg by any chance, do you maybe remember if that error was happening often or only from time to time?

olafurpg: I don't remember the details of how frequently this happened. Feel free to remove, but wrapping each individually also sounds good combined with logging to help troubleshoot.

src/main/kotlin/com/sourcegraph/cody/listeners/CodyCaretListener.kt
pkukielka: It should look better now :)

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -15,32 +14,36 @@ private constructor(
...
+          selectionStartPosition != selectionEndPosition) {
         return Range(
             Position(selectionStartPosition.line, selectionStartPosition.column),
             Position(selectionEndPosition.line, selectionEndPosition.column))
       }
-      val caret = editor.caretModel.primaryCaret
-      val position = Position(caret.logicalPosition.line, caret.logicalPosition.column)
-      // A single-offset caret is a selection where end == start.
-      return Range(position, position)
+      return null
```
pkukielka: That is actually pretty big change.
I noticed that passing empty range to simulate caret is on the agent side interpreted as a empty selection.
And e.g. in case of empty selection on empty lines, it seems to reliably cause skipping given file from the context at all (e.g. when using chat to ask the question about the current file).
So that change should improve context understanding for all our actions.

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
pkukielka: I reverted it as it was causing problems with chat context responses, but I think this change address most corner cases:
https://github.com/sourcegraph/cody/pull/3833

PR #1339, opened 2024-04-18
Display onboarding tooltip on first autocomplete suggestion

### LINKED ISSUE #1319: [User Testing] Add Gotit tooltips
User Testing results:

https://docs.google.com/document/d/1opkHzO3ZZUHs591ajjy6HYDkouFEI1O7kJzdqVYhEfQ/edit

In order to fix some of the issues found, we should include ["Got it tooltips"](https://jetbrains.design/intellij/controls/got_it_tooltip/) for:

- [ ] On the editor: First time user gets an Autocomplete

`To accept a Cody autocomplete, hit TAB`

- [ ] On context area: First time user opens the sidebar (Enterprise Only)

`Set the context for this discussion by selecting up to 10 relevant repositories indexed on [instance URL]. This helps Cody provide more tailored answers.`

olafurpg:
Fixes #1319.

Previously, users didn't get any onboarding instructions when using Cody for the first time in JetBrains. For example, many users didn't know they should press TAB to accept an autocomplete suggestion, or option-bracket to cycle through options. This PR fixes the problem by using the built-in `GotItTooltip` component, which is designed for this use-case.

## Test plan

Manually tested. 
CleanShot 2024-04-18 at 16 01 48@2x

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
@@ -417,7 +417,7 @@ class CodyAutocompleteManager {
                           KeymapUtil.getShortcutText("cody.cycleBackAutocompleteAction")),
                   inlay /* dispose tooltip alongside inlay */)
               .withHeader(CodyBundle.getString("gotit.autocomplete.header"))
-              .withPosition(Balloon.Position.atLeft)
+              .withPosition(Balloon.Position.above)
               .withIcon(Icons.CodyLogo)
               .andShowCloseShortcut()
       try {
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
Splendid.

IntelliJ handles the dismissed state management of these? Neat.

olafurpg:
I'm investigating a seeming regression where autocomplete inlays now appear one line below the cursor. I'm not able to reproduce on main so I think it's caused by the tooltip shifting stuff.

CleanShot 2024-04-18 at 16 07 44@2x

olafurpg:
@dominiccooney Yes! IntelliJ manages the state. All existing users will get the tooltip when updating, which isn't ideal, but it might be helpful for them to learn about the cycle shortcuts. I think the code will become very tricky to maintain if we try to avoid showing the tooltip for existing users.

pkukielka:
Wouldn't it be nice to also mention command to force autocomplete?

olafurpg:
cc/ @sourcegraph/jetbrains-plugin have you seen this behavior before for autocomplete? The inlay appears in an empty line below the cursor

CleanShot 2024-04-18 at 16 14 02@2x

I commented out `gotit.show()` and I'm still able to reproduce the issue so I'm going to assume it's not a regression in this PR. Worth keeping a close eye on this

pkukielka (approved):
LGTM

olafurpg:
@danielmarquespt merging this PR with the current copy. Feel free to propose a separate copy, or send a PR changing the text. 

I think we should also document the trigger shortcut in the onboarding tooltip, but it's not done in this PR>

Issue #1252, opened 2024-04-01
[Cody Ignore] Hook BaseCommandAction/doAction to consult Cody Ignore


dominiccooney:
src/main/kotlin/com/sourcegraph/cody/chat/actions/BaseCommandAction.kt needs to consult Cody Ignore rules (in sourcegraph/cody#3641) and display an error message when commands are disabled due to Cody Ignore.

@toolmantim 's [input on the RFC:](https://docs.google.com/document/d/10tD8n-5py4XWQD9c7doJB7zIuxa7lwiOkpo2d5NLlnA/edit#bookmark=id.2dzfwrd11qso)

> Running a command on an ignored file
>
> * Ideal change: notification error message on command execution saying the repo is disabled for Cody

We should add BaseCommandAction fields/methods so we can name the action that has been disabled.

Per the P2 PRD requirement for commands, inform people that their Sourcegraph admin can set the Cody Ignore policy (in sourcegraph/docs#222)

```[tasklist]
### Tasks
- [ ] Telemetry
- [X] BaseCommandAction consults Cody Ignore Policy
- [X] Notification UX
```

dominiccooney:
Lacks Telemetry, but I will file a follow-up issue for all Cody Ignore Telemetry.

PR #1340, opened 2024-04-18
Cody Ignore: Add an action to override the Cody Ignore policy for testing

dominiccooney:
This adds some plumbing for Cody Ignore, and a way to interactively test that locally in IntelliJ without admin access to a Sourcegraph instance.

- Adds Agent protocol stubs for Cody Ignore protocol. These depend on Agent-side changes in sourcegraph/cody#3858
- Adds an "internals" menu to the status bar, hidden behind a feature flag, for doing ad-hoc manual ignore policy settings

Part of #1252

## Test plan

- Run agent with sourcegraph/cody#3858
- Run the plugin with `CODY_JETBRAINS_FEATURES=cody.feature.internals-menu=true` and open a clone of the sourcegraph/sourcegraph repo
- On the status bar, Internals, Testing: Cody Ignore
- Check *Override policy for testing*
- Enter URI regex `\.go$` . Any repo regex is fine; that setting is currently unused on the extension side. Click OK.
- Verify that console output includes `handling ignore rules changing not yet implemented`. This indicates the testing override generated a notification that the rules have changed, through to the stub in the extension.
- Start a chat
- In the Enhanced Context Selector, ensure *Chat Context* is checked. Uncheck any specific repos.
- Open a Go file, for example `cmd/batcheshelper/main.go` and highlight some code
- In chat, ask Cody "what does this do?" There should be *no local context* and Cody should make something up unrelated to the highlighted code.
- On the status bar, Internals, Testing: Cody Ignore. Uncheck *Override policy for testing*. OK. Again verify console output.
- In chat, hit up and resubmit the same chat message. There should be local context from the open Go file and Cody should comment related to that code.
- Run the plugin again *without* `CODY_JETBRAINS_FEATURES=cody.feature.internals-menu=true` and verify that the Internals menu does not appear on the status bar.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Ignore.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Ignore.kt
@@ -1,8 +1,8 @@
 package com.sourcegraph.cody.agent.protocol
 
-data class IgnoreForUriParams(val uri: String)
+data class IgnoreTestParams(val uri: String)
 
-data class IgnoreForUriResponse(
+data class IgnoreTestResponse(
     val policy: String // "use" or "ignore"
 )
 
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Ignore.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Ignore.kt
@@ -1,0 +1,9 @@
+package com.sourcegraph.cody.agent.protocol
+
+data class IgnoreForUriParams(val uri: String)
+
+data class IgnoreForUriResponse(
+    val policy: String // "use" or "ignore"
+)
+
+data class TestingIgnoreOverridePolicy(val uriRe: String, val repoRe: String)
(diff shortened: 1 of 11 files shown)


pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Ignore.kt
```diff
@@ -0,0 +1,9 @@
+package com.sourcegraph.cody.agent.protocol
+
+data class IgnoreForUriParams(val uri: String)
+
+data class IgnoreForUriResponse(
+    val policy: String // "use" or "ignore"
```
odisseus, line 6: Why not enum?

dominiccooney: We tend to just enumerate the string values, but I will look into it in a follow-up. There's no JetBrains side calls to this yet.

PR #1348, opened 2024-04-19
add edit and document test plan

kalanchan:
Update QA test plan with Edit and Document command instructions. Missing features at the moment are show diff, and undo. I added the instructions anyways, we will unhide them after they are implemented

## Test plan
N/A

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -226,9 +226,9 @@ automatically.
 1. Run the `Edit Code` command with instructions for how Cody should edit the selected code.
 1. Verify that Cody automatically applies edits to the selected code, per the user’s instructions.
 1. Verify that there is a notification that "Cody is working..." while edits are being applied.
-1. Verify that you can see a list of code lenses with a Cody icon above the generated code: `Show diff`, `Accept`, `Retry`, and `Undo`.
-<!-- 1. Verify that you can see a diff view of the edit in a new tab by clicking `Show diff`. -->
-1. Verify that you can prompt Cody to retry the command by clicking `Retry` and entering new instructions.
+1. Verify that you can see a list of code lenses with a Cody icon above the generated code, with shortcuts: `Accept`, `Edit & Retry`, `Undo` and `Show Diff`.
+1. Verify that you can see a diff view of the edit in a new tab by clicking `Show Diff`.
+1. Verify that you can prompt Cody to retry the command by clicking `Edit & Retry` and entering new instructions.
 1. Verify that you can undo the edit by clicking `Undo`.
 1. Verify that the ghost text disappears by clicking `Accept`.
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -219,7 +219,7 @@ automatically.
 ### Edit
 
 #### Editing code
-<!-- unhide after full edit command is implemented -->
+<!-- TODO: unhide after full edit command is implemented -->
 <!-- - Verify that the option to run the `Edit Code` command is available from the sidebar, right-click menu, the command palette (option+c, /edit), or Option+K keyboard shortcut. -->
 1. Verify that the option to run the `Edit Code` command is available from the right-click menu.
 1. Highlight a section of code.
--- a/TESTING.md
+++ b/TESTING.md
@@ -14,7 +14,8 @@
     - [ ] [Explain Selected Code](#explain-code)
     - [ ] [Generate Test](#generate-test)
     - [ ] [Smell Code](#smell-code)
-    - [ ] Edit
+    - [ ] [Edit](#edit)
+    - [ ] [Document](#document)
 - Chat
     - [ ] [Autoscroll to latest message](#autoscroll-to-latest-message)
     - [ ] [Read chat history without interruptions](#read-chat-history-without-interruptions)
--- a/TESTING.md
+++ b/TESTING.md
@@ -14,6 +14,7 @@
     - [ ] [Explain Selected Code](#explain-code)
     - [ ] [Generate Test](#generate-test)
     - [ ] [Smell Code](#smell-code)
+    - [ ] Edit
 - Chat
     - [ ] [Autoscroll to latest message](#autoscroll-to-latest-message)
     - [ ] [Read chat history without interruptions](#read-chat-history-without-interruptions)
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
 * Chat responds with **potential issues** and **suggestions** like missing nullability checks or input sanitization (
   response may vary).
 
+### Edit
+
+#### Editing code
+<!-- unhide after full edit command is implemented -->
```
dominiccooney:
```suggestion
<!-- TODO: unhide after full edit command is implemented -->
```

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+<!-- - Verify that the option to run the `Edit Code` command is available from the sidebar, right-click menu, the command palette (option+c, /edit), or Option+K keyboard shortcut. -->
+1. Verify that the option to run the `Edit Code` command is available from the right-click menu.
+1. Highlight a section of code.
+1. Run the `Edit Code` command with instructions for how Cody should edit the selected code.
+1. Verify that Cody automatically applies edits to the selected code, per the user’s instructions.
+1. Verify that there is a notification that "Cody is working..." while edits are being applied.
+1. Verify that you can see a list of code lenses with a Cody icon above the generated code: `Show diff`, `Accept`, `Retry`, and `Undo`.
```
dominiccooney: Let's make these labels match the case and order in the product
```suggestion
1. Verify that you can see a list of code lenses with a Cody icon above the generated code, with shortcuts: `Accept`, `Edit & Retry`, `Undo` and `Show Diff`.
```

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+<!-- 1. Verify that you can see a diff view of the edit in a new tab by clicking `Show diff`. -->
```
dominiccooney: This landed in 618acfd601ab8d70b3457344ffb3f254130247cf
```suggestion
1. Verify that you can see a diff view of the edit in a new tab by clicking `Show Diff`.
```

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+1. Verify that you can prompt Cody to retry the command by clicking `Retry` and entering new instructions.
```
dominiccooney:
```suggestion
1. Verify that you can prompt Cody to retry the command by clicking `Edit & Retry` and entering new instructions.
```

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+1. Verify that you can undo the edit by clicking `Undo`.
```
dominiccooney, line 232: This needs to be have way more detail. For Edit & Retry, Undo and Show Diff we need to test:

- Edit the file by adding multiple lines of text before and after the region Cody is editing, while Cody is working and after it has completed, then use (Edit & Retry/Undo/Show Diff)
- Edit the file by *removing* multiple lines of text before and after the region Cody is editing, while Cody is working and after it has completed, then use (Edit & Retry/Undo/Show Diff)
- Edit the file by *removing* multiple lines of text before and *adding multiple lines of text after* the region Cody is editing, while Cody is working and after it has completed, then use (Edit & Retry/Undo/Show Diff)
- When testing Undo specifically, only the changes Cody made should be undone.
- When testing Show Diff specifically, only the changes in the region Cody edited should be highlighted in the diff.

You need a preamble before ALL of these commands which says "when we say CONCURRENT HUMAN EDIT, we mean: no editing, add lines of text before the region Cody is editing, ..." and then in the test plan, you need to say "For each kind of CONCURRENT HUMAN EDIT: ..." and then have the list of instructions with one level of indent.

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+1. Verify that the ghost text disappears by clicking `Accept`.
```
dominiccooney, line 233: VSCode has a behavior where editing within the region Cody edited auto-commits it. We should test this, too.

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+
+#### Expected behaviour
```
dominiccooney, line 235: This expected behavior is weak, it doesn't really describe what those buttons should *do*, which are peppered in the instructions. If you want to have "expected behavior" separate, let's break up the above test by each flow: Edit thru Accept; Edit &rarr; Undo; Edit &rarr; Show Diff; etc.

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+1. Highlight a section of code.
+1. Run the `Edit Code` command with instructions for how Cody should edit the selected code.
+1. Verify that Cody automatically applies edits to the selected code, per the user’s instructions.
+1. Verify that there is a notification that "Cody is working..." while edits are being applied.
+1. Verify that you can see a list of code lenses with a Cody icon above the generated code: `Show diff`, `Accept`, `Retry`, and `Undo`.
+<!-- 1. Verify that you can see a diff view of the edit in a new tab by clicking `Show diff`. -->
+1. Verify that you can prompt Cody to retry the command by clicking `Retry` and entering new instructions.
+1. Verify that you can undo the edit by clicking `Undo`.
+1. Verify that the ghost text disappears by clicking `Accept`.
+
```
dominiccooney, line 234: You haven't mentioned the keyboard shortcuts:
- That they work.
- That the text on the labels change if you change the key bindings.

Is there a separate "key bindings" test section for that? Are we not testing the keyboard shortcuts? In general, Cody uses a bunch of shortcuts which coincide with cedillas/umlauts type stuff, we should test them. VSCode and JetBrains may handle shortcuts on those kinds of keystrokes differently.

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+### Document
+
+<!-- unhide after full document command is implemented -->
+
+1. Verify that the option to run the `Document Code` command is available from the right-click menu.
+1. Highlight a section of code in a file.
+1. Run the `Document Code` command.
+1. Verify that an icon appears above the highlighted code with "Cody is working..." while Cody generates documentation.
+<!-- 1. Verify that you can see a diff view of the generated documentation in a new tab by clicking `Show diff`. -->
```
dominiccooney, line 261: Ditto above, show diff has landed

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+- [ ] Try to add a symbol to the Edit instruction, by using "@#" and searching for a symbol
+- [ ] Try to change the range of the Edit, by selecting "Range". Check that navigating through the options correctly updates the range shown in the editor.
+- [ ] Try to change the model of the Edit, by selecting "Model".
+- [ ] Submit edits after changing the above values, and check that the Edit performs correctly (e.g. uses correct range, uses correct context)
+- [ ] Verify that the selected configuration is shown in the input after clicking "Retry" on a completed edit.
+- [ ] Verify that you can still change the configuration shown in the input after clicking "Retry" on a completed edit. -->
+
+### Document
+
+<!-- unhide after full document command is implemented -->
```
dominiccooney, line 255: The following are not hidden, so remove the comment. It is unclear what you're referring to unhiding here.

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+
+1. Verify that the option to run the `Document Code` command is available from the right-click menu.
+1. Highlight a section of code in a file.
```
dominiccooney, line 258: Let's specify particular file(s) and function(s). Preferably ones which:

- Cover a language where we do have a folding range implementation in TreeSitter and one where we don't, so TypeScript and Ocaml say.
- Functions which are "below the fold"—not visible in the first screenful of code.
- One time, a scroll position where the top of the function is not visible, one time when it is.

TESTING.md
```diff
@@ -214,6 +216,60 @@ automatically.
...
+1. Run the `Document Code` command.
+1. Verify that an icon appears above the highlighted code with "Cody is working..." while Cody generates documentation.
+<!-- 1. Verify that you can see a diff view of the generated documentation in a new tab by clicking `Show diff`. -->
+1. Verify that you can prompt Cody to retry the command by clicking `Retry` and entering new instructions.
+1. Verify that you can undo the documentation by clicking `Undo`.
+1. Verify that the ghost text disappears by clicking `Accept`.
+1. Move your cursor inside a function in the file without highlighting code, before running the `Document Code` command again.
+1. Verify that Cody adds documentation above the function that contains your cursor.
+
```
dominiccooney, line 267: Question whether we want to do this test for each kind of "concurrent human edit." My gut feeling is that it is OK. The code is mostly shared, and we can just pound on concurrent human edits for Edit and we'll find the relevant bugs in Document.

Let's have the test plan for Document cover editing the function body while it is editing. That is a realistic use case that's a bit more rigorous than no editing, without the full rigor of all the kinds of edits.

dominiccooney (changes requested):
Thanks for these, some feedback inline.

Issue #1234, opened 2024-03-28
bug: java.util.concurrent.CompletionException: java.lang.IndexOutOfBoundsException: Wrong line: 800. Available lines count: 229


taylorsperry:
### Cody Version

5.5.1-nightly

### IDE Information

IntelliJ IDEA 2023.3.3 (Ultimate Edition)
Build #IU-233.14015.106, built on January 24, 2024
Licensed to Sourcegraph / Taylor Sperry
Subscription is active until May 12, 2024.
Runtime version: 17.0.9+7-b1087.11 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.6.4
GC: G1 Young Generation, G1 Old Generation
Memory: 4096M
Cores: 10
Metal Rendering is ON
Registry:
  ide.experimental.ui=true
Non-Bundled Plugins:
  com.intellij.ml.llm (233.14015.147)
  com.sourcegraph.jetbrains (5.5.1-nightly)
Kotlin: 233.14015.106-IJ

### Describe the bug

* I highlighted some code and ran an edit command
* Then I highlighted a different selection of code and ran edit again
* Cody hung on "Cody is working" and I saw this error

### Expected behavior

Cody should apply the edit

### Additional context

Stacktrace:

```text
java.util.concurrent.CompletionException: java.lang.IndexOutOfBoundsException: Wrong line: 800. Available lines count: 229
 at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
 at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
 at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:708)
 at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
 at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
 at com.sourcegraph.cody.agent.CodyAgentClient.lambda$acceptOnEventThread$0(CodyAgentClient.java:86)
 at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:209)
 at com.intellij.openapi.application.TransactionGuardImpl.access$100(TransactionGuardImpl.java:22)
[...]

PR #1365, opened 2024-04-22
Better error message for failed workspace edit operations

odisseus:
The logs would look like this:
```
INFO - FixupSession - Applying edits to a file: file://~/devel/cody-intellij/src/main/kotlin/com/sourcegraph/cody/auth/PersistentActiveAccountHolder.kt
SEVERE - org.eclipse.lsp4j.jsonrpc.RemoteEndpoint - Internal error: com.sourcegraph.cody.edit.sessions.EditException: Edit failed: TextEdit(type=replace, range=Range(start=Position(line=40, character=63), end=Position(line=40, character=64)), position=null, value=) at RangeMarker(1295,1296) 1714}
java.util.concurrent.CompletionException: com.sourcegraph.cody.edit.sessions.EditException: Edit failed: TextEdit(type=replace, range=Range(start=Position(line=40, character=63), end=Position(line=40, character=64)), position=null, value=) at RangeMarker(1295,1296) 1714}
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
...
```

This should make it easier to debug #1234.

## Test plan

No new functionality, no new tests needed

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
@@ -287,6 +287,7 @@ abstract class FixupSession(
           if (op.edits == null) {
             logger.warn("Workspace edit operation has no edits")
           } else {
+            logger.info("Applying edits to a file: ${op.uri}")
             performInlineEdits(op.edits)
           }
         }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -273,26 +273,30 @@ abstract class FixupSession(
...
           }
+          else ->
+              logger.warn(
+                  "DocumentCommand session received unknown workspace edit operation: ${op.type}")
         }
-        else ->
-            logger.warn(
-                "DocumentCommand session received unknown workspace edit operation: ${op.type}")
+      } catch (e: RuntimeException) {
+        throw WorkspaceEditException(op, e)
```
pkukielka: If we hope it will help us debug e.g. issue with incorrect ranges it would be much more useful to do those try/caches at the level of single operations.
Than you have much more rich and useful info to log:

```kt
data class TextEdit(
    // This tag will be 'replace', 'insert', or 'delete'.
    val type: String,

    // Valid for replace & delete.
    val range: Range? = null,

    // Valid for insert.
    val position: Position? = null,

    // Valid for replace & insert.
    val value: String? = null
)
```

I think we should also log the he open document url and length.

odisseus: Refactored.

mkondratek (approved):
lgtm

PR #1369, opened 2024-04-23
Onboarding tooltip: don't hide the code

olafurpg:
Previously, there were cases when the position of the onboarding tooltip would hide the underlying code. This goes against JB recommendations https://jetbrains.design/intellij/controls/got_it_tooltip/#location even if we are using the built-in position helpers. This PR changes the position dynamically based on

- Whether the project window is open
- Whether the cursor is close to the top

Depending on the different values, we now either display the tooltip to the left, above, or below.

## Test plan

Manually tested

Project view closed and at the top of the file

CleanShot 2024-04-23 at 11 26 57@2x

Project view is open at the top

CleanShot 2024-04-23 at 11 40 19@2x

Project view is closed but there is space above

CleanShot 2024-04-23 at 11 40 55@2x


dominiccooney (approved):
Looks great!

PR #1372, opened 2024-04-23
Overhaul of the Edit Code dialog, hotkeys, and Edit/Retry

steveyegge:
This PR introduces a rewrite of the **Edit Code with Cody** dialog to match Daniel's designs. The dialog has changed into a JFrame to give it maximum flexibility and customizability, though at the cost of having to write a lot of code to do things you get for free in dialogs.

Overview of the PR:

- All the theming has been redone to match the designs
- The input text field is now multi-line and scrollable
- The dialog is movable and resizable (although resizing is currently janky)
- Dialog handles theme changes correctly (though now you can't see because it closes on loss of focus)
- Dialog opens at the line where you invoked it
- All code lenses are registered as actions and have working keyboard shortcuts
- All keyboard shortcuts are rendered properly rather than being hardcoded
- The dialog's history handling has been cleaned up

I've made a start at re-theming the code lenses, but it has another day or two of work. I'll add it to this PR.

## Test plan

I created a test plan for the instructions dialog in `TESTING.md`. It is pretty comprehensive and has dozens of tests and assertions.

We _really_ need UI/Robot tests for the UI, which is already starting to become complicated.


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -112,7 +112,7 @@ private constructor(
                           ClientCapabilities(
                               edit = "enabled", editWorkspace = "enabled", codeLenses = "enabled")))
               .thenApply { info ->
-                logger.info("Connected to Cody agent " + info.name)
+                logger.warn("Connected to Cody agent " + info.name)
```
pkukielka, line 118: Nitpick: It does not sound like warning to me :)

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -161,7 +161,10 @@ private constructor(
               .redirectErrorStream(false)
               .redirectError(ProcessBuilder.Redirect.PIPE)
               .start()
-      process.onExit().thenAccept { token.abort() }
+      process.onExit().thenAccept {
+        logger.warn("Cody agent process exited with code " + it.exitValue())
```
pkukielka: Nitpick: Do we want to print that for 0 code as well? Then I think it just add noise to the log and it's not really a warning but an expected behaviour.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -36,6 +36,7 @@ class ChatPanel(
           ModelUsage.CHAT,
           project,
           onSetSelectedItem = ::setLlmForAgentSession,
+          null,
```
pkukielka: Could you please make it named parameter?

src/main/kotlin/com/sourcegraph/cody/edit/HistoryManager.kt
```diff
@@ -0,0 +1,37 @@
+package com.sourcegraph.cody.edit
+
+/** Manages user prompt history in memory for a session (project lifetime). */
+class HistoryManager<T>(private val capacity: Int) {
```
pkukielka, line 4: At some point we could maybe unify this and `CodyChatMessageHistory` component

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensWidgetGroup.kt
```diff
@@ -99,15 +110,26 @@ class LensWidgetGroup(val session: FixupSession, parentComponent: Editor) :
...
-  fun show(range: Range) {
-    commandCallbacks = session.commandCallbacks()
+  fun show(range: Range): CompletableFuture<Boolean> {
     val offset = range.start.toOffset(editor.document)
-    ApplicationManager.getApplication().invokeLater {
-      if (!isDisposed.get()) {
-        inlay = editor.inlayModel.addBlockElement(offset, false, true, 0, this)
-        Disposer.register(this, inlay!!)
+    val future = CompletableFuture<Boolean>()
+    onEventThread {
```
pkukielka: That `val future = CompletableFuture<Boolean>()` seems unnecessary.
`onEventThread` already returns `CompletableFuture` which you could use, no?

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensWidgetGroup.kt
```diff
@@ -240,7 +290,25 @@ class LensWidgetGroup(val session: FixupSession, parentComponent: Editor) :
     }
   }
 
+  private fun <T> onEventThread(handler: Supplier<T>): @NotNull CompletableFuture<T> {
```
pkukielka, line 288: Maybe it would make sense to put this in some utility class?
It seems to be generally useful.

pkukielka (approved):
I added a few minor comments but it looks good!

src/main/kotlin/com/sourcegraph/cody/edit/HistoryManager.kt
steveyegge: Maybe! We'll have to see how similar they are in practice. But I do think we should add persistence to the history for Inline Edits, so that may be a good solution.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
steveyegge: Sure, I'lll fix that.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
steveyegge: If your Cody agent isn't starting up, I _promise_ you it's useful information. I will change it to info as soon as we get UI that tells you the Agent status.

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensWidgetGroup.kt
steveyegge: Ah good catch -- that was leftover from some refactoring.

Issue #1253, opened 2024-04-01
[Cody Ignore] Add a banner to editors with Cody Ignored files open


dominiccooney:
Typical in JetBrains to display a warning [Banner](https://jetbrains.design/intellij/controls/banner/) on files which "informs the user about the state of a specific context in the project. ... to suggest actions that can impact the user’s workflow."

VSCode uses the status bar for the same purpose, see vscode/src/services/StatusBar.ts

[API docs entrypoint.](https://plugins.jetbrains.com/docs/intellij/notifications.html#editor-banner)

dominiccooney:
@danielmarquespt I welcome your input here.

My thinking here is that:
- The Cody Ignore PRD wants to communicate *why* a file is ignored, command disabled, etc. This is a good principle, it will cut down on people cutting and pasting from files their administrator intended to "protect."
- The communication happens with an error message. This is good. We could disable/take away a menu item, but that's confusing because there's no explanation why. The menu item label isn't the place to explain why a command is disabled—it should be short—and in some places it is hard to make those labels dynamic anyway. And you may be using a shortcut anyway and never *see* the menu item.
- But it is annoying to invoke a command, only to get an error message.
- Thus we should use a banner, so you know what's going on, and the meaning of the banner is reinforced if you do go into the commands through shortcut muscle memory or whatever.

danielmarquespt:
The UX for this is a bit different now (maybe simpler?)

It is covered on this discussion: https://sourcegraph.slack.com/archives/C05AGQYD528/p1712864303221559

And the figma file lives here: https://www.figma.com/file/m9Yvgv6W3aPRLfXnxgwnEz/JetBrains?type=design&node-id=1181-17006&mode=design

PR #1379, opened 2024-04-24
Add a service for checking Cody Ignore policy and decorate some UI with ignored status

dominiccooney:
This change is integrated through Agent, but the extension side is not integrated with server-set policy yet. Instead, this change uses a testing hook which pushes an ad-hoc policy to the extension—see the test plan.

As a result, this does not cause any user visible change for Enterprise users yet. For self-serve users of the experimental file-based Cody Ignore feature, this UX may trigger until extension-side integration is complete. However we have determined that to be very few users.

Exact UX is still being designed. These labels and icons are placeholders.

Part of #1252, #1253.

## Test plan

Tested manually:

1. Run the extension with `CODY_JETBRAINS_FEATURES=cody.feature.internals-menu=true` and open some files, say `Foo.java` and `Bar.go`
2. Status bar, Internals, Testing: Cody Ignore
3. Check "Override policy for testing", set URI regex to `\.java$`. OK.
4. Verify that a banner appears in `Foo.java` but not `Bar.go`.
5. Run a command like Explain Code with the right click context menu and verify that a notification appears
6. Run a command like Explain Code with a keyboard shortcut like ctrl-shift-E and verify that a notification appears
7. Verify when focus switches from `Foo.java` to `Bar.go`, a slashed Cody icon appears in the status bar when an ignored file is focused.


src/main/kotlin/com/sourcegraph/cody/ignore/IgnoreOracle.kt
```diff
@@ -0,0 +1,88 @@
...
+  IGNORE("ignore"),
+  USE("use"),
+}
+
+/**
+ * Provides details about whether files and repositories must be ignored as chat context, for
+ * autocomplete, etc. per policy.
+ */
+@Service(Service.Level.PROJECT)
+class IgnoreOracle(private val project: Project) {
```
odisseus, line 24: Hmph.
rc24-duke-java-mascot

src/main/kotlin/com/sourcegraph/cody/ignore/IgnoreOracle.kt
```diff
@@ -0,0 +1,88 @@
...
+    if (result != null) {
+      completable.complete(result)
+      return completable
+    }
+    CodyAgentService.withAgent(project) { agent ->
+      val policy =
+          when (agent.server.ignoreTest(IgnoreTestParams(uri)).get().policy) {
+            "ignore" -> IgnorePolicy.IGNORE
+            "use" -> IgnorePolicy.USE
+            else -> throw Exception("invalid ignore policy value")
```
odisseus: Consider using `IllegalStateException`, or any other class that extends `RuntimeException`. This might be a moot point for Kotlin, but in Java the possibility to throw any other type of Throwable affects the method's signature.

src/main/kotlin/com/sourcegraph/cody/chat/actions/BaseCommandAction.kt
```diff
@@ -1,17 +1,34 @@
...
+import com.sourcegraph.cody.ignore.IgnorePolicy
 
 abstract class BaseCommandAction : BaseChatAction() {
 
   abstract val myCommandId: CommandId
 
   override fun doAction(project: Project) {
-    FileEditorManager.getInstance(project).selectedTextEditor?.let {
+    FileEditorManager.getInstance(project).selectedTextEditor?.let { editor ->
+      ApplicationManager.getApplication().assertIsDispatchThread()
```
pkukielka: Actually `selectedTextEditor` should be called on EDT only as well, so you could move that assertion to the first line of the method.

pkukielka (approved):
LGTM

PR #1384, opened 2024-04-24
Add generate test command v.2

pkukielka:
## Changes

**Functional:**

* Added new command for automatic generation of unit tests in the editor

**Technical:**

* Added new test command in both context menu and Commands tab
* Implemented `textDocument/show` agent notification
* Refactored undoable commands to so contain their whole responsibility (e.g. creating and maintaining markers)
* Added `showDocument` method for navigation to a given file/range
* Fixed undo action for commands with multiple edits
* Ensured that diff works also for text v2 command
* 
## Test plan

Full manual QA of other commands and diff feature


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
```diff
@@ -1,6 +1,33 @@
 package com.sourcegraph.cody.agent
 
-import com.sourcegraph.cody.agent.protocol.*
```
odisseus, line 3: In this particular case, the star import might be justified.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
odisseus, line 1: What is the conceptual difference between the `Position` and `Range` from this package as opposed to those from `com.sourcegraph.cody.vscode`?

I'm not requesting changes within this PR, but if the only difference is a set of additional methods, perhaps we could use the Pimp My Library pattern instead.

src/main/kotlin/com/sourcegraph/cody/edit/fixupActions/ReplaceUndoableAction.kt
```diff
@@ -0,0 +1,72 @@
...
+
+  private constructor(
+      other: ReplaceUndoableAction,
+      document: Document
+  ) : this(other.project, other.edit, document) {
+    this.beforeMarker =
+        if (other.beforeMarker == null) null
+        else
+            document.createRangeMarker(
+                other.beforeMarker!!.startOffset, other.beforeMarker!!.endOffset)
```
odisseus:
```suggestion
        other.beforeMarker?.let {document.createRangeMarker(it.startOffset, it.endOffset)}
```

src/main/kotlin/com/sourcegraph/cody/edit/fixupActions/ReplaceUndoableAction.kt
```diff
@@ -0,0 +1,72 @@
...
+        else
+            document.createRangeMarker(
+                other.afterMarker!!.startOffset, other.afterMarker!!.endOffset)
+    this.originalText = other.originalText
+  }
+
+  override fun apply() {
+    val (start, end) =
+        if (beforeMarker != null) Pair(beforeMarker!!.startOffset, beforeMarker!!.endOffset)
+        else Pair(0, document.textLength)
```
odisseus:
```suggestion
        beforeMarker?.let{Pair(it.startOffset, it.endOffset)} ?: Pair(0, document.textLength)
```

src/main/resources/CodyBundle.properties
```diff
@@ -116,27 +116,22 @@ duration.x-months-ago={0} months ago
 duration.last-year=Last year
 duration.x-years-ago={0} years ago
 duration.x-ago={0} ago
-
```
odisseus: Why?

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
pkukielka: Discussed offline, let's go back to this tomorrow on standup.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/Range.kt
pkukielka: I think idea was to decouple protocol classes of the classes we are using for the business logic.
If that makes sense or not is an different question...

src/main/resources/CodyBundle.properties
pkukielka: My editor settings, fixed.

src/main/kotlin/com/sourcegraph/cody/edit/exception/EditCreationException.kt
```diff
@@ -0,0 +1,7 @@
+package com.sourcegraph.cody.edit.exception
+
+import com.sourcegraph.cody.agent.protocol.TextEdit
+
+class EditCreationException(val edit: TextEdit, cause: Throwable) : RuntimeException(cause) {
+  override val message: String = "Edit creation failed: $edit}"
```
odisseus: Looks like an extra closing brace.
```suggestion
  override val message: String = "Edit creation failed: $edit"
```

src/main/kotlin/com/sourcegraph/cody/edit/exception/EditExecutionException.kt
```diff
@@ -0,0 +1,8 @@
+package com.sourcegraph.cody.edit.exception
+
+import com.sourcegraph.cody.edit.fixupActions.FixupUndoableAction
+
+class EditExecutionException(action: FixupUndoableAction, cause: Throwable) :
+    RuntimeException(cause) {
+  override val message: String = "Edit application failed for $action"
```
odisseus: I don't like this error message. It may suggest that there is an application (a program) which has just failed, or that we have failed to edit said application.

I would put it like this: `"Failed to apply the edit for $action"`, or perhaps "Could not apply the edit to ..." etc.

src/main/kotlin/com/sourcegraph/cody/edit/fixupActions/InsertUndoableAction.kt
```diff
@@ -0,0 +1,68 @@
...
+    this.beforeMarker =
+        if (other.beforeMarker == null) null
+        else
+            document.createRangeMarker(
+                other.beforeMarker!!.startOffset, other.beforeMarker!!.endOffset)
+    this.afterMarker =
+        if (other.afterMarker == null) null
+        else
+            document.createRangeMarker(
+                other.afterMarker!!.startOffset, other.afterMarker!!.endOffset)
```
odisseus: Consider refactoring

src/main/kotlin/com/sourcegraph/cody/edit/fixupActions/ReplaceUndoableAction.kt
```diff
@@ -0,0 +1,75 @@
...
+    document.replaceString(start, end, replacementText)
+    afterMarker =
+        if (beforeMarker == null) null
+        else document.createRangeMarker(start, start + replacementText.length)
+  }
+
+  override fun undo() {
+    val (start, end) =
+        if (afterMarker == null) Pair(0, document.textLength)
+        else Pair(afterMarker!!.startOffset, afterMarker!!.endOffset)
```
odisseus: This can be refactored in the same way as the above method.

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -306,76 +330,67 @@ abstract class FixupSession(
...
-      "insert" -> {
-        val position = edit.position ?: return null
-        startOffset = document.getLineStartOffset(position.line) + position.character
-        endOffset = startOffset
-      }
-      else -> return null
-    }
-    return createMarker(startOffset, endOffset)
+  private fun adjustToDocumentRange(r: Range): Range {
+    val start = if (r.start.line < 0) Position(line = 0, character = r.start.character) else r.start
```
odisseus, line 367: In what situations is a negative `r.start.line` valid?

odisseus (changes requested):

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
pkukielka: Good question, I added comment in the code.

odisseus (approved):

mkondratek (approved):

PR #1399, opened 2024-04-26
Cody Ignore: Do not block the main thread on ignore checks

dominiccooney:
...actually, block it, but only for up to 16ms at a time.

This appears to have looping potential if retrieving a policy takes >16ms, eventually unblocks, and triggers a banner update which could restart policy retrieval. But because the policies are cached as the slow path unblocks, and the read side synchronously reads the cache, this should guarantee forward progress provided the policy is not changing.

This appears to have blocking potential if there's dozens of files visible, each blocking for up to 16ms. However `EditorNotificationsImpl` runs each `collectNotificationData` as a non-blocking task in the non-urgent executor. Delays may accumulate, but IntelliJ will prevent them from blocking interaction.

## Test plan

Tested manually starting the product, opening files, etc.:
- Dialing the 16ms timeout down to zero; throwing a TimeoutException in that `try`.
- Inducing a 5s delay on the agent side by adding `await new Promise(resolve => setTimeout(resolve, 5000))` to `ignore/test`.

Tested commands manually.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ignore/IgnoreOracle.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ignore/IgnoreOracle.kt
@@ -10,6 +10,8 @@ import com.sourcegraph.cody.agent.CodyAgentService
 import com.sourcegraph.cody.agent.protocol.IgnoreTestParams
 import com.sourcegraph.cody.statusbar.CodyStatusService
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
 
 enum class IgnorePolicy(val value: String) {
   IGNORE("ignore"),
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/ignore/IgnoreOracle.kt
```diff
@@ -85,4 +85,18 @@ class IgnoreOracle(private val project: Project) {
...
   }
+
+  /**
+   * Gets whether `uri` should be ignored for autocomplete, etc. If the result is not available
+   * quickly, returns null and invokes `orElse` on a pooled thread when the result is available.
+   */
+  fun policyForUriOrElse(uri: String, orElse: (policy: IgnorePolicy) -> Unit): IgnorePolicy? {
+    val completable = policyForUri(uri)
+    try {
+      return completable.get(16, TimeUnit.MILLISECONDS)
```
pkukielka, line 96: So in fact we want to either get it from cache or get a timeout?

16 milliseconds looks weird to me, It think I would be cleaner this way:

```kt
val result = policyForUri(uri).getNow(null)
if (result == null) ApplicationManager.getApplication().executeOnPooledThread { orElse(completable.get()) }
return result
```

dominiccooney: We want to retrieve it if Agent is up and responding quickly. Otherwise, we will be refreshing editor banners every time you open a file, because the first retrieval will always miss the cache. Refreshing banners refreshes all editor banners, not just the one that missed the cache.

pkukielka (approved):
LGTM

PR #1400, opened 2024-04-26
Feature-tracking: Add a JSON file describing JetBrains status of various features

dominiccooney:
This is the JetBrains-side description of the features in sourcegraph/cody#3846

## Test plan

This is a data-only change.


features.json5
```diff
@@ -0,0 +1,68 @@
+{
+    "features": [
```
kalanchan, line 2: just to confirm, we're still sticking with features being an array instead of an object with their keys as the feature name right?

kalanchan (approved):
gogogo!

features.json5
dominiccooney: I vacillated on this, sent you feedback and then deleted it. You're right, an array is better.

When you're joining the features, it is easier to have the features as an object because you can do lookups easily. But later when we're passing features around, it is easier to have the feature as *one object* with a name. Otherwise we will be passing `[name, featureData]` tuples everywhere or always converting `Object.entries` to a two-prop `{name: ..., data: ...}` with a lot of useless drilling.

So the array is better. We're just not feeling the benefits yet because we're not really doing much feature handling (my PR over here does one join for merging... sourcegraph/cody#3956 and the array is fine really.)

PR #1407, opened 2024-04-29
More work on Inline Edits dialog ui

steveyegge:
Changes include:
- refactored all the move/resize logic out of EditCommandPrompt into a helper class
- got the move/resize logic working correcty and cleaned up
- refactored some utilities into a helper class
- did a more thorough job of listener cleanup on disposal

## Test plan

This requires manual testing until we have UI/robot tests. It has a test in TESTING.md that covers this functionality (move/resize).


pkukielka (approved):
LGTM

PR #1410, opened 2024-04-29
Change "Check Usage" to "Manage Account" and add param for helping login mismatch problems

toolmantim:
This PR:
* Renames the "Check Usage" button to "Manage Account"
* Adds a `cody_client_user` parameter to the URL to help users where their web session is logged into a different account than their client
  * Uses sourcegraph/sourcegraph#62039)
  * Equivalent to sourcegraph/cody#3868

## Test plan

- Logged in to dot com
- Click "Manage Account"
- Verify URL has param

### DIFF
--- a/src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
@@ -81,14 +81,14 @@ class MyAccountTabPanel(val project: Project) : JPanel() {
         upgradeButton.component.putClientProperty(DarculaButtonUI.DEFAULT_STYLE_KEY, true)
       }
       button("Manage Account") {
+        val manageUrl = "${ConfigUtil.DOTCOM_URL}" + "cody/manage"
         val account = CodyAuthenticationManager.getInstance(project).getActiveAccount()
-        val url =
-            "${ConfigUtil.DOTCOM_URL}" +
-                "cody/manage" +
-                (if (account !== null)
-                    "?cody_client_user=${URLEncoder.encode(account.name, "UTF-8")}"
-                else null)
-        BrowserUtil.browse(url)
+        if (account != null) {
+          BrowserUtil.browse(
+              manageUrl + "?cody_client_user=" + URLEncoder.encode(account.name, "UTF-8"))
+        } else {
+          BrowserUtil.browse(manageUrl)
+        }
       }
     }
     if (accountTier == AccountTier.DOTCOM_FREE) {
--- a/src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
@@ -9,7 +9,6 @@ import com.intellij.ui.dsl.builder.panel
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.ui.JBUI
 import com.sourcegraph.cody.config.AccountTier
-import com.sourcegraph.cody.config.CodyAccount
 import com.sourcegraph.cody.config.CodyAuthenticationManager
 import com.sourcegraph.common.CodyBundle
 import com.sourcegraph.common.UpgradeToCodyProNotification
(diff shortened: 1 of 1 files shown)


src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
```diff
@@ -79,7 +81,10 @@ class MyAccountTabPanel(val project: Project) : JPanel() {
             button("Upgrade") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/subscription") }
         upgradeButton.component.putClientProperty(DarculaButtonUI.DEFAULT_STYLE_KEY, true)
       }
-      button("Check Usage") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/manage") }
+      button("Manage Account") {
+          val username = CodyAuthenticationManager.getInstance(project).getActiveAccount()?.let(CodyAccount::username)
```
toolmantim: I was hoping you could access `username` just as you do `CodyAccount::displayName` but alas it doesn't seem to work:

```
> Task :compileKotlin FAILED
e: file:///Users/tim/Codez/jetbrains/src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt:85:93 Not enough information to infer type variable R
e: file:///Users/tim/Codez/jetbrains/src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt:85:110 Unresolved reference: username
```

pkukielka: There is no such filed in the `CodyAccount`, you need `name` or `displayName` ;)

```
@Tag("account")
data class CodyAccount(
    @NlsSafe @Attribute("name") override var name: String = "",
    @Attribute("displayName") var displayName: String? = name,
    @Property(style = Property.Style.ATTRIBUTE, surroundWithTag = false)
    override val server: SourcegraphServerPath =
        SourcegraphServerPath.from(ConfigUtil.DOTCOM_URL, ""),
    @Attribute("id") override var id: String = generateId(),
) : ServerAccount() {
```

toolmantim: I thought I tried `name` but I mustn't have, because that worked! Thank you!!

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
```diff
@@ -79,7 +81,16 @@ class MyAccountTabPanel(val project: Project) : JPanel() {
             button("Upgrade") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/subscription") }
         upgradeButton.component.putClientProperty(DarculaButtonUI.DEFAULT_STYLE_KEY, true)
       }
-      button("Check Usage") { BrowserUtil.browse(ConfigUtil.DOTCOM_URL + "cody/manage") }
+      button("Manage Account") {
+        val username =
+            CodyAuthenticationManager.getInstance(project)
+                .getActiveAccount()
+                ?.let(CodyAccount::name)
```
dominiccooney: You should be able to chain the access:

```suggestion
                ?.name
```

The suggestion is equivalent to `x?.name` in TypeScript. The thing you wrote is a bit like `x ? Object.getOwnPropertyDescriptor(CodyAccount.prototype, 'name').get.apply(x) : null`.

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
```diff
@@ -79,7 +81,16 @@ class MyAccountTabPanel(val project: Project) : JPanel() {
...
+        BrowserUtil.browse(
+            ConfigUtil.DOTCOM_URL +
+                "cody/manage?cody_client_user=" +
+                URLEncoder.encode(username, "UTF-8"))
```
dominiccooney: This will throw a NullPointerException if `username` is null, but I guess that's impossible given we know this is a free account, it must have a name...

dominiccooney (approved):
Looks good, one bit of style feedback inline.

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
toolmantim: lol. Fixed!

src/main/java/com/sourcegraph/cody/MyAccountTabPanel.kt
toolmantim: I just updated it to explicit handle null accounts explicitly

PR #1419, opened 2024-04-30
Adding help to troubleshoot error 134

arafatkatze:
I get this very bizarre  error in my build of Jetbrains and its error 134. The causality of this error is that the computer does not have enough memory. Interestingly enough I have 48 Gigs of ram and I still get this error on the rare times when I have many different things running different ide's and other things together. Initially I thought this error was non deterministic because I thought that restarting the computer solved it for me one time. 

The reason it solved it was because it closed all the processs but yeah the memory exceeded is something to keep in mind since this error usually doesn't tell you anything and happens rarely I added a troubleshooting step too. 

## Test plan
It doesn't need it


pkukielka (approved):
LGTM

PR #1420, opened 2024-04-30
Move management of session lifecycle (undo, finish, etc) back to agent

pkukielka:
## Changes

* Remove manual `undo` and base on the one from agent
* Allow finishing of the session being controled by the agent as well instead of managing it manually
* Remove code for custom scrolling/selection restoration after undo - lost of the selection was caused by the bug in the `showDocument` method and proper scrolling should be managed by agent as well: 
https://github.com/sourcegraph/cody/pull/4061/

## Test plan

Not until we have test framework.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
@@ -28,14 +28,12 @@ class FixupService(val project: Project) : Disposable {
   /** Entry point for the inline edit command, called by the action handler. */
   fun startCodeEdit(editor: Editor) {
     if (!isEligibleForInlineEdit(editor)) return
-    cancelActiveSession()
     currentEditPrompt.set(EditCommandPrompt(this, editor, "Edit Code with Cody"))
   }
 
   /** Entry point for the document code command, called by the action handler. */
   fun startDocumentCode(editor: Editor) {
     if (!isEligibleForInlineEdit(editor)) return
-    activeSession?.finish()
     DocumentCodeSession(this, editor, editor.project ?: return)
   }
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -60,22 +58,17 @@ class FixupService(val project: Project) : Disposable {
...
 
-  // Fully cancels/retracts any current session.
-  fun cancelActiveSession() {
-    try {
-      activeSession?.finish()
-    } catch (x: Exception) {
-      logger.warn("Error while disposing session", x)
+  fun waitUntilActiveSessionIsFinished() {
+    while (activeSession != null) {
+      Thread.sleep(100)
```
pkukielka, line 68: That is kind of bad, and proper fix would be to manage that kind of state purely on the agent side.
But that would require bigger changes, so it is what it is for now.
I think I will add N number of retries there and if it will fail to finish I will use error lenses to communicate an issue.

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -94,30 +94,29 @@ abstract class FixupSession(
     CodyAgentService.withAgent(project) { agent ->
       workAroundUninitializedCodebase()
 
+      // All this because we can get the workspace/edit before the request returns!
```
steveyegge, line 97: I think we really ought to fix this problem too. The client protocol requires you to prepare for receiving information about a session/task that hasn't been created yet, because of a race condition. I've filed a [tracking issue](https://github.com/sourcegraph/jetbrains/issues/1455) for it.

steveyegge (approved):
Thanks for doing this cleanup. LGTM.

mkondratek (approved):
lgtm, and it's a code removal 💟

PR #1423, opened 2024-04-30
Fix :runIde gradle command

pkukielka:
## Changes

Make sure `:runIde` properly copies `node` artefacts so it also works in development mode.

## Test plan

N/A


dominiccooney (approved):
Thanks for the quick turnaround!

Issue #1256, opened 2024-04-01
[Cody Ignore] Update Sidebar UI with Cody Ignore changes


dominiccooney:
Other UI guidance about open files should basically be met by #1253 .

```[tasklist]
### Tasks
- [x] Commands: Add banner #1432
- [x] Commands: Disable buttons #1432
- [ ] Repo selector: Repo is listed but is disabled with a label/message saying it is ignored
- [x] At-mentions:  File is shown but is disabled with a label/message saying it is ignored. Okay if file is excluded from fuzzy @-mention filtering.
- [x] Confirm JetBrains doesn't/won't add for GA a right click on a file, add to chat affordance
```

Adapted from [Figma](https://www.figma.com/file/m9Yvgv6W3aPRLfXnxgwnEz/JetBrains?type=design&node-id=1297-78439&mode=design&t=8YgkN0zf2728hX74-0) and @toolmantim 's summary of [product interactions in the RFC:](https://docs.google.com/document/d/10tD8n-5py4XWQD9c7doJB7zIuxa7lwiOkpo2d5NLlnA/edit#bookmark=id.8smvyvjidz0a)

danielmarquespt:
Relevant UX direction on this issue
https://github.com/sourcegraph/jetbrains/issues/1253#issuecomment-2057055486

danielmarquespt:
> @ mentioning an ignored file in the message input ⇒ File is shown but is disabled with a label/message saying it is ignored. Okay if file is excluded from fuzzy @-mention filtering.

AFAIK this is not a requirement since @-mentioning is only available for Current-repo files, and blocking chat all together for ignored repos means that @mentioning is not possible

PR #1432, opened 2024-05-01
Cody Ignore: Manual autocomplete trigger notification, sidebar command button sidebar UX and sundry

### LINKED ISSUE #1253: [Cody Ignore] Add a banner to editors with Cody Ignored files open

dominiccooney:
- Intercepts the manual autocomplete trigger and displays a notification
- Disables the buttons in the sidebar command tab when in an ignored file
- Displays a banner in the sidebar command tab when in an ignored file
- No longer displays banners on individual editors
- Notification text, tooltip text has been updated to match latest UX proposal

https://www.loom.com/share/3a417616ef4f4701ac0e02bcc0c61e64

Part of #1252, #1256

Closes #1253

## Test plan

Manually tested:

- Run plugin with environment `CODY_JETBRAINS_FEATURES=cody.feature.internals-menu=true`
- Use status bar, Internals, Testing: Cody Ignore and change the policy to match your repo, turn the policy override on.
- Check that autocomplete does not happen
- Check that the manual autocomplete trigger displays a notification
- Check that right click menu actions for Explain, Smell and "Test V1" display a notification; same for their keyboard shortcuts
- Check that the buttons in the sidebar command tab are disabled and a banner is displayed
- Check that the status bar Cody icon displays a slash and tooltip when in an ignored file
- Check that all the above and UX affordances respond when the policy changes (override off, different repo name, etc.)


src/main/resources/CodyBundle.properties
```diff
@@ -183,5 +183,9 @@ gotit.autocomplete.message=This is how Cody displays autocomplete suggestions.<b
 
 # Other Actions
 action.cody.restartAgent.text=Restart Cody Agent
-ignore.action-in-ignored-file.detail=Cody has ignored this file because of your Sourcegraph admin policy
-ignore.action-in-ignored-file.learn-more-cta=Learn more
+ignore.action-in-ignored-file.detail=This file has been ignored by an admin. Autocomplete, commands, and other Cody features are disabled.
```
odisseus, line 186: "Ignored by an admin" sounds a bit ambiguous. How about "excluded" or "put on the ignore list"?

src/main/resources/CodyBundle.properties
```diff
@@ -183,5 +183,9 @@ gotit.autocomplete.message=This is how Cody displays autocomplete suggestions.<b
...
+ignore.action-in-ignored-file.learn-more-cta=Learn about Cody Ignore
+ignore.action-in-ignored-file.title=Cody is Limited
```
odisseus, line 188: "Cody assistance is disabled"?

src/main/resources/CodyBundle.properties
dominiccooney: These strings come from our UX designers @toolmantim and @danielmarquespt (although I took the liberty of replacing "repo" with "file" in some instances... "file" will be more durable if planned changes to Ignore arrive and we don't loop back to update these.)

src/main/resources/CodyBundle.properties
dominiccooney: Same reason as above... UX and PM have 👍  these

dominiccooney: Sounds like the label may change. Still, would like to get this landed. It is easy enough to update the string later.

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -234,42 +238,52 @@ class CodyAutocompleteManager {
...
-                if (triggerKind == InlineCompletionTriggerKind.INVOKE ||
-                    !UpgradeToCodyProNotification.isFirstRLEOnAutomaticAutocompletionsShown) {
-                  handleError(project, error)
+      if (triggerKind == InlineCompletionTriggerKind.INVOKE &&
+          IgnoreOracle.getInstance(project).policyForUri(virtualFile.url, agent).get() !=
+              IgnorePolicy.USE) {
+        runInEdt { ActionInIgnoredFileNotification().notify(project) }
+        resetApplication(project)
+        resultOuter.cancel(true)
+        null
```
pkukielka, line 247: Is that `null` needed everywhere?

/e I just noticed it was there already, but still it seems like it could be removed from all branches, `withAgent` closure does not require from us to return anything.

src/main/kotlin/com/sourcegraph/cody/commands/ui/CommandsTabPanel.kt
```diff
@@ -35,6 +45,16 @@ class CommandsTabPanel(
       addInlineEditActionButton("cody.documentCodeAction")
       addInlineEditActionButton("cody.testCodeAction")
     }
+
+    addHierarchyListener {
+      if (!project.isDisposed) {
+        if (it.component.isShowing) {
```
pkukielka, line 51: Any reason we need to keep adding/removing it and cannot just leave it be?

src/main/resources/CodyBundle.properties
danielmarquespt: The "By admin" has the objective of letting the user know that there is nothing they can do, or is not a setting.

Saying "excluded" or on an "ignore list" might make the user think its a local setting or and make then look for it thinking they can actually change it.

However the actual wording can be tweaked

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/commands/ui/CommandsTabPanel.kt
dominiccooney: Happy to change this in a follow up... what would the trigger for removing the listener be?

src/main/kotlin/com/sourcegraph/cody/ignore/CommandPanelIgnoreBanner.kt
```diff
@@ -0,0 +1,41 @@
...
+              CodyBundle.getString("ignore.sidebar-panel-ignored-file.learn-more-cta"),
+              { BrowserUtil.browse(CODY_IGNORE_DOCS_URL) },
+              false)
+          icon(Icons.CodyLogoSlash)
+        })
+
+    // These colors cribbed from EditorComposite, createTopBottomSideBorder
+    val scheme = EditorColorsManager.getInstance().globalScheme
+    val borderColor =
+        scheme.getColor(EditorColors.SEPARATOR_ABOVE_COLOR)
```
odisseus, line 31: Looks like this field has been [removed in 2022.3](https://github.com/JetBrains/intellij-community/commit/ee2edf552f8cf54c183a5138afc6ba12fc353bae). I've just got a `NoSuchFieldError` while using Idea CE 2023.3.6.

**P.S.** https://github.com/sourcegraph/jetbrains/pull/1453

PR #1439, opened 2024-05-02
Finished up the Edit/Retry feature

steveyegge:
Edit/Retry had a couple of issues, both fixed:

1. It wasn't Undoing before retrying, so it was leaving the original text in the document. Pretty bad bug.
2. Document Code needed to show the "instruction" (prompt) generated by the Agent. This mostly a snazzy extra from VS Code.

This PR goes hand-in-hand with a [PR in sourcegraph/cody](https://github.com/sourcegraph/cody/pull/4010) that sends the instruction along with task updates, so that it's available if the user chooses to retry.

This PR also fixes a couple of bugs, including a really annoying one with Undo, where it was scrolling to the top of the document.

## Test plan

Test plan is in TESTING.md.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
@@ -2,6 +2,7 @@ package com.sourcegraph.cody.edit.sessions
 
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.runInEdt
 import com.intellij.openapi.command.WriteCommandAction
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.editor.Document
@@ -78,7 +79,7 @@ abstract class FixupSession(
 
   init {
     triggerFixupAsync()
-    ApplicationManager.getApplication().invokeLater { Disposer.register(controller, this) }
+    runInEdt { Disposer.register(controller, this) }
   }
 
   private val document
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -239,15 +247,14 @@ abstract class FixupSession(
   }
 
   fun retry() {
-    // TODO: The actual prompt we sent is displayed as ghost text in the text input field, in VS
-    // Code.
-    // E.g. "Write a brief documentation comment for the selected code <etc.>"
-    // We need to send the prompt along with the lenses, so that the client can display it.
+    val instruction = instruction
+    undo()
```
pkukielka, line 251: Hey @steveyegge, I assume this is the part you mentioned you still plan to rework, right?

steveyegge: I understand what you're saying, but consider:

- it's completely broken for users today.
- fixing it properly is a lot of work. At least 2 days, maybe more. I tried already, and backed out, because the VS Code extension implementation of Retry is just completely tangled up with the UI. It will be a full reimplementation with new tests, etc.
- fixing it properly is nowhere near as high priority as some of the other issues on our plate. That's why I made the call to do it this way.

So. With that in mind, can we please let this through for now?

pkukielka: Sure, let's do it.
I still have one concern regarding potential edge case - calling undo in agent not causes it to send us edit information which is doina a real undo.
In most cases we were just missing/ignoring it, because current session was already disposed when that messages was coming in, and so it was ignored.
I'm not sure if that will be the case in your situation. It may be dependent on timing and only manifest as an issue sometimes I think.

That said let's merge it, and I will give it some testing as well as part of my other work.

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
steveyegge: Ah,, interesting. I've seen that happening quite a bit, and yeah, we do ignore the extra, but that does need to be cleaned up.

PR #1448, opened 2024-05-03
Implemented Cody Ignore for Inline Edits

steveyegge:
This was a relatively simple set of changes:

- There is a new pub/sub topic for ignore policy changes
- It is published every time the policy changes for a file
- The FixupService subscribes to it
- Profit.

Whenever the file is ignored, the FixupService, which manages sessions, cancels any active session and refuses to do any more (showing the notification instead) until the file is un-ignored.

Two other misc fixes:

- fixed a runtime "not on EDT" exception in CodyDocumentListener
- added a hotkey for Document Code, as it was missing one

## Test plan

I tested this manually for now, using Dominic's internals override for IgnoreOracle.

I put a test plan for it in TESTING.md.

I will write an integration test for it next week when those tests are online.


TESTING.md
```diff
@@ -414,6 +414,47 @@ All tests involving prompt history should end with
...
+an "Internals" submenu at the bottom of your IDE window, next to Cody.
+
+Ensure that if the file is ignored, no inline edits can be performed.
+
+1. Open a source file in any project. 
+2. Open the Internals submenu.
+3. Check the box to override for testing.
+4. Change the repo to match whatever repo the source file is in.
+5. Close the submenu.
+6. Click somewhere in the code and initiate an Inline Edit (control-shift-enter0
```
dominiccooney:
```suggestion
6. Click somewhere in the code and initiate an Inline Edit (control-shift-enter)
```

TESTING.md
```diff
@@ -414,6 +414,47 @@ All tests involving prompt history should end with
...
+
+No matter what combination of enabling/disabling policies you use,
+all of the following should be true for each test:
+
+1. Whenever the current repo/file is ignored, inline edits should stop working.
+   - The Edit Instructions dialog should not display, and disappear if active.
+   - The Document Code function should not work at all (via keyboard or menu).
+2. If an edit is in progress, and its file becomes ignored, then the edit in
+   progress should cancel/undo.
+3. When the ignore policy for the current file returns to USE, inline edits
```
dominiccooney: USE is implementation language, let's make this accessible to the tester at the keyboard without reference to code.

src/main/kotlin/com/sourcegraph/cody/edit/EditUtil.kt
```diff
@@ -91,4 +100,28 @@ object EditUtil {
       color.darker()
     }
   }
+
+  /**
+   * Return true if the currently selected text editor for `project` is visiting a file that is
+   * currently ignored by CodyIgnore.
+   */
+  fun currentlySelectedFileIsIgnored(project: Project?): Boolean {
+    if (project == null || project.isDisposed) {
```
dominiccooney: Because this may block for 30 seconds, let's add an assertion that we are not on EDT

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -45,13 +75,20 @@ class FixupService(val project: Project) : Disposable {
     TestCodeSession(this, editor, editor.project ?: return)
   }
 
+  @RequiresEdt
```
dominiccooney, line 58: Until the annotation processor supports Kotlin, should we back these up with manual assertions?

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -45,13 +75,20 @@ class FixupService(val project: Project) : Disposable {
...
       return false
     }
     if (!CodyEditorUtil.isEditorValidForAutocomplete(editor)) {
-      logger.warn("Inline edit invoked when editing not available")
+      logger.warn("Edit code invoked when editing not available")
+      return false
+    }
+    val policy = IgnoreOracle.getInstance(project).policyForEditor(editor)
+    if (policy != IgnorePolicy.USE) {
+      runInEdt { ActionInIgnoredFileNotification().notify(project) }
```
dominiccooney, line 70: If this requires EDT, then it isn't worth spinning the event loop to show the notification, just show it.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -81,6 +118,20 @@ class FixupService(val project: Project) : Disposable {
     activeSession = null
   }
 
+  private fun handlePolicyChange(context: IgnoreMessageBus.Context) {
```
dominiccooney: What behavior does this produce?

It doesn't seem right to have FixupService producing a notification for "any file that someone `policyForUri`-ed in the past hour(s)" (the last time we would have fetched the policy and started caching results for individual files.)

src/main/kotlin/com/sourcegraph/cody/edit/EditUtil.kt
```diff
@@ -91,4 +100,28 @@ object EditUtil {
       color.darker()
     }
   }
+
+  /**
+   * Return true if the currently selected text editor for `project` is visiting a file that is
+   * currently ignored by CodyIgnore.
+   */
+  fun currentlySelectedFileIsIgnored(project: Project?): Boolean {
```
dominiccooney: First, is this function used? Cmd-F doesn't find it (but GitHub code review sucks for cmd-F...)

Second, let's de-dup this with `IgnoreOracle.getInstance(project).isEditingIgnoredFile`, which the autocomplete status indicator in the status bar is built on.

That does not block, it just gives you whether the last focused file is ignored, with some some crude handling of races if the focus changes quickly (the `focusedPolicy != policy && willFocusUri == uri` part.) (Crude because the reads of those two volatile fields could be out of sync...)

dominiccooney (changes requested):
Some feedback inline.

The topic needs work: The impression the API gives is you can hang out listening for policy updates for a URI, but these updates are driven by someone pumping policies with `policyForUri`.

The key issues to clarify are:
- Whether IgnoreOracle is holding onto a set of URIs and providing updates, or whether the client is responsible for the set of URIs it is interested in.
- How the lifetimes of listeners, etc. work.

The original API... policyForUri/didChange, settles the questions with the clients being responsible for tracking what URIs they're interested in—`cache` is just a cache—and the lifetimes are independent.

The thing I liked about that approach is it lets ephemeral uses, like "hey in the past hour I have opened 100 files but only care about the 3 remaining ones", or "hey I am an ephemeral file autocomplete popup, I am asking about 40 files but I don't care about updates" very simply: The results go in the cache, when the policy updates, the cache is expunged and nobody is notified or going to bother re-fetching their policy because nobody cares about those files any more.

The topic here relies on some clients re-polling with `policyForUri` to "set" the URIs of interest, which seems indirect if not fragile. If you solve that by making the IgnoreOracle responsible for tracking the URIs of interest, now you have a harder lifetime question—we need an API to remove URIs clients are no longer interested in, or give them an API which encourages them re-ping policyForUri when the policy *did* change (which you would do by *not* providing the new policy in the update.)

dominiccooney (approved):
Looks great.

The summary description needs an update since the topic is gone.

PR #1453, opened 2024-05-06
Removed the reference to SEPARATOR_ABOVE_COLOR

odisseus:
## Test plan

Build the plugin and load it in IntelliJ 2022.3 or newer. Check the log for occurrences of `NoSuchFieldError`.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ignore/CommandPanelIgnoreBanner.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ignore/CommandPanelIgnoreBanner.kt
@@ -27,9 +27,7 @@ class CommandPanelIgnoreBanner() : NonOpaquePanel() {
 
     // These colors cribbed from EditorComposite, createTopBottomSideBorder
     val scheme = EditorColorsManager.getInstance().globalScheme
-    val borderColor =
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):
Thanks!

src/main/kotlin/com/sourcegraph/cody/ignore/CommandPanelIgnoreBanner.kt
```diff
@@ -27,9 +27,7 @@ class CommandPanelIgnoreBanner() : NonOpaquePanel() {
 
     // These colors cribbed from EditorComposite, createTopBottomSideBorder
     val scheme = EditorColorsManager.getInstance().globalScheme
-    val borderColor =
-        scheme.getColor(EditorColors.SEPARATOR_ABOVE_COLOR)
```
pkukielka, line 31: @dominiccooney Is there a reason it was implemented this way in the first place??
Code suggests that `scheme.getColor(EditorColors.TEARLINE_COLOR)` may not be sometimes present, and if that the case I'm not convinced that fix is proper?
If you are sure this is ok you can ignore me, just double checking.

PR #1454, opened 2024-05-06
Apply IntelliJ formatting to the edited code snippet that comes from the agent

odisseus:
## Test plan

Tested manually, mostly on [this file](https://github.com/odisseus/java-ipfs-http-client/blob/master/src/main/java/io/ipfs/api/Pair.java). 

### Things that are confirmed to work (prompts might need to be reworded)
Undoing these operations also works perfectly.

1. Select all text, prompt "Remove all blank lines".
    * The blank lines will be removed only inside the methods. The blank lines between the methods are removed by the agent but restored by the formatter.
2. Select a whole method, prompt "Remove all blank lines".
3. Select a piece of the code that references a variable, prompt "Check that the <variable> is not null".
    * This should add some lines of code, growing the selection accordingly.

### Intentionally misformatting the file
This flow technically works as expected, but the result might confuse the user.
1. Open a Java file. Make sure that it has formatting problems, e.g. multiple statements (ending with `;`) in the same line.
5. Select a few lines above and below the misformatted line, and invoke "Cody → Edit Code".
6. In the prompt, write "Remove all indents".
7. The result should come out correctly formatted, with the proper indents and each statement in its own line.
    * But if you place a breakpoint inside `FixupSession.performInlineEdits`, you'll see that the edit arrives from the agent without indents (as requested), and is later formatted by IntelliJ.
8. Click "Undo" or "Edit & Retry".
9. The code will return to its original state, complete with the original formatting problem.

### Things that don't work
These things seem to be broken in master.
* In some cases, the selection is smaller than the modified part of the code.
* If you ask Cody to edit code without selecting a range in the file, the generated code will not be selected a,d the Undo action will be broken.
* Show Diff action seems to be broken in many cases.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
@@ -320,7 +320,7 @@ abstract class FixupSession(
     }
   }
 
-  fun performInlineEdits(edits: List<TextEdit>) {
+  fun performInlineEdits(edits: List<TextEdit>, isUndo: Boolean = false) {
     // TODO: This is an artifact of the update to concurrent editing tasks.
     // We do need to mute any LensGroup listeners, but this is an ugly way to do it.
     // There are multiple Lens groups; we need a Document-level listener list.
@@ -355,22 +355,26 @@ abstract class FixupSession(
           }
         }
 
-        // It is safe to use selectionRange here, because it is set in the constructor
-        // by calling triggerFixupAsync() which in turn calls ensureSelectionRange().
-        // The value of this property does not change
-        // when the user selects a different snippet in the UI.
-        // FIXME this causes a problem when the entire file becomes shorter than the original selection range.
-        selectionRange!!.let {
-          val tr = TextRange(it.start.toOffset(document), it.end.toOffset(document))
-          val input = document.getText(tr)
-          val formatted =
-              CodyFormatter.formatStringBasedOnDocument(
-                  input, project, document, tr, editor.caretModel.offset)
-          if (input.replace(Regex("\\s"), "") != formatted.replace(Regex("\\s"), "")) {
-            logger.error(
-                "Skipped internal formatting because the formatted text is different from the original text")
+        // Skip the formatting if these edits are part of an undo operation.
+        if(!isUndo){
+          // It is safe to use selectionRange here, because it is set in the constructor
+          // by calling triggerFixupAsync() which in turn calls ensureSelectionRange().
+          // The value of this property does not change
+          // when the user selects a different snippet in the UI.
+          // FIXME this causes a problem when the entire file becomes shorter than the original
+          // selection range.
+          selectionRange!!.let {
+            val tr = TextRange(it.start.toOffset(document), it.end.toOffset(document))
+            val input = document.getText(tr)
+            val formatted =
+                    CodyFormatter.formatStringBasedOnDocument(
+                            input, project, document, tr, editor.caretModel.offset)
+            if (input.replace(Regex("\\s"), "") != formatted.replace(Regex("\\s"), "")) {
+              logger.error(
+                      "Skipped internal formatting because the formatted text is different from the original text")
+            }
+            document.replaceString(tr.startOffset, tr.endOffset, formatted)
           }
-          document.replaceString(tr.startOffset, tr.endOffset, formatted)
         }
 
         performedActions += currentActions
--- a/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
@@ -92,6 +92,8 @@ abstract class FixupSession(
         }
       }
 
+  @Volatile private var undoInProgress: Boolean = false
+
   init {
     triggerFixupAsync()
 
@@ -248,6 +250,7 @@ abstract class FixupSession(
   fun undo() {
     runInEdt { showWorkingGroup() }
     CodyAgentService.withAgent(project) { agent ->
+      undoInProgress = true
       agent.server.undoEditTask(TaskIdParam(taskId!!))
     }
   }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -347,6 +349,23 @@ abstract class FixupSession(
           }
         }
 
+        // It is safe to use selectionRange here, because it is set in the constructor
```
pkukielka: What if the edit itself adds or removes lines?
Is that reflected by that selection?
What if you select whole big file but edit only changes one line?

odisseus: I've just tested this situation manually, and this property doesn't seem to change. Nevertheless, both the highlight in the editor and the range being sent to the agent in `textDocument/didChange` are correct (i.e. increased). 

**P.S.** Refactored

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -349,6 +354,31 @@ abstract class FixupSession(
...
 
+        // Skip the formatting if these edits are part of an undo operation.
+        if (!undoInProgress) {
+          // It is safe to use selectionRange here, because it is set in the constructor
+          // by calling triggerFixupAsync() which in turn calls ensureSelectionRange().
+          // The value of this property does not change
+          // when the user selects a different snippet in the UI.
+          // FIXME this causes a problem when the entire file becomes shorter than the original
+          // selection range.
+          selectionRange!!.let {
```
pkukielka, line 368: Please do not use !!

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -349,6 +354,31 @@ abstract class FixupSession(
...
+            val tr = TextRange(it.start.toOffset(document), it.end.toOffset(document))
+            val input = document.getText(tr)
+            val formatted =
+                CodyFormatter.formatStringBasedOnDocument(
+                    input, project, document, tr, tr.startOffset)
+            if (input.replace(Regex("\\s"), "") != formatted.replace(Regex("\\s"), "")) {
```
pkukielka: Do we need a regex there? isn't it the same as just calling `.trimStart()`?

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -349,6 +354,31 @@ abstract class FixupSession(
           }
         }
 
+        // Skip the formatting if these edits are part of an undo operation.
+        if (!undoInProgress) {
+          // It is safe to use selectionRange here, because it is set in the constructor
+          // by calling triggerFixupAsync() which in turn calls ensureSelectionRange().
+          // The value of this property does not change
+          // when the user selects a different snippet in the UI.
+          // FIXME this causes a problem when the entire file becomes shorter than the original
```
pkukielka: Is that going to be fixed as part of this PR?

src/main/kotlin/com/sourcegraph/utils/CodyFormatter.kt
```diff
@@ -34,6 +34,7 @@ class CodyFormatter {
                 .createFileFromText("TEMP", file.fileType, appendedString)
 
         val codeStyleManager = CodeStyleManager.getInstance(project)
+        // FIXME this will fail if cursor < range.startOffset + completionText.length
```
pkukielka: Is that going to be fixed as part of this PR?

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
odisseus: I want to normalise all whitespace in the string, not just the prefix. 

This is a sanity check to cancel formatting if the result is obviously wrong and applying it would break the code. However, I've already fixed the mistake that had caused such results, so this check is probably redundant by now.

**P.S.** Removed

src/main/kotlin/com/sourcegraph/utils/CodyFormatter.kt
odisseus: I have fixed my code that was passing wrong parameters, but IMO there is a problem with the interface of `CodyFormatter.formatStringBasedOnDocument`. I would rather leave it for another time.

**P.S.** Marked as TODO

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
odisseus: I believe that `!!` is justified here. However, I just confirmed that this FIXME still needs to be fixed...

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
odisseus: I have refactored `FixupSession` to use the `RangeMarker` from the IntelliJ API. Unlike the dumb data classes such as `Range`, `TextRange` etc., the `RangeMarker` remains tied to its document, and updates itself whenever the document is modified. The downside is that we need to acquire a read lock before creating a new `RangeMarker`.

This has eliminated the problems with shortening the file. However, I've run into other problems...

src/main/kotlin/com/sourcegraph/utils/CodyFormatter.kt
```diff
@@ -34,6 +34,8 @@ class CodyFormatter {
                 .createFileFromText("TEMP", file.fileType, appendedString)
 
         val codeStyleManager = CodeStyleManager.getInstance(project)
+        // This will fail if cursor < range.startOffset + completionText.length
```
pkukielka, line 37: I think we should not crash whole edit if formatter fails.
Instead we should backoff from formatting, log the error, and continue edit without reformatting.

pkukielka:
LGTM

src/main/kotlin/com/sourcegraph/utils/CodyFormatter.kt
odisseus: Actually, this is the current behaviour. However, the root cause (incorrect ranges being passed) still needs to be fixed.

PR #1457, opened 2024-05-07
Roll the extension to 6c659814

dominiccooney:
This is the extension commit in VSCode 1.16.2, plus the change to send client-name, client-version query string parameters required by Sourcegraph 6.

## Test plan

Manually tested:

- Signed in
- Chatted
- Generated an autocomplete


kalanchan (approved):
gogogo!

PR #1458, opened 2024-05-07
Run plugin verifier in CI

dominiccooney:
This would have caught the regression in #1432 fixed in #1453

## Test plan

CI


pkukielka (approved):
LGTM

odisseus (dismissed):

odisseus:
It looks like the plugin verifier tool attempts to download several versions of IntelliJ (which constitute a few GB of data). I have no idea why this step fails in CI, but it might be caused by exceeding some kind of traffic quota.

odisseus (dismissed):
Looks like you need to fix the formatting. Otherwise the changes look good.

PR #1460, opened 2024-05-07
Cody Ignore: Ignore Chat UX updates

dominiccooney:
- @-mention selector for files badges ignored files
- Context inspection summary displays elaborate count of ignored files
- Context inspection disclosure displays ignored files with strikeout
- Repo picker displays ignored repos with a slashed circle, tooltips, etc.

This also changes the enhanced context selector to reflect the state of the extension. This has many effects:

- The implicitly included repository is displayed (adds tooltips, etc. for this.)
- Removing repos is done by clicking the checkboxes, and they disappear. This needs UX feedback. It is close to VSCode, but it is weird to have the items disappear.
- There's a bug in Agent where the git extension shim doesn't initialize the repository `state.remotes`, causing us to pick up the implicit repo poorly.

Part of #1256

## Test plan

Screenshot 2024-05-07 at 23 12 17

- Sign in to sg02.sourcegraphcloud.com
- Open the Cody repo, add a bunch of repos and verify the repos have a circle-slash in line with the sg02 policy
- Mention files with @ and the picker should say the files are ignored
- Sending a chat with ignored mentioned files and the context disclosure should display the files with strikeout text

### DIFF
--- a/src/main/java/com/sourcegraph/cody/PromptPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/PromptPanel.kt
@@ -314,11 +314,12 @@ data class DisplayedContextFile(val contextItem: ContextItem) {
     val contextItemFile = contextItem as? ContextItemFile
     val isIgnored = contextItemFile?.isIgnored == true
     val isTooLarge = contextItemFile?.title == "large-file" || contextItemFile?.isTooLarge == true
-    val warnIfNeeded = when {
-      isIgnored -> "<i> - ⚠ Ignored by an admin setting.</i>"
-      isTooLarge -> "<i> - ⚠ File too large</i>"
-      else -> ""
-    }
+    val warnIfNeeded =
+        when {
+          isIgnored -> "<i> - ⚠ Ignored by an admin setting.</i>"
+          isTooLarge -> "<i> - ⚠ File too large</i>"
+          else -> ""
+        }
     return "<html>${contextItem.displayPath()}$warnIfNeeded</html>"
   }
 }
--- a/src/main/java/com/sourcegraph/cody/PromptPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/PromptPanel.kt
@@ -316,7 +316,7 @@ data class DisplayedContextFile(val contextItem: ContextItem) {
     val isTooLarge = contextItemFile?.title == "large-file" || contextItemFile?.isTooLarge == true
     val warnIfNeeded =
         when {
-          isIgnored -> "<i> - ⚠ Ignored by an admin setting.</i>"
+          isIgnored -> "<i> - ⚠ Ignored by an admin setting</i>"
           isTooLarge -> "<i> - ⚠ File too large</i>"
           else -> ""
         }
--- a/src/main/java/com/sourcegraph/cody/PromptPanel.kt
+++ b/src/main/java/com/sourcegraph/cody/PromptPanel.kt
@@ -312,8 +312,13 @@ class PromptPanel(project: Project, private val chatSession: ChatSession) : JLay
 data class DisplayedContextFile(val contextItem: ContextItem) {
   override fun toString(): String {
     val contextItemFile = contextItem as? ContextItemFile
+    val isIgnored = contextItemFile?.isIgnored == true
     val isTooLarge = contextItemFile?.title == "large-file" || contextItemFile?.isTooLarge == true
-    val warnIfNeeded = if (isTooLarge) "<i> - ⚠ File too large</i>" else ""
+    val warnIfNeeded = when {
+      isIgnored -> "<i> - ⚠ Ignored by an admin setting.</i>"
+      isTooLarge -> "<i> - ⚠ File too large</i>"
+      else -> ""
+    }
     return "<html>${contextItem.displayPath()}$warnIfNeeded</html>"
   }
 }
--- a/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
+++ b/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
@@ -39,6 +39,7 @@ data class ExtensionMessage(
     val errors: String?,
     val query: String? = null,
     val configFeatures: ConfigFeatures? = null,
+    val enhancedContextStatus: EnhancedContextContextT? = null,
 ) {
 
   object Type {
@@ -46,6 +47,7 @@ data class ExtensionMessage(
     const val ERRORS = "errors"
     const val USER_CONTEXT_FILES = "userContextFiles"
     const val SET_CONFIG_FEATURES = "setConfigFeatures"
+    const val ENHANCED_CONTEXT_STATUS = "enhanced-context"
   }
 }
 
@@ -53,4 +55,29 @@ data class WebviewPostMessageParams(val id: String, val message: ExtensionMessag
 
 data class ConfigFeatures(
     val attribution: Boolean,
+)
+
+data class EnhancedContextContextT(
+  val groups: List<ContextGroup>
+)
+
+data class ContextGroup(
+  val dir: String? = null, // URI
+  val displayName: String,
+  val providers: List<ContextProvider>
+)
+
+// This is a subset of the ContextProvider type in lib/shared/src/codebase-context/context-status.ts
+// It covers remote search repositories.
+data class ContextProvider(
+  val kind: String, // "embeddings", "search"
+
+  // if kind is "search"
+  val type: String? = null, // "local", "remote"
+
+  // if kind is "search" and type is "remote"
+  val state: String? = null, // "ready", "no-match",
+  val id: String? = null,
+  val inclusion: String? = null, // "auto" or "manual"
+  val isIgnored: Boolean? = null,
 )
--- a/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
+++ b/src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
@@ -57,27 +57,25 @@ data class ConfigFeatures(
     val attribution: Boolean,
 )
 
-data class EnhancedContextContextT(
-  val groups: List<ContextGroup>
-)
+data class EnhancedContextContextT(val groups: List<ContextGroup>)
 
 data class ContextGroup(
-  val dir: String? = null, // URI
-  val displayName: String,
-  val providers: List<ContextProvider>
+    val dir: String? = null, // URI
+    val displayName: String,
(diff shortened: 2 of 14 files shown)


src/main/java/com/sourcegraph/cody/PromptPanel.kt
```diff
@@ -312,8 +312,14 @@ class PromptPanel(project: Project, private val chatSession: ChatSession) : JLay
 data class DisplayedContextFile(val contextItem: ContextItem) {
   override fun toString(): String {
     val contextItemFile = contextItem as? ContextItemFile
+    val isIgnored = contextItemFile?.isIgnored == true
     val isTooLarge = contextItemFile?.title == "large-file" || contextItemFile?.isTooLarge == true
-    val warnIfNeeded = if (isTooLarge) "<i> - ⚠ File too large</i>" else ""
+    val warnIfNeeded =
+        when {
+          isIgnored -> "<i> - ⚠ Ignored by an admin setting.</i>"
+          isTooLarge -> "<i> - ⚠ File too large</i>"
```
mkondratek: for consistency - pls either user a dot at the end in the second case or remove it from the first message

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
```diff
@@ -54,3 +56,26 @@ data class WebviewPostMessageParams(val id: String, val message: ExtensionMessag
 data class ConfigFeatures(
     val attribution: Boolean,
 )
+
+data class EnhancedContextContextT(val groups: List<ContextGroup>)
```
mkondratek, line 60: there is smth wrong with this name, isn't it?

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
```diff
@@ -18,7 +20,20 @@ class ContextFileActionLink(
...
+          contextItemFile.isIgnored == true || contextItemFile.isTooLarge == true ->
+              Font(
+                  super.getFont().attributes +
+                      (TextAttribute.STRIKETHROUGH to TextAttribute.STRIKETHROUGH_ON))
+          else -> super.getFont()
+        }
+    toolTipText =
+        when {
+          contextItemFile.isIgnored == true -> "File ignored by an admin setting"
+          contextItemFile.isTooLarge == true -> "Excluded due to context window limit"
```
mkondratek, line 34: so it was a valid scenario that we displayed a context file used while actually it has been ignored due to the size, right? then this was missing - thanks for fixing btw! 🙇‍♂️

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
```diff
@@ -18,7 +20,20 @@ class ContextFileActionLink(
 
   init {
     text = contextItemFile.getLinkActionText(project.basePath)
-    toolTipText = contextItemFile.uri.path
+    font =
+        when {
+          contextItemFile.isIgnored == true || contextItemFile.isTooLarge == true ->
+              Font(
+                  super.getFont().attributes +
+                      (TextAttribute.STRIKETHROUGH to TextAttribute.STRIKETHROUGH_ON))
```
mkondratek, line 28: not blocking but worth to verify on windows

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -55,12 +55,15 @@ class ContextFilesPanel(
...
-    val lineCount = contextItemFiles.sumOf { it.range?.length() ?: 0 }
-    val fileCount = filteredFiles.size
-    val lines = "$lineCount line${if (lineCount > 1) "s" else ""}"
-    val files = "$fileCount file${if (fileCount > 1) "s" else ""}"
+    val excludedFileCount = excludedFiles.distinctBy { it.uri }.size
+    val includedFileCount = includedFiles.distinctBy { it.uri }.size
+    val lineCount = includedFiles.sumOf { it.range?.length() ?: 0 }
+    val lines = "$lineCount ${"line".pluralize(lineCount)}"
+    val files =
+        "$includedFileCount ${"file".pluralize(includedFileCount)}${if (excludedFileCount > 0) " — $excludedFileCount ${"file".pluralize(excludedFileCount)} excluded" else ""}"
```
mkondratek: image

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFilesPanel.kt
```diff
@@ -103,3 +106,12 @@ class ContextFilesPanel(
...
 }
+
+// Can pluralize "file" and "line" by adding -s
+private fun String.pluralize(count: Int): String {
+  return if (count == 1) {
+    this
+  } else {
+    "${this}s"
+  }
+}
```
mkondratek:
```suggestion
  private fun String.pluralize(count: Int): String = when {
    count == 1 -> this
    else -> "${this}s"
  } 
```


nit: it can be shorter while still readable

src/main/kotlin/com/sourcegraph/cody/context/RemoteRepoInsight.kt
```diff
@@ -369,6 +380,8 @@ class RemoteRepoCompletionContributor : CompletionContributor(), DumbAware {
             prefixedResult.restartCompletionOnAnyPrefixChange()
             try {
               runBlockingCancellable {
+                // TODO: Extend repo search to consult Cody Ignore and denote repositories that are
+                // ignored.
```
mkondratek, line 384: ☝️

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -218,6 +221,36 @@ class EnterpriseEnhancedContextPanel(project: Project, chatSession: ChatSession)
     return toolbar.createPanel()
   }
 
+  override fun createCheckboxPolicy(): CheckboxTreeBase.CheckPolicy =
+      CheckboxTreeBase.CheckPolicy(
+          /* checkChildrenWithCheckedParent = */ false,
+          /* uncheckChildrenWithUncheckedParent = */ false,
+          /* checkParentWithCheckedChild = */ false,
```
mkondratek, line 228: since these changed we will need to double check the check-uncheck behavior

src/main/resources/CodyBundle.properties
```diff
@@ -189,3 +189,6 @@ ignore.action-in-ignored-file.title=Cody is Limited
 ignore.status-bar-ignored-file.tooltip=This file has been ignored by an admin.\nAutocomplete, commands, and other Cody features are disabled.
 ignore.sidebar-panel-ignored-file.text=<html>This file has been marked as ignored by an admin, which means commands that rely on its contents cannot be executed.</html>
 ignore.sidebar-panel-ignored-file.learn-more-cta=Learn more
+context-panel.tree.node-ignored.tooltip=Repo ignored by an admin setting
```
mkondratek: no strong feelings but personally I prefer the complete name instead of abbreviation so `Repository` instead of `Repo`

mkondratek (commented):
comments

src/main/java/com/sourcegraph/cody/agent/WebviewPostMessageParams.kt
dominiccooney: This matches [the name on the extension side,](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/codebase-context/context-status.ts?L68) where the name kiiiinda makes sense, because it is webview state for `EnhancedContext` but propagated by React Context, so the key for the context should be called `~Context`. We can't export a value and a type with the same name, so we have `EnhancedContextContextT`. Every step makes sense but the result makes no sense...

For now keeping them in sync for easy grepping, and to line up with the generated Kotlin bindings if we switch to those...

src/main/kotlin/com/sourcegraph/cody/context/RemoteRepoInsight.kt
dominiccooney: I looked at VSCode and it isn't doing any badging here (yet... @valerybugakov might be looking at that) so I think this is fine to leave as a TODO for now. If you add ignored repos, you get feedback about it in the tree view.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
dominiccooney: This behavior has only changed for the enterprise context panel, and this is the behavior we need.

Without this change, turning off enhanced context unchecks the repos, *which removes the repos.* That's not how the extension works:

- There's a set of repos: zero or one implicitly included repos, plus explicitly chosen repos up to a limit of 10.
- There's enhanced context toggle, which controls fetching from the set of repos and the open tabs in the editor.
- You can control the repo setting and the enhanced context toggle independently.

Setting these all to false turns off the automatic checking/unchecking.

src/main/kotlin/com/sourcegraph/cody/chat/ui/ContextFileActionLink.kt
dominiccooney: No worries, I was just following along from the VSCode implementation.

src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -218,6 +221,36 @@ class EnterpriseEnhancedContextPanel(project: Project, chatSession: ChatSession)
...
+          /* uncheckParentWithUncheckedChild = */ false)
+
+  override fun updateFromExtension(enhancedContextStatus: EnhancedContextContextT) {
```
steveyegge: Consider whether you want to lean into the term _extension_ leaking into the client-side terminology and mental model. There seem to be an order of magnitude more uses of _agent_ than _extension_. Frankly reading it here gave me pause more than once, wondering if it meant something subtly different from Agent. I assume it does not, so perhaps we should standardize on a naming convention.

Edit:  It gets especially tricky when the client plugins are called "extensions", which is not the case in JetBrains but is in other clients (e.g. Emacs).

steveyegge (approved):
LGTM, although if you agree on the name change, we might want to rip off the band-aid and change at least some of the other uses of "extension".

PR #1463, opened 2024-05-07
Removed API usages that were deprecated in IntelliJ 2022.1.

odisseus:
This PR is related to #1458.

## Test plan

`./gradlew runPluginVerifier`

### DIFF
--- a/src/main/java/com/sourcegraph/find/browser/BrowserAndLoadingPanel.java
+++ b/src/main/java/com/sourcegraph/find/browser/BrowserAndLoadingPanel.java
@@ -11,7 +11,6 @@ import com.intellij.util.ui.StatusText;
 import com.sourcegraph.cody.config.ui.AccountConfigurable;
 import java.awt.*;
 import javax.swing.*;
-import org.apache.commons.lang.SystemUtils;
 import org.apache.commons.lang.WordUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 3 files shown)


odisseus:
It looks like the plugin verifier tool attempts to download several versions of IntelliJ (which constitute a few GB of data). I have no idea why this step fails in CI, but it might be causedd by exceedding some kind of traffic quota.

src/main/java/com/sourcegraph/find/browser/BrowserAndLoadingPanel.java
```diff
@@ -76,7 +75,7 @@ private void refreshUI() {
 
     } else if (errorMessage != null) {
       String wrappedText = WordUtils.wrap("Error: " + errorMessage, 100);
-      String[] lines = wrappedText.split(SystemUtils.LINE_SEPARATOR);
+      String[] lines = wrappedText.split(System.lineSeparator());
```
mkondratek: I remember some troubles with this one. Please check and review the UI in the latest and the oldest supported versions 🙏

odisseus: I'm just going to split by `\n`. This should work on all platforms.

pkukielka: > I remember some troubles with this one

There was a problem with a separator but it was for UI separator element - maybe you think about that other one?:)

pkukielka (approved):
LGTM

Issue #1322, opened 2024-04-16
Enterprise repo picker follow up polish


dominiccooney:
These are follow up issues from #982 and code review feedback on #1308

```[tasklist]
### Tasks
- [ ] Add [tooltip](https://jetbrains.design/intellij/controls/tooltip/) on hover of "Chat Context" for enterprise users with below text.
- [ ] Remove the "Help" toolbar icon, make the toolbar icon one entry tall (at least in Enterprise), and make the minimum size smaller.
- [x] Dismiss the repo selection window when clicking outside of it
- [ ] The cmd-enter to accept language is small and low contrast, but we are using the theme colors; consider using autosave on close @danielmarquespt
- [ ] Add "Cody Context" entry as part of the [Search Everywhere](https://www.jetbrains.com/help/idea/searching-everywhere.html)
- [ ] Add "Cody Context" as part of the Keymap list on JetBrains settings
- [ ] Try to switch to Action
- [ ] Make tree view reflect language below
- [ ] Show the "inferred" repo, exact UX TBD @danielmarquespt
- [ ] Consider switching to ActionManager.createActionToolbar to get more toolbar padding like the git log viewer. Note sporadic problem with the toolbar button hot zone being the right hand side of the button.
- [ ] When individual repos are un/checked, reflect that in the count in the summary line ("Chat Context X repos ...")
- [ ] #1318
- [ ] Make it accept full URLs and strip the https://
```

Tooltip text:

```
About chat context
Select up to 10 repositories from [customer_instance_url] to use as reference context in this chat.
 
[Context Docs](https://sourcegraph.com/docs/cody/core-concepts/context)
```

Tree view language:

```
Chat Context X repos on [instance]
    Repositories
         example/example
         example/example2
```

danielmarquespt:
Added `Make it accept full URLs and strip the https://` as a user just faced this problem and was unsure why the repo was not being accepted

kalanchan:
waiting on review

toolmantim:
I've jumped in here and done a review at the request of @dominiccooney so he can get started on the problems that were raised in the call last night, as he needed to get started on the eng side. Apologies for the dump here, just wanting to get them out quickly…

---

1. The treeview items should match what’s added in the edit textarea 1:1 (aside from duplicates)
    - e.g. if you add “dev.azure.com/sourcegraph-mike/Test Project/repo1“ on s2, it doesn’t appear in the treeview at all with no feedback as to why
[...]

PR #1479, opened 2024-05-10
Polish the enterprise enhanced context panel

dominiccooney:
Changes to the enterprise enhanced context panel:

- Pluralizes repo/repos in the repo tree view summary line
- Updates counts in the repo tree view summary line
- Mentions the number of ignored repos in the repo tree view summary line (if any)
- Adds a help tooltip with a link to documentation
- Removes the help button from the right-hand side toolbar
- Clicking outside the repo picker popup accepts the edits instead of cancelling them
- Alt-tabbing while the repo picker popup is open does not dismiss it

Part of #1322

## Test plan

Tested manually.

1. Run with `CODY_JETBRAINS_FEATURES=cody.feature.internals-menu=true`
2. Sign into an Enterprise account
3. Verify the repo picker is collapsed, one line and the edit pencil is visible
4. Hover the repo picker and verify the tooltip, opening the link goes to context documentation

Screenshot 2024-05-10 at 20 23 52

5. Add repos and verify the summary line language updates, pluralization is correct, etc.
6. Use Internals, Testing: Cody Ignore and ignore some repositories, check the "ignored" language updates, etc.

Screenshot 2024-05-10 at 20 25 19

7. Verify the details of adding repositories with the pencil:
    - Change text, press ESC, repo list should *not* update
    - Change text, press cmd-enter, repo list should update
    - Change text, click outside popup, repo list *should* update (new behavior)

8. Open the pencil popup, verify that cmd-tab/alt-tab away from IntelliJ and back does not close the popup


pkukielka (approved):
LGTM

src/main/resources/CodyBundle.properties
```diff
@@ -90,17 +90,25 @@ context-panel.remote-repo.error-resolution-failed.title=Repository resolution fa
...
+context-panel.tree.help-tooltip.description=Select up to {0} repositories from {1} to use as reference context in this chat.
+context-panel.tree.help-tooltip.link.text=Context documentation
+context-panel.tree.help-tooltip.link.href=https://sourcegraph.com/docs/cody/core-concepts/context
 messages-panel.welcome-text=Hello! I'm Cody. I can write code and answer questions for you. See [Cody documentation](https://sourcegraph.com/docs/cody) for help and tips.
 EndOfTrialNotification.link-action-name=Setup Payment Info
 EndOfTrialNotification.do-not-show-again=Don't show again
 TrialEndingSoonNotification.ignore=cody.ignore.notification.trial-ending-soon
 TrialEndingSoonNotification.ending-soon.title=Your Cody Pro trial is ending soon
-TrialEndingSoonNotification.ending-soon.content=Setup your payment information to continue using Cody Pro, you won't be charged until February 21.
+TrialEndingSoonNotification.ending-soon.content=Set up your payment information to continue using Cody Pro, you won't be charged until February 21.
```
mkondratek, line 111: not for this PR but we may be able to get rid of `TrialEndingSoonNotification`

mkondratek (commented):
I just tried it on macOs. Things work ok according to your Test Plan. 

---

I noticed this. 
1. Hover on "Chat Context X Repos"
2. "About chat context" tooltip appears
3. Hover on a repository
4. Both "About chat context" and repository url (or info about included repo)
image

Is it ok that we present these two tooltips at once?

mkondratek (approved):
I tried it on windows too - works 🚀

dominiccooney:
> Is it ok that we present these two tooltips at once?

@mkondratek yeah that's not ideal... @danielmarquespt wanted the tooltip to just trigger for the top row, actually. But that will take some work because the tree view items' tooltips are only text.

Another problem with this is the positioning gets in the way of the treeview expand/collapse disclosure arrow.

@toolmantim is working on a refresh of this UI, so this may not live long.

If it is still an issue after the refresh I'll work on custom triggering. It is cheesy but maybe we can watch the mouse movement and push the tooltip in and out of the tooltip properties when the mouse is in the "hot zone" on the top row. Needs investigation whether that will hide the tooltip once it is visible.

PR #1496, opened 2024-05-13
Do not pass tokens, extra headers to non-endpoint URLs.

dominiccooney:
Fixes sourcegraph/security-issues#388

Note, this treats the host as the principal's identity, not scheme+host+port.

Note, in the case of redirects (eg HTTP/302), we will *not* pass credentials to the redirected URL, *even if it is the same principal.*

## Test plan

1. Log into sg02.sourcegraph.com
2. Settings, change the profile URL to http://localhost:1234/foo/bar
3. Run `mitmproxy -p 8123 --mode reverse:https://sourcegraph.com:443`
4. Run JetBrains, log in to sg02
5. Open the Cody settings panel which also lists accounts
6. Consult mitmproxy logs. Requests should *not* include `Authorization: token sgp_...` etc.
7. Finally, test that Cody settings, plus, log in to Sourcegraph.com is not broken and the avatar is displayed; log in to s2 and verify avatars are displayed; clear the profile URL setting from s02 and verify the default avatars is displayed; etc.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/api/SourcegraphApiRequestExecutor.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/api/SourcegraphApiRequestExecutor.kt
@@ -11,16 +11,23 @@ import com.intellij.util.concurrency.annotations.RequiresBackgroundThread
 import com.intellij.util.io.HttpRequests
 import com.intellij.util.io.HttpSecurityUtil
 import com.intellij.util.io.RequestBuilder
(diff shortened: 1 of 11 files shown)


dominiccooney:
@shivasurya I'm using the hostname as the security principal here, is that what we want or should we do something narrower?

shivasurya (approved):

pkukielka (approved):
LGTM

PR #1501, opened 2024-05-14
Boolean response type for textDocument/edit

odisseus:
## Test plan

Test manually and check the trace log. The response for `textDocument/edit` should be either `true` or `false`.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
@@ -65,7 +65,13 @@ class CodyAgentService(project: Project) : Disposable {
       }
 
       agent.client.onTextDocumentEdit = Function { params ->
-        FixupService.getInstance(project).getActiveSession()?.performInlineEdits(params.edits)
+        try {
+          FixupService.getInstance(project).getActiveSession()?.performInlineEdits(params.edits)
+          true
+        } catch (e: RuntimeException) {
+          logger.error(e)
+          false
+        }
       }
 
       agent.client.onTextDocumentShow = Function { params ->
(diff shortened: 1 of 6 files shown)


pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -64,7 +64,7 @@ class CodyAgentService(project: Project) : Disposable {
         FixupService.getInstance(project).getActiveSession()?.performWorkspaceEdit(params)
       }
 
-      agent.client.onTextDocumentEdit = Consumer { params ->
+      agent.client.onTextDocumentEdit = Function { params ->
         FixupService.getInstance(project).getActiveSession()?.performInlineEdits(params.edits)
```
pkukielka: I would maybe also try to catch exception there, just to be sure, although I see you replaced the obvious ones already so it might be good to go.

odisseus: Good point

umpox:
@odisseus @pkukielka Awesome thanks for catching this.

This does mean we run into the error that I mentioned on Slack too. Maybe that should be a follow-up fix though.

To reproduce:
1. Run "Generate Tests" on a function where this is already a test file
2. Notice that Cody starts generating and then immediately throws an error

It seems that we're using an incorrect range for some reason 🤔  I wonder if there's some race condition where we attempt the next edit before the document has updated with the latest range.

I will see if I can replicate this with a test in the Agent, but curious if you have any ideas what might be happening here 

https://github.com/sourcegraph/jetbrains/assets/9516420/5c50accb-ef18-4a39-a206-b742cdc61a43

umpox (approved):

PR #1502, opened 2024-05-14
Adding a polling method to check proxy updates and restart the agent 

arafatkatze:
## Test plan
Test by building locally


pkukielka (approved):
LGTM

PR #1503, opened 2024-05-14
Delete some dead code and assets

dominiccooney:
In particular, remove the "prompter" which predates PromptString and doesn't track the origin of the prompt. (And was unused.)

Also remove a handful of other unused classes, methods and assets.

## Test plan

Tested manually: Chat, generate an autocompletion.


pkukielka (approved):
LGTM

PR #1505, opened 2024-05-14
Fixing bug in reading for cases where http proxy is not turned on

arafatkatze:
## Test plan
Tested in development


pkukielka (approved):
LGTM

mkondratek (approved):
tested. works fine now ! thanks

PR #1509, opened 2024-05-14
Edit: Return Boolean response for workspace/edit

umpox:
## Description

Follow up on https://github.com/sourcegraph/jetbrains/pull/1501

## Test plan

Test manually and check the trace log. The response for workspace/edit should be either `true` or `false`.


odisseus (approved):

pkukielka (approved):
LGTM

PR #1511, opened 2024-05-15
Updating proxy settings to support http tunneling for https connections

arafatkatze:
Without setting both http and https proxy to have the prefix of http we cannot make requests to instances which are https that are routed through the http proxy to do http tunneling
If we don't make this change we get 
```
2024-05-15 12:45:45,615 [ 151442]   WARN - #c.s.c.a.CodyAgentClient - Cody by Sourcegraph: █ GraphQLTelemetryExporter: telemetry: failed to evaluate server version: Error: accessing Sourcegraph GraphQL API: FetchError: request to https://demo.sourcegraph.com/.api/graphql?SiteProductVersion failed, reason: write EPROTO C03ACAF601000000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:355:
 (https://demo.sourcegraph.com/.api/graphql?SiteProductVersion)
```
And in chat we see

```
Cody encountered an error when processing your message: 
⚠ org.eclipse.lsp4j.jsonrpc.ResponseErrorException: Request chat/new failed with message: No chat model found in server-provided config
If the problem persists, please create a support ticket.
```

Or we see `
⚠ write EPROTO 280A0000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:c:\ws\deps\openssl\openssl\ssl\record\ssl3_record.c:355:`

## Test plan
Tested in development

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
@@ -174,7 +174,7 @@ private constructor(
       }
       if (proxy.USE_HTTP_PROXY) {
         processBuilder.environment()["HTTP_PROXY"] = "http://$proxyUrl"
-        processBuilder.environment()["HTTPS_PROXY"] = "https://$proxyUrl"
+        processBuilder.environment()["HTTPS_PROXY"] = "http://$proxyUrl"
       }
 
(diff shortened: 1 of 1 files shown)


pkukielka (approved):
LGTM

PR #1526, opened 2024-05-16
Do not remove unchecked enhanced repos from the list

pkukielka:
## Changes

Unchecking the enhanced repo checkbox does not remove it from the list

## Test plan

Squirrel test from testplan


kalanchan:
I think the uncheck’d boxes feel more natural, but now the “click to enter” remote repositories feel weird when the boxes are unchecked, then checked again on dismiss :sweat_smile: 

https://github.com/sourcegraph/jetbrains/assets/51868853/b1e54806-105e-4513-9197-8b19c6733b27

@danielmarquespt thoughts?

kalanchan (approved):
Approving this to unblock the release. I acknowledge there are a few UX quirks around the repo selector, but we can fix this in a subsequent release afterwards. We don't have to make this version public

cc @danielmarquespt

PR #1537, opened 2024-05-17
Scroll to the bottom on submit in the chat (fixes #340)

### LINKED ISSUE #340: We should auto-scroll to the bottom when submitting a new message
To reproduce:

- Send message with long reply
- Scroll to the top
- Submit new message
- Notice that you have to manually scroll to the bottom to see the new streaming reply

mkondratek:
Fixes #340

## Test plan
1. having a few messages in the chat (so that it takes the whole height)
2. scroll up (so that the view does not touch the bottom) 
3. send a chat message

Expected:
view goes down and autoscrolls

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
@@ -80,11 +80,15 @@ class ChatPanel(
 
   @RequiresEdt
   fun addOrUpdateMessage(message: ChatMessage, index: Int) {
-    if (messagesPanel.componentCount == 1) {
+    val numberOfMessagesBeforeAddOrUpdate = messagesPanel.componentCount
+    if (numberOfMessagesBeforeAddOrUpdate == 1) {
       llmDropdown.updateAfterFirstMessage()
       promptPanel.updateEmptyTextAfterFirstMessage()
     }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/ui/ChatPanel.kt
```diff
@@ -80,11 +80,15 @@ class ChatPanel(
...
   @RequiresEdt
   fun addOrUpdateMessage(message: ChatMessage, index: Int) {
-    if (messagesPanel.componentCount == 1) {
+    val numberOfMessagesBeforeAddOrUpdate = messagesPanel.componentCount
+    if (numberOfMessagesBeforeAddOrUpdate == 1) {
       llmDropdown.updateAfterFirstMessage()
       promptPanel.updateEmptyTextAfterFirstMessage()
     }
     messagesPanel.addOrUpdateMessage(message, index)
+    if (numberOfMessagesBeforeAddOrUpdate < messagesPanel.componentCount) {
```
pkukielka, line 89: In which situation that if may not evaluate to `true`?

mkondratek: For message update

pkukielka (approved):
LGTM

PR #1544, opened 2024-05-17
add cody ignore testing instructions

kalanchan:
Testing instructions for Cody Ignore. We can make additional refinements to the test steps if needed, but we should hand these off to QA to start

## Test plan
N/A


taylorsperry (approved):
LGTM for now, but I'll step through on Monday and ask questions/make suggestions if I have 'em!

PR #1545, opened 2024-05-17
Fixing Intellij Bug to support socks proxy edge case

arafatkatze:
## Test plan
Tested in development locally

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
@@ -172,7 +172,7 @@ private constructor(
       if (proxy.PROXY_TYPE_IS_SOCKS) {
         processBuilder.environment()["HTTP_PROXY"] = "socks://$proxyUrl"
       }
-      if (proxy.USE_HTTP_PROXY) {
+      else {
         processBuilder.environment()["HTTP_PROXY"] = "http://$proxyUrl"
         processBuilder.environment()["HTTPS_PROXY"] = "http://$proxyUrl"
       }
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -172,7 +172,7 @@ private constructor(
       if (proxy.PROXY_TYPE_IS_SOCKS) {
         processBuilder.environment()["HTTP_PROXY"] = "socks://$proxyUrl"
       }
-      if (proxy.USE_HTTP_PROXY) {
+      else {
```
pkukielka: I do not understand this change, old code looks valid to me and proposed change seems to be incorrect.
Could you please elaborate on the proposed logic change?

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -172,7 +172,7 @@ private constructor(
       if (proxy.PROXY_TYPE_IS_SOCKS) {
```
pkukielka: That code is not correct. If you switch proxy type to socks, and then disable proxy, `PROXY_TYPE_IS_SOCKS` is still set to `true` (I checked in debugger), but it should not be passed to agent.
Correct implementation would be:

```kt
 if (proxy.USE_HTTP_PROXY) {
    if (proxy.PROXY_TYPE_IS_SOCKS) {
        processBuilder.environment()["HTTP_PROXY"] = "socks://$proxyUrl"
    } else {
        processBuilder.environment()["HTTP_PROXY"] = "http://$proxyUrl"
        processBuilder.environment()["HTTPS_PROXY"] = "http://$proxyUrl"
    }
}

pkukielka (approved):
LGTM

PR #1551, opened 2024-05-20
Add support for providing visible document ranges

pkukielka:
## Changes

Add support for visible ranges.
See also: https://github.com/sourcegraph/cody/pull/4226

## Test plan

I'm not able to find scenario in which visible range wins over smart selection (see [selection.ts:43](https://github.com/sourcegraph/cody/blob/main/vscode/src/commands/context/selection.ts#L43)) but I verified in the debugger that it works as expected, and it might fix some corner case issues we observed with not perfect document context.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -1,16 +1,20 @@
 package com.sourcegraph.cody.agent.protocol
 
+import com.intellij.codeInsight.codeVision.ui.popup.layouter.bottom
+import com.intellij.codeInsight.codeVision.ui.popup.layouter.right
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.vfs.VirtualFile
+import java.awt.Point
 import java.nio.file.FileSystems
 import java.util.Locale
 
 class ProtocolTextDocument
 private constructor(
-    var uri: String,
-    var content: String?,
-    var selection: Range?,
+    val uri: String,
+    val content: String?,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -32,6 +36,15 @@ private constructor(
       return Range(position, position)
     }
 
+    private fun getVisibleRange(editor: Editor): Range {
```
mkondratek, line 39: 🚀

mkondratek (approved):
lgtm

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -32,6 +36,15 @@ private constructor(
...
+      val visibleArea = editor.scrollingModel.visibleArea
+      val startOffset = editor.xyToLogicalPosition(visibleArea.location)
+      val endOffset = editor.xyToLogicalPosition(Point(visibleArea.right, visibleArea.bottom))
```
odisseus, line 42: The [javadoc](https://github.com/JetBrains/intellij-community/blob/42524aed5922e910f76c77113a8c6f6893b2803a/platform/editor-ui-api/src/com/intellij/openapi/editor/LogicalPosition.java#L14) for `LogicalPosition` says that it can represent a position which doesn't exist in the file (column number larger than the current line length). Are you sure the server would handle that correctly?

pkukielka: Maybe it can represent it, but here you have the opposite situation: you get the `LogicalPosition` based on range of the visible code, so there is no such problem.

PR #1562, opened 2024-05-21
Fix for visible ranges bounds

pkukielka:
## Test plan

Tested with the debugger, but opening file with empty sceen space after the end of the file, and verifying that end range is not after the real end of the file.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -8,6 +8,8 @@ import com.intellij.openapi.vfs.VirtualFile
 import java.awt.Point
 import java.nio.file.FileSystems
 import java.util.Locale
+import kotlin.math.max
+import kotlin.math.min
 
 class ProtocolTextDocument
 private constructor(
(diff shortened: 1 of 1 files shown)


mkondratek (approved):

odisseus:
Do I understand correctly that if the first line (from the beginning to the first newline) of the document is folded and takes 10 lines on the screen, the character preceding that newline has line 10 as its [_logical_ position](https://github.com/JetBrains/intellij-community/blob/master/platform/editor-ui-api/src/com/intellij/openapi/editor/LogicalPosition.java), but line 1 as its [_visual_ position](https://github.com/JetBrains/intellij-community/blob/master/platform/editor-ui-api/src/com/intellij/openapi/editor/VisualPosition.java)? Such naming is mighty confusing to me.

odisseus (approved):

pkukielka:
> Do I understand correctly that if the first line (from the beginning to the first newline) of the document is folded and takes 10 lines on the screen, the character preceding that newline has line 10 as its [_logical_ position](https://github.com/JetBrains/intellij-community/blob/master/platform/editor-ui-api/src/com/intellij/openapi/editor/LogicalPosition.java), but line 1 as its [_visual_ position](https://github.com/JetBrains/intellij-community/blob/master/platform/editor-ui-api/src/com/intellij/openapi/editor/VisualPosition.java)? Such naming is mighty confusing to me.

You are correct, and that would be another bug on my side :) Fixed.

Folded lines generally makes the whole concept a bit broken, in such case we should be sending multiple visible ranges instead. But since usage of visible ranges is minimal anyway I'm not sure it makes sense to overcomplicate it for now.

odisseus:
Well, as far as I understand, sending the visual position (i. e. line 1 in my example) makes more sense. I don't know how exactly the agent processes these data, but to my understanding it should neither know nor care about the presentation of long lines in the client.

pkukielka:
> Well, as far as I understand, sending the visual position (i. e. line 1 in my example) makes more sense. I don't know how exactly the agent processes these data, but to my understanding it should neither know nor care about the presentation of long lines in the client.

Problem looks like that:

```
 Logical     Visual     Folded?     Code
Position    Position   
-----------------------------------------------
        1         1         no       function1 {
        2         -         yes        body1
        3         -         yes      }
        4         2         no       function2 {
        5         -         yes        body2
        6         -         yes      }
```

Now what you see in the editor are two lines (visual positions 1 and 2). Let's say they are function names, and the folded lines are their bodies.
If you select them, and send request to Cody to explain the code, what should it explain?
I would say both functions. If you send visible range as `[1,2]` it will be clearly incorrect, as that will be just part of the `function1`. 
`[1-6]` is more correct, although it also passes lines which were not really visible to the user.
Proper solution would be to send `[1-1], [4-4]`, but that would require sending multiple ranges and implementation of that would be more complex.

odisseus (approved):

PR #1564, opened 2024-05-21
Improve chat state management

pkukielka:
Fixes https://github.com/sourcegraph/cody-issues/issues/262

## Changes
1. Duplicated message no longer shows in the chat panel
2. It is possible to cancel question even if agent is down
3. Various scenarios with cancellation and retry are handled better now

#### Scenario 1:
1. Open Cody chat panel
4. Kill Cody agent process
5. Quickly type something in the chat
6. No duplicated message should appear

#### Scenario 2:
1. Open Cody chat panel
2. Kill Cody agent process
3. Quickly type something in the chat
4. Click "Stop generating"
7. Chat should abort further actions and allow you to type again
8. Send new question to Cody and allow it to finish
9. Messages should show in the right order (the one which was not sent to the agent will disappear)


mkondratek (approved):
lgtm

Issue #1571, opened 2024-05-21
JetBrains: Crash occurred while running "Generate Unit Test" command.


jay-fibi:
### Cody Version

v5.5.12-nightly

### IDE Information

IntelliJ IDEA 2023.3.4 (Community Edition)
Build #IC-233.14475.28, built on February 13, 2024
Runtime version: 17.0.10+1-b1087.17 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Windows 11.0
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 12
Registry:
ide.experimental.ui=true
Non-Bundled Plugins:
de.sgalinski.typoscript.enterprise.plugin.id (2.8.0)
org.jetbrains.plugins.hocon (2023.1.0)
PythonCore (233.14475.9)
com.soywiz.korge.korge-intellij-plugin (4.0.4)
com.sourcegraph.jetbrains (5.5.12-nightly)
Kotlin: 233.14475.28-IJ

### Describe the bug

Platform: Windows

**Steps to reproduce:**

1. Have Cody plugin installed in your JB IntelliJ editor
2. Login to Cody with Enterprise user
3. Open any source code file and select code snippet
4. Right click to open context menu and choose "Cody > Generate Unit Tests"

### Observed behavior:
Crash occurred while running the command with following stacktrace,
Stacktrace:
```text
java.lang.RuntimeException: Cannot invoke (class=CodyFileEditorListener, method=fileOpened, topic=FileEditorManagerListener)
	at com.intellij.util.messages.impl.MessageBusImplKt.invokeListener(MessageBusImpl.kt:677)
	at com.intellij.util.messages.impl.MessageBusImplKt.deliverMessage(MessageBusImpl.kt:423)
	at com.intellij.util.messages.impl.MessageBusImplKt.pumpWaiting(MessageBusImpl.kt:402)
	at com.intellij.util.messages.impl.MessageBusImplKt.access$pumpWaiting(MessageBusImpl.kt:1)
	at com.intellij.util.messages.impl.MessagePublisher.invoke(MessageBusImpl.kt:461)
	at jdk.proxy2/jdk.proxy2.$Proxy46.fileOpened(Unknown Source)
	at com.intellij.openapi.fileEditor.impl.FileEditorManagerImpl$openFileInEdt$3.invokeSuspend(FileEditorManagerImpl.kt:2205)
[...]

PR #1574, opened 2024-05-22
Added fine-grained error handling and logging to CodyFileListener

steveyegge:
This logging will help us identify the root cause of [Issue 1571](https://github.com/sourcegraph/jetbrains/issues/1571). Also, we will no longer show a big crash dump to the user when any `CodyFileListener` method encounters an error.

## Test plan

We're going to ask the tester who reported it to try to reproduce it with this new logging in place. The error should go away, and if there is any bad behavior, we can check the logs to see more detailed stack traces.


steveyegge:
closed because it had some weird unrelated commit

PR #1575, opened 2024-05-22
Added finer-grained error handling and logging for issue 1571

steveyegge:
Can't see the root cause in the stack trace with Issue 1571, so I've added finer-grained exception handling and logging to `CodyFileEditorFactory`.

## Test plan

Will ask @jay-fibi to repro after we put this fix in.

- If it fixes the error, great! 
- If there is still bad behavior, we can now check the logs to get the root cause.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/listeners/CodyFileEditorListener.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/listeners/CodyFileEditorListener.kt
@@ -1,6 +1,7 @@
 package com.sourcegraph.cody.listeners
 
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.editor.EditorFactory
 import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.fileEditor.FileEditorManager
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/listeners/CodyFileEditorListener.kt
```diff
@@ -12,39 +13,60 @@ import com.sourcegraph.cody.agent.CodyAgentService.Companion.withAgent
...
-          val file = fileDocumentManager.getFile(editor.document)
-          if (file != null) {
-            val textDocument = fromVirtualFile(editor, file)
-            codyAgent.server.textDocumentDidOpen(textDocument)
+          fileDocumentManager.getFile(editor.document)?.let { file ->
+            try {
+              val textDocument = fromVirtualFile(editor, file)
+              codyAgent.server.textDocumentDidOpen(textDocument)
+            } catch (x: Exception) {
+              logger.warn("Error calling textDocument/didOpen for file: ${file.path}", x)
```
pkukielka, line 58: Technically it can also be error from `fromVirtualFile` ;)

pkukielka (approved):
LGTM

PR #1576, opened 2024-05-22
sort offsets before creating range markers in replace operations

steveyegge:
In issue 1570, the start offset was greater than the end offset in a replace action.

Given that it is not reliably reproducible (as far as I know -- I couldn't repro it), it will be tricky to track down where they are getting switched, as it could be happening in the agent or in the client, in any of multiple places.

Given that start and end were `start=653; end=616`, this may well have been a correct range with the offsets reversed, in which case sorting the offsets "undoes" the bug before performing the edit operation.

If, however, this is just a "bad range", and we are replacing something invalid, then it should fail much more gracefully than it does now, even after sorting the offsets to create potentially yet another bad range. First, it won't throw a big exception, and second, the results will be both visibly wrong (and thus easy to report accurately) and Undoable (so it's a better user experience when it fails this way.)

## Test plan

Should test Show Diff thoroughly, and the reporter for issue 1570 should validate that it is fixed.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/fixupActions/ReplaceUndoableAction.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/fixupActions/ReplaceUndoableAction.kt
@@ -22,25 +22,24 @@ class ReplaceUndoableAction(
       other: ReplaceUndoableAction,
       document: Document
   ) : this(other.project, other.edit, document) {
-    this.beforeMarker =
-        other.beforeMarker?.let { document.createRangeMarker(it.startOffset, it.endOffset) }
-    this.afterMarker =
-        other.afterMarker?.let { document.createRangeMarker(it.startOffset, it.endOffset) }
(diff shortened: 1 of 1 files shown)


pkukielka:
@steveyegge good catch with an inverted range!
I'm a bit worried about just automatically reverting it though.

Especially if we are not sure if they are really inverted or if this is bug of other sort.
We can introduce all kind of messed up text in the documents, and not all of it might be noticed by user at the first glance (some of ti might be minor and subtle, or even - in the worst case - still compile).
That could lead to braking user code without their knowledge, which would be really bad.
TBH I would prefer to log all the details and fail.

That said I tried to come up with hypothesis why it could be failing (I thought about document/ranges going out of sync between agent and client) and while describing it and looking into a code I think I found much more probable root cause - a bug (which I think I introduced). Let me raise a PR with a fix and link it here when ready.

pkukielka:
I think that might be a proper fix: https://github.com/sourcegraph/jetbrains/pull/1581

PR #1585, opened 2024-05-22
Fix error of project alredy disposed

pkukielka:
## Test plan

Quickly switch projects multiple times to reproduce the issue.


mkondratek (approved):
whoops, rebase needed

PR #1601, opened 2024-05-22
Update README.md for new Cody messaging

iskyOS:
Updates the README for new Cody messaging (available in Notion). Also drop beta language.

## Test plan

README updates only.


iskyOS:
@taylorsperry Can I get your review on README changes?

taylorsperry (approved):

taylorsperry:
Don't want to block this PR, but do you think we should add info about the new inline edit features in a follow-up? I also wonder if we should de-emphasize Code Search since we don't have anyone supporting it now.

PR #1603, opened 2024-05-23
Reinstated the client-side logic to auto-accept on document edits

steveyegge:
We tracked down a class of bugs, possibly multiple classes, including [Issue 315](https://github.com/sourcegraph/cody-issues/issues/315), to a race condition in which the Agent didn't realize we were still applying its edits, and we notified it of the text change, and it canceled the operation.

The user-visible result is usually that the Accept lens group just never shows up. The doc string is generated and then it's done. You can use system Undo but that's it. It's accepted. Which we don't want yet.

Only the client knows when it is safe to do auto-accept without accidentally cancelling the entire session early.

This PR is not the ideal long-term solution; ideally it would be decided in the Agent. But the client has to notify the agent that it's done, etc. and that's potentially destabilizing for GA, whereas this fix was very simple and clean.

## Test plan

Everyone has try to break it -- specifically:

- the Accept lens group (with Edit/Retry, Undo, Show Diff) should _always_ be shown after requesting a Document Code
- It should auto-accept your current edit if you start another one somewhere else in the document.

This is already covered by the test plan, so it just needs to be re-run with this specific feature in mind.


src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -225,6 +227,10 @@ abstract class FixupSession(
 
   private fun showAcceptGroup() {
     showLensGroup(LensGroupFactory(this).createAcceptGroup())
+
+    // This is the specific moment after which any edits to the document,
+    // including edits generated by the Agent, will result in an auto-accept.
+    documentListener.setAcceptLensGroupShown(true)
```
abeatrix, line 233: not sure if it'd help but maybe we could add a check to verify if the task.state is `CodyTaskState.Applied` before we add the listener? Not a blocker though!

abeatrix (approved):
not familiar with kt but the change makes sense to me. Thank you!

PR #1604, opened 2024-05-23
Indent code lenses to the line below them

steveyegge:
@danielmarquespt expressed that he would prefer to have the lenses indent to the same level as the next non-blank line below them, rather than always at a fixed offset.

This was very easy to add, and low-risk, so I added it. It handles errors and falls back to the old indentation if there are any problems at all.

## Test plan

I updated the test plan to include a visual inspection for the lens indentation.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/widget/LensWidgetGroup.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/widget/LensWidgetGroup.kt
@@ -26,6 +26,7 @@ import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.ui.UIUtil
 import com.sourcegraph.cody.agent.protocol.Range
 import com.sourcegraph.cody.edit.sessions.FixupSession
+import org.jetbrains.annotations.NotNull
 import java.awt.Cursor
 import java.awt.Font
 import java.awt.FontMetrics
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/sourcegraph/cody/edit/widget/LensWidgetGroup.kt
```diff
@@ -250,6 +252,39 @@ class LensWidgetGroup(val session: FixupSession, parentComponent: Editor) :
...
+        // Compute the pixel width of the indentation.
+        if (lineText.isNotBlank()) {
+          val tabSize = EditorUtil.getTabSize(editor)
+          val spaceWidth = EditorUtil.getSpaceWidth(Font.PLAIN, editor)
+          val indentationLevel =
+              lineText
+                  .takeWhile { it.isWhitespace() }
+                  .sumOf { if (it == '\t') tabSize * spaceWidth else spaceWidth }
+
+          return indentationLevel
```
mkondratek, line 289: how often this calculations happen? is it on every UI update? we should be careful and avoid heavy operations on AWT (I am not saying it's heavy - just asking). 

@steveyegge, can you verify/confirm it's safe? 🙏

steveyegge: Great callout @mkondratek -- it was getting called a lot. I cached it and it's only called when it changes now.

mkondratek (approved):
lgtm

danielmarquespt (approved):

PR #1607, opened 2024-05-23
Document synchronization: fix bugs and improve performance

olafurpg:
This PR makes several changes to document synchronization:

* Caret changes now no longer send the full text contents. This was redundant since the text hadn't changed.
* Selection changes. Same as caret changes, no more redundant content traffic.
* Document changes. We still do full document synchronization, but we do it unconditionally now. There were cases where we disabled synchronization (undo in progress, and autocomplete was disabled).

These improvements should both address performance issues as well as correctness bugs (esp. disabling of synchronization during document changes).

Co-Author: @RXminuS 

## Test plan

Manually tested with the command

```
❯ tail -f build/sourcegraph/cody-agent-trace.json  | grep -A 30 textDocument/didChange
```
* Make changes to the caret/selection and observe that the content is no longer populated
* Disable autocomplete, and confirm that we still send `textDocument/didChange` notifications (we don't do this on main).

I have not tried yet to reproduce the undo behavior.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -3,20 +3,23 @@ package com.sourcegraph.cody.agent.protocol
 import com.intellij.codeInsight.codeVision.ui.popup.layouter.bottom
 import com.intellij.codeInsight.codeVision.ui.popup.layouter.right
 import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.LogicalPosition
+import com.intellij.openapi.editor.event.DocumentEvent
 import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.vfs.VirtualFile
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -3,20 +3,23 @@ package com.sourcegraph.cody.agent.protocol
...
 class ProtocolTextDocument
 private constructor(
     val uri: String,
-    val content: String?,
-    val selection: Range?,
-    val visibleRange: Range?
+    val content: String? = null,
+    val selection: Range? = null,
+    val visibleRange: Range? = null,
+    val contentChanges: List<ProtocolTextDocumentContentChangeEvent>? = null,
```
olafurpg, line 22: I left this in the diff even if we don't use it yet.

RXminuS (commented):
LGTM but I'll leave approval to another reviewer as I was part of it

pkukielka (approved):
LGTM

PR #1609, opened 2024-05-23
Improve fixup session handling and disposal

mkondratek:
Possible a step towards fixing https://github.com/sourcegraph/cody-issues/issues/314.

## Test plan
1. Try to get to the "invalid state" when the inline actions are disabled

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
@@ -79,20 +79,24 @@ class CodyAgentService(private val project: Project) : Disposable {
       }
 
       agent.client.onWorkspaceEdit = Function { params ->
+        val activeSession = FixupService.getInstance(project).getActiveSession()
         try {
-          FixupService.getInstance(project).getActiveSession()?.performWorkspaceEdit(params)
+          activeSession?.performWorkspaceEdit(params)
           true
         } catch (e: RuntimeException) {
+          activeSession?.dispose()
           logger.error(e)
           false
         }
       }
 
       agent.client.onTextDocumentEdit = Function { params ->
+        val activeSession = FixupService.getInstance(project).getActiveSession()
         try {
-          FixupService.getInstance(project).getActiveSession()?.performInlineEdits(params.edits)
+          activeSession?.performInlineEdits(params.edits)
           true
         } catch (e: RuntimeException) {
+          activeSession?.dispose()
           logger.error(e)
           false
         }
@@ -250,6 +254,7 @@ class CodyAgentService(private val project: Project) : Disposable {
               try {
                 callback.accept(agent)
               } catch (e: Exception) {
+                logger.warn(e)
                 onFailure.accept(e)
               }
             }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -250,6 +254,7 @@ class CodyAgentService(private val project: Project) : Disposable {
               try {
                 callback.accept(agent)
               } catch (e: Exception) {
+                logger.warn(e)
```
mkondratek, line 257: The error: `IllegalStateException("Cannot start new session when one is already active")` was there but we ignored it 🙃

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -411,6 +411,7 @@ abstract class FixupSession(
     if (project.isDisposed) return
     performedActions.forEach { it.dispose() }
     editor.document.removeDocumentListener(documentListener)
+    cancellationToken.dispose()
```
mkondratek, line 403: so the controller unsets active session

pkukielka: We need a bit of additional cleanup:
1. On `finish` and `error` state we now call `cancellationToken.dispose()`
We should either call `dispose()`, or we should move all this actions to `cancellationToken.onFinished` in the constructor.
2. There is also `fun dismiss()` which looks like incomplete `dispose` now. I guess we should delete it and replace with `dispose()`?

mkondratek: fixed

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -88,14 +88,6 @@ abstract class FixupSession(
         } catch (x: Exception) {
           logger.debug("Session cleanup error", x)
         }
-
-        runInEdt {
-          try { // Disposing inlay requires EDT.
-            Disposer.dispose(this)
```
mkondratek, line 94: the sessions calls `cancelationToken.dispose()` now

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -411,6 +399,8 @@ abstract class FixupSession(
     if (project.isDisposed) return
     performedActions.forEach { it.dispose() }
     editor.document.removeDocumentListener(documentListener)
```
mkondratek, line 401: @pkukielka, the problem was that we called it twice - we should not see more errors from this one

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
pkukielka: Cool!

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
pkukielka: Great :)

pkukielka (approved):
LGTM

PR #1611, opened 2024-05-23
Undo: revert change - disable document synchronization during undo

olafurpg:
The PR https://github.com/sourcegraph/jetbrains/pull/1607 introduced a change where we start synchronizing when
`CommandProcessor.getInstance().isUndoTransparentActionInProgress` returns true. This PR reverts this change since we have heard reports that something is acting weirdly around undo after merging this change.

## Test plan

Not tested because I wasn't able to reproduce the behavior of the regression.


pkukielka (approved):
LGTM

PR #1616, opened 2024-05-23
Fix listener logic, bump cody commit

pkukielka:
## Test plan

1. Run Document Code action in the IDE
2. Cancel it before accept lens appear


mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -398,7 +397,14 @@ abstract class FixupSession(
   override fun dispose() {
     if (project.isDisposed) return
     performedActions.forEach { it.dispose() }
-    editor.document.removeDocumentListener(documentListener)
+
+    try {
+      editor.document.removeDocumentListener(documentListener)
+    } catch (e: Throwable) {
```
mkondratek, line 403: it will not help

mkondratek: the error is already logger as internal error by `logger.error()`

pkukielka: Yea you are right. We need to look at it tomorrow then.

abeatrix: is it possible to check if `documentListener` is valid before `editor.document.removeDocumentListener`? or does the error come from somewhere else?

mkondratek: the error is already reported to the IDE within `removeDocumentListener` - we cannot catch it 😕 

I merged this one and I am going to do the release. I will try to fix this issue later.

mkondratek: a few moments later: https://github.com/sourcegraph/jetbrains/pull/1621

PR #1621, opened 2024-05-23
Fix `removeDocumentListener` and a few others

mkondratek:
In the scope of this PR:
- [Do not send chat/new when no account signed in](https://github.com/sourcegraph/jetbrains/pull/1621/commits/0854f838a49796f91209d869d5cfdee92b3994bf)
- [Fix (remove,add)DocumentListener](https://github.com/sourcegraph/jetbrains/pull/1621/commits/7b33025dacfd0a570e7df70e732fb5dee7ebf5e2)
- [Clean up runInEdt calls](https://github.com/sourcegraph/jetbrains/pull/1621/commits/cd645b481c936a07a708c7b4ed3b2d1b90296a37)

### new chat creation 

1. no account
    1. open a project with no account signed in
    2. sign in
    3. write chat
2. an account
    1. open a project with an account signed in
    2. write chat

### Fix document listener issues
1. Document Code
2. Cancel


src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -93,7 +93,8 @@ abstract class FixupSession(
   @Volatile private var undoInProgress: Boolean = false
 
   init {
-    runInEdt { Disposer.register(controller, this) }
+    editor.document.addDocumentListener(documentListener, /* parentDisposable = */ this)
```
mkondratek, line 96: simply as that - we can pass a disposable 🚀

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -214,14 +215,15 @@ abstract class FixupSession(
 
   private fun showWorkingGroup() {
     showLensGroup(LensGroupFactory(this).createTaskWorkingGroup())
+    documentListener.setAcceptLensGroupShown(false)
```
mkondratek, line 218: We can call `showAcceptGroup` and `showWorkingGroup` multiple times in the session - e.g. when retrying or undoing. Hence, we should properly handle the listener setting.

kalanchan (approved):

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
pkukielka: @mkondratek that is not really true ;)
Then you undo session is ended.
Retry will start new edit after an undo.

PR #1624, opened 2024-05-24
fix: handle empty documents and convert 1-based to 0-based line numbers

abeatrix:
zero

CHANGES:

- For an empty document, return a zero-width selection at the start of the document
- Subtract 1 to convert 1-based line numbers to 0-based line numbers before sending to Agent
- Keep current selection as is when sending the document to the agent
- Add selection information to ProtocolTextDocument created from Editor

FIX:
- active cursor without selection would break Cody
- random leftover code at the end of unit test files due to incorrect line number used
- inserting code at the incorrect line

## Test plan

Verify the following works in the supported editor:
- Open the Jetbrains repo in this build
- Put your cursor inside a function and run the Doc & Test command 
- Highlight a function to run the Doc & Testcommand

## Limitation

- Active cursor without selection only works on Editor that has folding range support
- This means it would break Cody if you open a typescript repo in IDE
  - We can look into this in a follow up to prevent Cody from stop working due to error returned by Agent

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -24,7 +24,11 @@ private constructor(
 
   companion object {
 
-    private fun getSelection(editor: Editor): Range {
+    private fun getSelection(editor: Editor, isZeroBased: Boolean = false): Range {
+      // For an empty document, return a zero-width selection at the start of the document.
+      if (editor.document.lineCount == 0) return Range(Position(0, 0), Position(0, 0))
+      // Subtract 1 to convert 1-based line numbers to 0-based line numbers.
+      val baseline = if (isZeroBased) 1 else 0
       val selectionModel = editor.selectionModel
       val selectionStartPosition =
           selectionModel.selectionStartPosition?.let { editor.visualToLogicalPosition(it) }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -69,7 +77,8 @@ private constructor(
     @JvmStatic
     fun fromEditorWithRangeSelection(editor: Editor): ProtocolTextDocument? {
       val file = FileDocumentManager.getInstance().getFile(editor.document) ?: return null
-      return ProtocolTextDocument(uri = uriFor(file), selection = getSelection(editor))
+      // Keep current selection as is when sending the document to the agent.
+      return ProtocolTextDocument(uri = uriFor(file), selection = getSelection(editor, false))
```
olafurpg, line 81: It's often clearer to use named arguments for boolean literals

```suggestion
      return ProtocolTextDocument(uri = uriFor(file), selection = getSelection(editor, isZeroBased = false))
```

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -106,8 +116,9 @@ private constructor(
       return ProtocolTextDocument(
           uri = uriFor(file),
           content = text,
-          selection = getSelection(editor),
-          visibleRange = getVisibleRange(editor))
+          selection = getSelection(editor, true),
```
dominiccooney, line 119: Ditto here about naming the parameter.

You can also use an enum instead of a boolean to make the meaning of the value obvious.

dominiccooney (approved):

olafurpg:
Thank you @abeatrix for looking into this! It turns out that IDEA logical positions are in fact 0-based just like in VS Code. We were able to reproduce a bug related to offset selection (start == end) when triggering "Document Code", and this seems to be a bug on the agent side.

Superseded by #1626

PR #1626, opened 2024-05-24
Tests: add basic test cases for document synchronization

olafurpg:
Previously, it was very easy to introduce regressions to document synchronization because the feedback loop was very slow, you had to manually run IntelliJ and look at raw line/character numbers.

This PR adds a basic test suite so that we can start debugging these kinds of issues more effectively, without having to manually spin up IntelliJ.

## Test plan

Green CI.

See new test cases!


pkukielka (approved):
LGTM!

PR #1628, opened 2024-05-24
Testing: make the agent start during integration tests

olafurpg:
Previously, the agent wasn't starting in our tests because we didn't set the right system property. Now, the agent manages to start correctly during the tests.

While debugging this, I made a few other small improvements:

- Fix .tool-version file so we use the right node/pnpm version. Note that this file had the wrong name (.tool-version instead of .tool-versions)
- Move "Check updates" task off the blocking path during activation. Starting the agent was taking ~2s longer than it should have because we were checking if the Cody plugin had an available update before starting. If you have a very slow network connection, this overhead would have been even worse.

## Test plan

Tested locally. Confirmed that the run 2s faster.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,3 @@
 java adoptopenjdk-11.0.21+9
+nodejs 20.4.0
+pnpm 8.6.7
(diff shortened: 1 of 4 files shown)


RXminuS (approved):
TEEEEESTS!

PR #1633, opened 2024-05-24
Agent: enable testing mode that panics on sync errors

olafurpg:
## Test plan
- Validated that on runIde the agent panics and logs on desync issues

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -296,10 +296,10 @@ tasks {
   // `./gradlew test` or when testing inside IntelliJ
   val agentProperties =
       mapOf<String, Any>(
-          "cody.agent.trace-path" to "$buildDir/sourcegraph/cody-agent-trace.json",
-          "cody.agent.directory" to buildCodyDir.parent,
+          "cody-agent.trace-path" to "$buildDir/sourcegraph/cody-agent-trace.json",
+          "cody-agent.directory" to buildCodyDir.parent,
           "sourcegraph.verbose-logging" to "true",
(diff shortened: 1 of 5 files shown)


build.gradle.kts
```diff
@@ -369,9 +380,13 @@ tasks {
   runIde {
     dependsOn(project.tasks.getByPath("buildCody"))
     jvmArgs("-Djdk.module.illegalAccess.silent=true")
+    // TODO: Figure out why systemProperties(...) doesn't work
```
RXminuS: It's because one was `cody-agent` and the other is `cody.agent`. Fix incoming...

olafurpg: Wow, nice catch (and thanks Cody for suggesting the wrong key 😢 )

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -287,7 +287,7 @@ private constructor(
     }
 
     private fun traceWriter(): PrintWriter? {
-      val tracePath = System.getProperty("cody-agent.trace-path", "")
+      val tracePath = System.getProperty("cody.agent.trace-path", "")
```
olafurpg: We should not change this as we have existing (internal dev) usage of this property (example: in my VM options).

RXminuS: Ok switched it to kebab-case again

RXminuS (approved):

pkukielka (approved):
LGTM

PR #1635, opened 2024-05-24
Send only incremental content updates on the document change

pkukielka:
## Changes

Instead of sending full content we are now sending only modified fragments.
That should:
* decrease amount of traffic
* improve accuracy of the diffs (previously we were calculating them on the agent side, which sometimes cannot be guaranteed to be correct)

## Test plan
- Manually tested all inline edit and cody completion features
- Ran unit tests

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -102,10 +102,12 @@ private constructor(
       val oldFragment = event.oldFragment.toString()
       val file = FileDocumentManager.getInstance().getFile(editor.document) ?: return null
       val startPosition = editor.offsetToLogicalPosition(event.offset).codyPosition()
+      // allocate List once to avoid three unnecessary duplicate allocations
+      val oldFragmentLines = oldFragment.lines()
       val endCharacter =
-          if (oldFragment.lines().size > 1) oldFragment.lines().last().length
+          if (oldFragmentLines.size > 1) oldFragmentLines.last().length
           else startPosition.character + oldFragment.length
-      val endPosition = Position(startPosition.line + oldFragment.lines().size - 1, endCharacter)
+      val endPosition = Position(startPosition.line + oldFragmentLines.size - 1, endCharacter)
 
       return ProtocolTextDocument(
           uri = uriFor(file),
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -4,11 +4,12 @@ import com.intellij.codeInsight.codeVision.ui.popup.layouter.bottom
 import com.intellij.codeInsight.codeVision.ui.popup.layouter.right
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.LogicalPosition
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -71,6 +72,25 @@ private constructor(
       return ProtocolTextDocument(uri = uriFor(file), selection = getSelection(editor))
     }
 
+    @JvmStatic
+    fun fromEditorForDocumentEvent(editor: Editor, event: DocumentEvent): ProtocolTextDocument? {
+      val oldFragment = event.oldFragment.toString()
+      val file = FileDocumentManager.getInstance().getFile(editor.document) ?: return null
+      val startPosition = editor.offsetToLogicalPosition(event.offset).codyPosition()
+      val endCharacter =
```
pkukielka, line 107: For `endCharacter` we cannot compute it using `editor.offsetToLogicalPosition` because document is already changed. Unfortunately IntelliJ gives us only offsets and not lines and columns, so we need to do a bit of calculations ourselves.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -96,6 +97,25 @@ private constructor(
           uri = uriFor(file), selection = getSelection(editor), testing = getTestingParams(editor))
     }
 
+    @JvmStatic
+    fun fromEditorForDocumentEvent(editor: Editor, event: DocumentEvent): ProtocolTextDocument? {
+      val oldFragment = event.oldFragment.toString()
+      val file = FileDocumentManager.getInstance().getFile(editor.document) ?: return null
+      val startPosition = editor.offsetToLogicalPosition(event.offset).codyPosition()
+      val endCharacter =
+          if (oldFragment.lines().size > 1) oldFragment.lines().last().length
```
olafurpg: nit: is this equivalent?
```suggestion
          if (oldFragment.indexOf("\n") >= 0) oldFragment.lines().last().length
```

If so, we can move to a helper function

olafurpg: I see we also use `oldFragment.lines()` below. This PR is already a gigantic perf win so these minor things are only drops in the ocean, but would be nice to avoid allocations as much as possible

src/test/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocumentTest.kt
```diff
@@ -53,6 +55,147 @@ class ProtocolTextDocumentTest : BasePlatformTestCase() {
...
+    EditorChangesBus.addListener { _, textDocument -> lastTextDocument = textDocument }
+
+    myFixture.openFileInEditor(myFixture.createFile("newFile.txt", ""))
+
+    assertEquals("file:///src/newFile.txt", lastTextDocument!!.uri)
+  }
+
+  fun test_documentListener_appendAndPrepend() {
+    var lastTextDocument: ProtocolTextDocument? = null
+    EditorChangesBus.addListener { _, textDocument -> lastTextDocument = textDocument }
```
olafurpg, line 87: I think it's fine to move this to `setUp()` to make each test case a bit smaller.

olafurpg (commented):
None of my comments are blocking, I haven't tested this locally but go ahead and merge!

RXminuS (approved):

PR #1642, opened 2024-05-25
FixupSession: handle exceptions during edit session

abeatrix:
CLOSE https://github.com/sourcegraph/cody-issues/issues/355

- Wrap the entire edit session in a try-catch block to handle any exceptions that may occur
- Log a user-friendly error message when an exception is caught, instead of just canceling the session
- Improve the error handling in the `ensureSelectionRange` function by catching and logging any exceptions that occur when getting the folding ranges without blocking the session from starting
- Add a new error message when an exception occurs during the editing action

The behavior of the FixupSession class has changed to better handle exceptions during the edit session

## Test plan

https://github.com/sourcegraph/jetbrains/assets/68532117/fb83cf15-8a5c-447e-add8-8f1a753ccf25

1. Open a repository in a JetBrains IDE where folding range is not available (e.g Typescript)
2. Place your cursor inside a function
3. Execute an edit command (edit / doc / test)
4. Verify Cody is able to complete the request
  - Before: request will fail and you will not be able to run any edit commands until restart

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
@@ -102,30 +102,33 @@ abstract class FixupSession(
     CodyAgentService.withAgent(project) { agent ->
       workAroundUninitializedCodebase()
 
-      fixupService.startNewSession(this)
+      try {
+        fixupService.startNewSession(this)
+        // Spend a turn to get folding ranges before showing lenses.
+        ensureSelectionRange(agent, textFile)
+        showWorkingGroup()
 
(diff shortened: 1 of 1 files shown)


kalanchan:
nice! the try catch blocks are handling the errors nicely, and prevents the process from crashing. I checked the logs, we're still running into the `Message could not be parsed` problem, but I think this is a really fix for the error handling 

https://github.com/sourcegraph/jetbrains/assets/51868853/35ecdccc-4021-4aca-9e3a-1a137823faf9

kalanchan (approved):
going to approve this for now to help unblock current bugs found in edit commands, feel free to re-review if further changes are needed

odisseus (approved):

steveyegge (approved):
Looks great!

PR #1644, opened 2024-05-25
Fix mouse cursor when it leaves a lens action button

steveyegge:
The mouse cursor turns to a hand when over a `LensAction` button-shaped widget, since it acts like clickable link. It was not restoring properly when the cursor left the button bounds, remaining as a hand. It also had overcomplicated logic. This PR addresses both of those issues.

## Test plan

Manually tested.


pkukielka (approved):
LGTM

PR #1645, opened 2024-05-25
Improve content state synchronisation and testing

pkukielka:
## Test plan

Do any code edit actions in the client - they should not be crashing the agent.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -300,7 +300,7 @@ tasks {
           "cody-agent.directory" to buildCodyDir.parent,
           "sourcegraph.verbose-logging" to "true",
           "cody-agent.panic-when-out-of-sync" to
-              (System.getProperty("cody-agent.panic-when-out-of-sync") ?: "false"),
+              (System.getProperty("cody-agent.panic-when-out-of-sync") ?: "true"),
           "cody.autocomplete.enableFormatting" to
               (project.property("cody.autocomplete.enableFormatting") ?: "true"))
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -300,7 +300,7 @@ tasks {
           "cody-agent.directory" to buildCodyDir.parent,
           "sourcegraph.verbose-logging" to "true",
           "cody-agent.panic-when-out-of-sync" to
-              (System.getProperty("cody-agent.panic-when-out-of-sync") ?: "true"),
+              (System.getProperty("cody-agent.panic-when-out-of-sync") ?: "false"),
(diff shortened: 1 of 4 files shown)


RXminuS (approved):

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -24,8 +26,15 @@ private constructor(
...
 
+  init {
+    if (!ApplicationManager.getApplication().isDispatchThread) {
+      throw IllegalStateException("ProtocolTextDocument must be be created on EDT")
+    }
+  }
+
   companion object {
 
+    @RequiresEdt
```
mkondratek, line 37:
```
/**
 * Methods and constructors annotated with {@code RequiresEdt} must be called from the Event Dispatch Thread only.
 * Parameters annotated with {@code RequiresEdt} must be callables and are guaranteed to be called from the Event Dispatch Thread.
 *
 * <p/>Aside from a documentation purpose, the annotation is processed by the {@link org.jetbrains.jps.devkit.threadingModelHelper}.
 * The plugin instruments annotated elements with {@link Application#assertIsDispatchThread()} calls
 * to ensure annotation's contract is not violated at runtime. The instrumentation can be disabled
 * by setting {@link RequiresEdt#generateAssertion()} to {@code false}.
 *
 * <p/> <b>Important:</b> the instrumentation has limitations. Please read the docs
 * of the {@link org.jetbrains.jps.devkit.threadingModelHelper} to learn about them.
 *
 * @see <a href="http://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/general_threading_rules.html">General Threading Rules</a>
 * @see Application#assertIsDispatchThread()
 * @see Application#invokeLater(Runnable, ModalityState)
 * @see Application#invokeAndWait(Runnable, ModalityState)
 */
@Retention(RetentionPolicy.CLASS)
@Target({ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.PARAMETER})
public @interface RequiresEdt {
  /**
   * @return {@code false} if annotated element must not be instrumented with the assertion.
   */
  boolean generateAssertion() default true;
}
```

nice 🚀

build.gradle.kts
RXminuS, line 1: Let's set this to false for now until we are confident there's no more false positives etc. Last thing want devs to do is add a false override because it is annoying and then never turn it back on once we've fixed things.

PR #1647, opened 2024-05-25
improvements to the error handling when the Agent encounters errors

steveyegge:
We show the Error lens group more often now, rather than hiding the error. We also automatically restart now if we discover that the Agent is down; there was a code path where the Agent wouldn't restart and Cody would just hang.

## Test plan

We don't have an effective way to test generating errors until integration testing lands. For now, manual testing.


src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -95,37 +95,39 @@ abstract class FixupSession(
...
-          .exceptionally { error: Throwable? ->
-            if (!(error is CancellationException || error is CompletionException)) {
-              showErrorGroup("Error while generating code: ${error?.localizedMessage}")
+            .exceptionally { error: Throwable? ->
+              if (!(error is CancellationException || error is CompletionException)) {
+                showErrorGroup("Error while generating code", error)
+              }
+              null
             }
-            cancel()
```
pkukielka: Is removal of the manual cancellation intentional?

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -144,10 +146,16 @@ abstract class FixupSession(
...
-        .get()
+
+    try {
+      val result =
+          agent.server
+              .getFoldingRanges(GetFoldingRangeParams(uri = textFile.uri, range = selection))
+              .get()
+      selectionRange = result.range.toRangeMarker(document, true)
+    } catch (e: Exception) {
+      logger.warn("Error getting folding range", e)
```
pkukielka: Does that mean that instead of failing we will be now showing selection edit lenses in possibly wrong place? (I'm not saying it's wrong decision, just asking about the end result ;))

pkukielka (approved):
I added some comments/questions, but nothing blocking.

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
steveyegge: This is really funny!  Ah, me. I hope we can get this fixed with tree-sitter very soon, so that it just never happens.

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
steveyegge: Yep, good callout. If we are showing the Error group, then it will wait to cancel the session until dismissing/disposing the Error lens group.

It's a bit of a stylistic choice to keep the session going until its Error lens is dismissed. One reason is that looking forward to concurrent edits, I'd like to keep error lenses associated with their originating session.

One small user benefit is starting a new session doesn't dismiss your current error group before you've acknowledged or possibly even seen it.

steveyegge:
I've just done a pretty big merge with main, lots of conflicts. I'm going to close the PR and reopen it after I've tested a bit more.

PR #1663, opened 2024-05-27
Reverse classloader hack logic

pkukielka:
## Changes

Reverse classloader hack logic.
Instead of removing jar from a classloader for IJ > 2022.1 add it for 2022.1 only.

## Test plan

Install plugin in In 2022.1 and 2024.1 and check if no exceptions are thrown on startup.


kalanchan (approved):
stamping to unblock the release

src/main/kotlin/com/sourcegraph/cody/initialization/ActionUpdateThreadClassloaderPatcher.kt
```diff
@@ -1,26 +1,15 @@
...
 import com.intellij.util.lang.UrlClassLoader
-import kotlin.io.path.name
-import kotlin.reflect.full.memberFunctions
-import kotlin.reflect.jvm.javaMethod
 
 object ActionUpdateThreadClassloaderPatcher {
   init {
     val appInfo = ApplicationInfo.getInstance()
-    if (appInfo.majorVersion > "2022" || appInfo.minorVersion > "1") {
+    if (appInfo.majorVersion == "2022" && appInfo.minorVersion <= "1") {
```
RXminuS, line 9: We don't support anything before 2022.1 right? Because otherwise this check wouldn't work

RXminuS (commented):
Yeah...that makes a lot more sense :-D

src/main/kotlin/com/sourcegraph/cody/initialization/ActionUpdateThreadClassloaderPatcher.kt
pkukielka: We don't:
`<idea-version since-build="221.5080.210"/>`

mkondratek (approved):

PR #1665, opened 2024-05-27
Do not skip `textDocument/didChange` on `isUndoTransparentActionInProgress` in `CodyDocumentListener.kt`

### LINKED ISSUE #1648: bug: PANIC on a new line
### Cody Version

local build (but 5.5.15 actually)

### IDE Information

windows 11
intellij 2024.1.2
panic flag on

### Describe the bug

playing with autocompletions (triggering automatically, triggering manually, ctrl+Z, removing lines, etc)

I received a panic error. I cannot tell more precise details. my trace and logs are in the attachments. 

### Expected behavior

no panic

### Additional context

[idea.log.txt](https://github.com/sourcegraph/jetbrains/files/15444289/idea.log.txt)

[agent-trace.json](https://github.com/sourcegraph/jetbrains/files/15444286/agent-trace.json)

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/1648.

Having the scenario like in the Test plan IntelliJ processes two events.
1️⃣ The first event adds `"\n "` (a new line and a space). I am not 100% why there is the space there. I assume that IntelliJ tries to calculate the indent for the new line based on the previous line.
2️⃣ The second event removes the redundant space. IntelliJ realises that the space is redundant and removes it.

Before this change, we ignored the second event (we did not send it to the agent).

## Test plan
```kotlin
class A {
 }<cursor>
```
1. Having above ☝️ 
2. Insert a new line
3. Write a character 

Expected:
no panic

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
@@ -2,7 +2,6 @@ package com.sourcegraph.cody.listeners
 
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.runInEdt
-import com.intellij.openapi.command.CommandProcessor
 import com.intellij.openapi.editor.event.BulkAwareDocumentListener
 import com.intellij.openapi.editor.event.DocumentEvent
 import com.intellij.openapi.fileEditor.FileEditorManager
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
LGTM 👍🏻 My intuition is that we always want to sync document contents. Removing this condition makes sense to me.

PR #1670, opened 2024-05-28
Attach the edit-accept listener to the generated file for **Generate Unit Tests**

mkondratek:
## Test plan
1. `Generate Unit Tests` for a file that has no tests yet
2. Wait for edits, so accept group appears
3. Switch to original file and modify it
4. Switch back to the tests (the accept lens group should be available)
5. Modify the file (the accept lens group should disappear)


pkukielka (commented):
Good catch!

pkukielka (approved):
LGTM

PR #1671, opened 2024-05-28
Fixed the keyboard shortcuts for "Accept" and "Show diff".

odisseus:
Fixes https://github.com/sourcegraph/cody-issues/issues/379.

IIRC IntelliJ automagically converts Windows keyboard shortcuts that use Ctrl into Mac OS shortcuts that use Cmd. This somehow causes an additional mapping to appear, e.g. `Cmd-Shift-'` for `Ctrl-Shift-'`; however, this additional mapping doesn't actually work.

It seems that, when IntelliJ is launched by `./gradlew runIDE`, the `$default` keymap happens to be called "Windows" even on Mac OS. At least this is what I observed in a debug session; in the UI, however, the keymap that's selected by default is "macOS". I haven't investigated this in depth, but this looks like a bug in IntelliJ.

Whatever the actual cause, setting [`replace-all="true"`](https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html#idea-plugin__actions__action__keyboard-shortcut) wherever we define more than one shortcut has solved this problem.

After #1409, there are no places in the code where `DIFF_SESSION_DATA_KEY` is added to the action data context. As a consequence, the algorithm for mapping the keyboard shortcut to an action (which IntelliJ does dynamically every time) decided that this action is disabled, and removed it from the list of possible action mappings.

## Test plan

Tested manually.

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -207,18 +207,16 @@
         <action id="cody.command.Explain"
                 icon="/icons/chat/chat_command.svg"
                 class="com.sourcegraph.cody.chat.actions.ExplainCommand">
-            <keyboard-shortcut first-keystroke="control shift E"
-                               keymap="Mac OS X 10.5+"/>
-            <keyboard-shortcut first-keystroke="alt shift 1" keymap="$default"/>
+            <keyboard-shortcut replace-all="true" first-keystroke="control shift E" keymap="Mac OS X 10.5+"/>
+            <keyboard-shortcut replace-all="true" first-keystroke="alt shift 1" keymap="$default"/>
             <add-to-group group-id="CodyEditorActions" anchor="last"/>
         </action>
 
         <action id="cody.command.Smell"
                 icon="/icons/chat/chat_command.svg"
                 class="com.sourcegraph.cody.chat.actions.SmellCommand">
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -260,8 +260,8 @@
                 text="Accept Edit"
                 description="Accept the fixup">
             <!-- Intended as ctrl shift PLUS but that is not recognized by IntelliJ. -->
-            <keyboard-shortcut first-keystroke="ctrl shift EQUALS" keymap="$default"/>
-            <keyboard-shortcut first-keystroke="ctrl shift EQUALS" keymap="Mac OS X 10.5+"/>
+            <keyboard-shortcut first-keystroke="ctrl shift EQUALS" keymap="$default" replace-all="true"/>
```
pkukielka: Can we do that for al the binding in the `plugin.xml`?
If we will change any of those shortcuts in future we will run into the same issue.

odisseus: I have applied this change to all actions which have more than one shortcut. I'm not sure if we should apply this to the other actions. After all, there must be some reasons why the API developers haven't made this behaviour the default.

pkukielka (approved):
LGTM

PR #1673, opened 2024-05-28
added error handling to the Document and Edit actions

steveyegge:
This allows the Agent to restart if they encounter an IO error -- it wouldn't restart previously

## Test plan

Manually tested in debugger for now because it's tricky to cause this particular code path to trigger in a test.


pkukielka (approved):
LGTM

PR #1679, opened 2024-05-28
fix: exclude edit & retry lens from unit test command

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1716914610459299
CLOSE https://github.com/sourcegraph/cody-issues/issues/404

This PR aims to disable the `Edit & Retry` lens for the Unit Test command to match the behavior in VS Code

This commit introduces a new `commandName` property to the `DocumentCodeSession`, `EditCodeSession`, and `TestCodeSession` classes. This property is used to determine if a command is a unit test command in the `FixupSession` class.

It also modifies the `createAcceptGroup` method in the `LensGroupFactory` class to exclude the "Edit & Retry" action for unit test commands.

## Test plan

1. Run the Unit Test command to verify the `Edit & Retry` lens does not show up.
2. Manually test the "Accept" group creation for both regular and unit test commands to verify the correct actions are included.

image

Confirm the lenses are showing up correctly for other edit commands.


src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -181,7 +184,11 @@ abstract class FixupSession(
       // Then they transition to finished.
       CodyTaskState.Finished -> dispose()
       CodyTaskState.Error -> dispose()
-      CodyTaskState.Pending -> {}
+      CodyTaskState.Pending -> {
+        // TODO: This is a hack to allow the unit test command to be recognized.
+        // Currently we only set the task state to Pending for the unit test command.
+        isUnitTestCommand = true
```
mkondratek: This looks very hackish 😕 I would rather override the value in the impl (`com.sourcegraph.cody.edit.sessions.TestCodeSession`)

mkondratek (changes requested):

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
abeatrix: thanks for the suggestion! I'm not familiar with kotlin so could use your feedback on the updated approach. Does the updated code looks like what you have in mind?

mkondratek (approved):
thanks for fixes, lgtm

PR #1705, opened 2024-05-30
Several fixes for the Edit dialog

steveyegge:
Fixes for prompt handling:

  * do not open with the last prompt if the last action was an Accept
  * fixed a race condition in recording task acceptance
  * fixed up/down history arrows when only 1 history item was present

Other fixes:
  * Switched to a built-in utility for displaying hotkey text in code lenses
  * Fixed a JBInsets warning

## Test plan

I tested this manually. Will need integration/polly tests in order to write real tests for it.


pkukielka (approved):
LGTM

PR #1711, opened 2024-05-31
Make the Add/Edit repositories popup dialog activate more readily

dominiccooney:
Fixes sourcegraph/cody-issues#427 by supporting tiny drags versus clicks, narrow chat sidebars, etc.

## Test plan

Tested manually:

- Open the enhanced repo selector and try double clicking add/edit repositories rapidly
- Add some repositories and try rapidly selecting/deselecting them, then quickly edit repositories
- Position window lower right of monitor, make chat sidebar narrow, try adding a repository
- Activate with left and "right" (depends on platform) clicks, but not middle button clicks, etc.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
@@ -199,8 +199,6 @@ class EnterpriseEnhancedContextPanel(project: Project, chatSession: ChatSession)
     private const val ENTER_MAP_KEY = "enter"
   }
 
-  // TODO: We need to kick off setting the agent state with
-  // controller.loadFrom...(getContextState()) etc.
   private var controller =
       EnterpriseEnhancedContextStateController(
           project,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/context/ui/EnhancedContextPanel.kt
```diff
@@ -256,28 +254,17 @@ class EnterpriseEnhancedContextPanel(project: Project, chatSession: ChatSession)
...
-            pressedTarget = targetForEvent(e)
-          }
-
-          override fun mouseClicked(e: MouseEvent) {
-            var clickTarget = targetForEvent(e)
-            if (e.clickCount == 1 &&
-                e.button == MouseEvent.BUTTON1 &&
-                pressedTarget === clickTarget &&
-                clickTarget is ContextTreeEditReposNode) {
+            if (targetForEvent(e) is ContextTreeEditReposNode) {
```
odisseus: Does this mean that the same action will be invoked on all mouse clicks, regardless of whether it's the left or right (or middle) button? I'm not sure that's a good idea.

dominiccooney: It is, but I can limit it to left or right clicks.

odisseus (approved):

PR #1655, opened 2024-05-26
Fix IntelliJ 2021/2024 compatibility issues

pkukielka:
## Changes

Problem: IJ Intellij 2024 `AnAction` classes need to override `getActionUpdateThread` to define if they are running on `EDT` or `BCG` thread:

```kt
override ActionUpdateThread getActionUpdateThread() {
    return ActionUpdateThread.EDT;
  }
```

Unfortunately both `getActionUpdateThread` method in `AnAction` interface, and `ActionUpdateThread` which it returns, were added in IJ `2022.2`, and we still supports IJ `2022.1`.
That means we need to use classes which, for some of the supported by use versions, does not exist.

My solution is based on adding `libs/ActionUpdateThread.jar` library which contains just one simple enum `com.intellij.openapi.actionSystem.ActionUpdateThread`, which is a copy of enum which exsists in IJ versions > `2022.2`.
That allows us to compile it and run targeting `2022.1`. 
But it creates problems for `2022.2` and above, as here `ActionUpdateThread` already exists on the classpath, and IJ classloader does not allow for such duplication and will fail on runtime.
To fix that problem we added class `ActionUpdateThreadClassloaderPatcher`. It uses the fact that JVS loads classes when they are first needed, and removes `ActionUpdateThread.jar` from classpath before it is needed for a first time.
A bit of reflection hacking was still needed, because `reset` method in the meantime changed signature slightly so compile time binding was not working as expected.

## Test plan

Full QA is needed after this change as it touches almost all IJ actions

PR #1712, opened 2024-05-31
Drop support for 2022.1 to unblock 6.0 release.

dominiccooney:
Marketplace plugin verifier doesn't understand the custom classloader in #1655 and complains about "Access to unresolved class ActionUpdateThread" in the 2022.1s

This drops support for 2022.1 entirely, and we can clean up extra JAR and custom classloader later.

## Test plan

```
./gradlew :runPluginVerifier
```

Note that compatibility warnings are not skipped in the plugin verifier now.


odisseus (approved):

olafurpg (approved):
LGTM 👍🏻

PR #1714, opened 2024-05-31
Document Synchronization: stop using logical positions

olafurpg:
As we recently discovered, logical positions interpret character positions differently than the agent server. For example, in IDEA, the tab character is interpret as two characters while the agent server interprets it as one character. This PR replaces usage of logical positions with manually calculated line/character positions.

## Test plan

Green CI


pkukielka (approved):
LGTM!

PR #1717, opened 2024-05-31
Revert "Drop support for 2022.1 to unblock 6.0 release. (#1712)"

dominiccooney:
This reverts commit 5af5bf6b525da385b5a39f2286090289cd5df77e (#1712)

Verification failures don't block unhiding releases. The verification failures in 2022.1 are because the static plugin verifier doesn't use the custom classloader.

## Test plan

CI


dominiccooney:
Done already by @pkukielka in #1717

PR #1741, opened 2024-06-05
Make the CODY_AGENT_DEBUG=true setting work even if CODY_DIR is not set.

dominiccooney:
For collecting performance data it is handy to run Agent with `--inspect`, even if you're not building it. Enable this by making `CODY_AGENT_DEBUG` work even if `CODY_DIR` is not set.

## Test plan

Manually tested. Run the steps in CONTRIBUTING.md and collect and analyze a trace.


pkukielka (approved):
LGTM

PR #1745, opened 2024-06-05
Windows: use x64 emulation on ARM64

olafurpg:
Previously, Cody didn't work on ARM64 Windows computers because we don't distribute an ARM64 Node.js binary for the Cody agent. This PR fixes the problem by falling back to the x64 Node.js binary, which should be executable on ARM64 Windows computers via emulation.

## Test plan

@danielmarquespt verified this work with an ARM64 VM.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
@@ -244,7 +244,17 @@ private constructor(
 
     private fun nodeBinaryName(): String {
       val os = if (SystemInfoRt.isMac) "macos" else if (SystemInfoRt.isWindows) "win" else "linux"
-      val arch = if (CpuArch.isArm64()) "arm64" else "x64"
+      val arch =
+          if (CpuArch.isArm64()) {
+            if (SystemInfoRt.isWindows) {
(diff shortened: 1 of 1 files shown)


danielmarquespt:
Just tested it on a local arm64 VM and it works:

Screenshot 2024-06-05 at 11 47 25

danielmarquespt (approved):

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -244,7 +244,17 @@ private constructor(
 
     private fun nodeBinaryName(): String {
       val os = if (SystemInfoRt.isMac) "macos" else if (SystemInfoRt.isWindows) "win" else "linux"
-      val arch = if (CpuArch.isArm64()) "arm64" else "x64"
+      val arch =
+          if (CpuArch.isArm64()) {
+            if (SystemInfoRt.isWindows) {
+              // Use x86 emulation on arm64 Windows
```
odisseus, line 250: Did you mean "x64 emulation"?

olafurpg: yes! my bad

PR #1761, opened 2024-06-07
Implemented ghost text discoverability for Inline Edits

steveyegge:
In VS Code's Cody client, we insert ghost text with a hint that you can do inline edits.  

image

In order to improve discoverability, we've added this to JetBrains per @danielmarquespt's design. He also helped implement the final L&F.

## Test plan

Updated `TESTING.md` with a new section, `General commands availability when selection is active`, containing detailed checks to perform on the ghost text.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -175,16 +175,14 @@ automatically.
 
 #### Expected behaviour
 
-It should show something like this: `Ctrl + Shift + ⏎ to Edit,  Alt + = to Chat`
+It should show something like this: `Ctrl + Shift + ⏎ to Edit`
 (Note: that text above is not intended to be accurate. The actual hotkeys displayed
 should match whatever we have most recently set them to; they change now and then.)
 
 * There should only be one hint visible at a time, and should update as the selection changes.
 * The hint should disappear as soon as the selection disappears.
-* The hotkey help should match the editor background if not on same line as the caret.
-* The hotkey help should match the caret-line background  not on same line as the caret.
-* The hotkeys should be correct.
-* The colors should work with light and dark themes.
+* The hotkeys displayed in the hint should be correct.
+* The colors chosen should be clearly visible on all themes.
 * It is OK if the hint is not visible in some places because it's offscreen.
 
 ### Explain Code
--- a/TESTING.md
+++ b/TESTING.md
@@ -168,6 +168,25 @@ automatically.
 * All commands are visible in the Commands panel and can be selected.
 * All commands works after selection.
 
+### General commands availability when selection is active
+
+1. Open file with source code and select some fragment.
+2. Visually confirm that the hotkeys for Edit and Chat are displayed at the end of the selection.
+
+#### Expected behaviour
+
+It should show something like this: `Ctrl + Shift + ⏎ to Edit,  Alt + = to Chat`
+(Note: that text above is not intended to be accurate. The actual hotkeys displayed
+should match whatever we have most recently set them to; they change now and then.)
+
+* There should only be one hint visible at a time, and should update as the selection changes.
+* The hint should disappear as soon as the selection disappears.
+* The hotkey help should match the editor background if not on same line as the caret.
+* The hotkey help should match the caret-line background  not on same line as the caret.
+* The hotkeys should be correct.
+* The colors should work with light and dark themes.
+* It is OK if the hint is not visible in some places because it's offscreen.
+
 ### Explain Code
 
 1. Paste the following Java code:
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
```diff
@@ -0,0 +1,162 @@
...
+import java.awt.Graphics2D
+import java.awt.Rectangle
+import java.awt.event.KeyEvent
+import java.awt.geom.GeneralPath
+import java.util.*
+
+class CodySelectionInlayManager {
+
+  private var currentInlay: Inlay<*>? = null
+  private var inlayBounds: Rectangle? = null
```
mkondratek: How is it used? 🤔

src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
```diff
@@ -0,0 +1,162 @@
...
+    val endLine = editor.document.getLineNumber(endOffset)
+    val selectionEndLine = if (startOffset > endOffset) startLine else endLine
+
+    val editShortcutText = getKeyStrokeText("cody.editCodeAction")
+    val inlayContent = " $editShortcutText  to Edit    "
+
+    updateInlay(editor, inlayContent, selectionEndLine)
+  }
+
+  fun dispose() {
```
mkondratek, line 141: it seems to be unused

src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
```diff
@@ -0,0 +1,162 @@
...
+          if (isMac && modifiers and KeyEvent.META_DOWN_MASK != 0) append("Cmd$separator")
+        }
+        return (modText + key).removeSuffix(separator)
+      }
+    }
+    return ""
+  }
+
+  fun handleSelectionChanged(editor: Editor, event: SelectionEvent) {
+    val startOffset = event.newRange.startOffset
```
mkondratek: nit: I see there are two execution branches here
1. startOffset == endOffset -> clearInlay -> return
2. startOffset != endOffset -> updateInlay -> clearInlay -> set new inline

IIUC you can extract `clearInlay()` and call it at the beginning of this method.

src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
steveyegge: Yep good call -- wasn't used; I removed it.

src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
steveyegge: Done!

src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
steveyegge: Fixed!

steveyegge:
@mkondratek ready for another look! Thanks for the good catches.

TESTING.md
```diff
@@ -168,6 +168,23 @@ automatically.
 * All commands are visible in the Commands panel and can be selected.
 * All commands works after selection.
 
+### General commands availability when selection is active
```
mkondratek, line 171: image

It looks like this for me. Is it expected? 

Probably a space distance b/w the blue box and the last character in the line would make it better. Can you link Daniel's design?

steveyegge: cc: @danielmarquespt 

OK, I have solved several problems. Thank you @mkondratek for making me go reexamine it!

The root cause of several problems was that we were using the IDE font for our calculations instead of the Editor font, but they can actually be very different, including the size. The problems included the right margin being too short, the text drawing out of bounds at different font sizes, the widget resizing with the IDE font, etc.

I'm still playing with it to make sure it works with different fonts as well. Will update the PR in a bit.

steveyegge:
OK! I have finished overhauling how we compute everything. I've tested this at 8, 13, 14, 16, 24, 36 and 40 point font sizes for 6 different monospaced fonts, and they all draw within reasonable tolerances of "centered" in both directions with good padding/margins.

There were many bits that did not scale before, such as the corner arc of the "tab", which is now scaled to the font size.

@danielmarquespt @mkondratek please take another look!

Here is JetBrains Mono 13-point:
image

Here is JB Mono at 40-point:
image

mkondratek (approved):
lgtm

gauravkumar37:
Related: https://github.com/sourcegraph/jetbrains/issues/1798

PR #1766, opened 2024-06-10
Implement integration tests for Jetbrains

### LINKED ISSUE #1155: Integration Test Suite
```[tasklist]
## Testing Suite
```
[RFC](https://docs.google.com/document/d/1n23Kd9z_QNqIFNrDqpftvOV0f6VRRMdakxpCLv8QBYU/edit#heading=h.trqab8y0kufp)

pkukielka:
Partially fixes https://github.com/sourcegraph/jetbrains/issues/1155

This PR contains testing framework created by @steveyegge and some of my finishing touches.

## Changes

New framework allows us to test end-to-end Jetbrains interactions, including appearance and layout of various UI elements!
It should help us significantly  improve code coverage, correctness, and speed of the delivery of the new features
 
For the big picture please look at [RFC](https://docs.google.com/document/d/1n23Kd9z_QNqIFNrDqpftvOV0f6VRRMdakxpCLv8QBYU/edit#heading=h.trqab8y0kufp) by [@steveyegge](https://github.com/steveyegge)

Biggest changes:

* brand new integration test suite framework was added
* it supports polly.js recordings for all network requests/responses (same as agent tests)
* tests for DocumentCode action were already added, both as example and a proof that new frameworks really works and is simple to use

This PR also adds 3 new Gradle tasks:

* integrationTest - runs the tests using pre-recorded networks requests/responses using polly.js
* recordingIntegrationTest - runs the tests and records missing network requests/responses (they are added to the [recordings](https://github.com/sourcegraph/jetbrains/pull/1766/files#diff-dfc5f909b2f643c169ea205822de4110110cae72660a814f5d736eea75b52aeb) file)
* passthroughIntegrationTest - runs the tests running all the network requests/responses from scratch; does not touch recordings in any way

For simplicity we are using [the same access tokens as Cody agent tests](https://github.com/sourcegraph/cody/blob/main/agent/scripts/export-cody-http-recording-tokens.sh). Currently only SRC_DOTCOM_PRO_ACCESS_TOKEN is really used.
Used access token and other test parameters are defined in [build.gradle.kts](https://github.com/sourcegraph/jetbrains/pull/1766/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06)
## Test plan

N/A - this PR is adding automated tests which should be run on every `./gradelw check` run.


build.gradle.kts
```diff
@@ -84,6 +91,9 @@ dependencies {
   implementation("com.googlecode.java-diff-utils:diffutils:1.3.0")
   testImplementation("org.awaitility:awaitility-kotlin:4.2.0")
   testImplementation("org.mockito:mockito-core:5.12.0")
+  testImplementation("org.junit.jupiter:junit-jupiter:5.7.0")
```
mkondratek: why not the latest (5.10.2)?

build.gradle.kts
```diff
@@ -84,6 +91,9 @@ dependencies {
...
+  testImplementation("org.jetbrains.kotlin:kotlin-test-junit:1.9.22")
```
mkondratek: the latest is (2.0.0)

build.gradle.kts
```diff
@@ -101,6 +111,7 @@ spotless {
     ktfmt()
     trimTrailingWhitespace()
     target("src/**/*.kt")
+    toggleOffOn()
```
mkondratek, line 114: why do we need it?

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.sourcegraph.cody.edit.DocumentCodeTest
+import com.sourcegraph.cody.util.CodyIntegrationTextFixture
+import java.util.concurrent.TimeUnit
+import org.junit.AfterClass
+import org.junit.runner.RunWith
+import org.junit.runners.Suite
+
+// We need a single tearDown() method running after all tests are complete,
+// so agent can be closed gracefully and recordings can be saved properly.
+// Due to the limitations of JUnit 4 this can be done only using SuiteClasses and AfterClass.
```
mkondratek: 1. What are the exact limitations of JUnit 4? 
2. Do we use JUnit 4 of 5? 
3. What would be a better solution ("unlimited")?

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
```diff
@@ -0,0 +1,37 @@
...
+@RunWith(Suite::class)
+@Suite.SuiteClasses(DocumentCodeTest::class)
+class AllSuites {
+  companion object {
+    @AfterClass
+    @JvmStatic
+    internal fun tearDown() {
+      CodyAgentService.withAgent(CodyIntegrationTextFixture.myProject!!) { agent ->
+        val errors = agent.server.testingRequestErrors().get()
+        val missingRecordings = errors.filter { it.error?.contains("`recordIfMissing` is") == true }
```
mkondratek, line 33: I don't get it. What is this check:
```
it.error?.contains("`recordIfMissing` is")
```
??? 

Looks random without a context. Can we have a comment here?

src/main/resources/CodyBundle.properties
```diff
@@ -193,7 +193,7 @@ gotit.autocomplete.message=This is how Cody displays autocomplete suggestions.<b
 # User Friendly Error Notifications
 notification.general.cody-not-running.title=Cody is starting...
 notification.general.cody-not-running.detail=Please wait a few seconds or check the Cody logs for errors.
-notifications.edits.editing-not-available.title=Cody can't edit the file
+notifications.edits.editing-not-available.title=Cody can't edit this file
```
mkondratek: IIRC I changed it consciously to cover the scenario like this:
1. Generate Unit Tests in `FunnyUtils.kt`
2. <Cody switches the document to `FunnyUtilsTest.kt`>
3. Delete `FunnyUtilsTest.kt` while Cody is generating the tests. 
4. <The error appears>

At point 4 we are back in `FunnyUtils.kt` - not in `FunnyUtilsTest.kt`. The word **this** is inaccurate.

src/main/kotlin/com/sourcegraph/cody/initialization/PostStartupActivity.kt
```diff
@@ -38,7 +38,10 @@ class PostStartupActivity : StartupActivity.DumbAware {
       // Scheduling because this task takes ~2s to run
       CheckUpdatesTask(project).queue()
     }
-    if (ConfigUtil.isCodyEnabled()) CodyAgentService.getInstance(project).startAgent(project)
+    // For integration tests we do not want to start agent immediately as we would like to first do
+    // some setup
+    if (ConfigUtil.isCodyEnabled() && !ConfigUtil.isIntegrationTestModeEnabled())
```
mkondratek: let's use brackets please

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -134,15 +134,19 @@ class CodyAuthenticationManager(val project: Project) : Disposable {
...
-                serverUrlChanged = previousUrl != account?.server?.url,
-                accessTokenChanged = previousToken != account?.let { getTokenForAccount(it) }))
+      val serverUrlChanged = previousUrl != account?.server?.url
+      val accessTokenChanged = previousToken != account?.let { getTokenForAccount(it) }
+      val tierChanged = previousTier != account?.isDotcomAccount()
+
+      if (serverUrlChanged || accessTokenChanged || tierChanged) {
+        CodyAgentService.withAgentRestartIfNeeded(project) { agent ->
+          agent.server.configurationDidChange(ConfigUtil.getAgentConfiguration(project))
+          publisher.afterAction(AccountSettingChangeContext(serverUrlChanged, accessTokenChanged))
```
mkondratek: I guess we should pass `accountTierChanged` here too (however it seems we do not use it - so we can remove the field from the context class too).

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -251,7 +251,8 @@ class CodyAgentService(private val project: Project) : Disposable {
         restartIfNeeded: Boolean,
         callback: Consumer<CodyAgent>,
         onFailure: Consumer<Exception> = Consumer {}
-    ) {
+    ): CompletableFuture<Boolean> {
```
mkondratek, line 254: why do we need this? It seem we are not using the return object anywhere

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
```diff
@@ -461,6 +478,23 @@ abstract class FixupSession(
     return documentListener.isAcceptLensGroupShown.get()
   }
 
+  private fun publishProgress(topic: Topic<CodyInlineEditActionNotifier>) {
```
mkondratek, line 481: this one publishes on edt too (the same as `publishProgressOnEdt`) - the method name is misleading (having both, this suggests that the publishing does not happen on edt). pls rename

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -70,18 +71,28 @@ class FixupService(val project: Project) : Disposable {
...
     if (!CodyEditorUtil.isEditorValidForAutocomplete(editor)) {
       runInEdt { EditingNotAvailableNotification().notify(project) }
       logger.warn("Edit code invoked when editing not available")
       getActiveSession()?.cancel()
       return false
     }
+
+    // TODO: We'll have to figure out a way to integration-test the ignore stuff.
+    // Right now during testing the policy comes back null, which depending on the
+    // account type might mean ignore-by-default. So we override it for now.
```
mkondratek: let's have it somewhere in Linear

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensSpinner.kt
```diff
@@ -1,61 +1,15 @@
...
 import java.awt.Graphics2D
-import java.awt.geom.AffineTransform
 import javax.swing.Icon
-import javax.swing.Timer
 
 class LensSpinner(group: LensWidgetGroup, private val icon: Icon) : LensWidget(group) {
-  private var rotationDegrees = 0f
-  private val animationDelay = 50 // Milliseconds between frames
-
-  private val timer =
```
mkondratek, line 14: why is it redundant now? 🤔

mkondratek (commented):
comments and questions 💬

build.gradle.kts
steveyegge: This is specifically to allow you to exclude certain regions from formatting, which is necessary when you are configuring an integration test via a string in the source code. I don't have any such tests at the moment but it is a supported configuration.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
steveyegge: Another facility that can be useful for testing. I was using it extensively for a while. I wound up redoing those tests, but I still think it might be useful.

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensSpinner.kt
steveyegge: I think the issue here was that the test was failing sporadically because the spinner wasn't getting disposed. But @pkukielka found a way to do it without that code, using an icon with rotation information encoded in it.

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
pkukielka: This is because we are using JB testing classes (e.g. `BasePlatformTestCase`) which are based on specific `junit` version, and `junit4.jar` is part of the `ideaIC` package.
We will upgrade to `junit5` when we will upgrade base IJ version.

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensSpinner.kt
pkukielka: That is correct :)

The real change which enabled it was:
```
- icon.paintIcon(null, g, x.toInt(), y.toInt())
+ icon.paintIcon(parentGroup.editor.contentComponent, g, x.toInt(), y.toInt())
```

Proper definition of component from which icon painting should get it properties fixed the default animation and allowed to remove all the custom code.

src/main/kotlin/com/sourcegraph/cody/edit/sessions/FixupSession.kt
pkukielka: Right! I removed `publishProgressOnEdt` completely.

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
pkukielka: The story there is a bit more complicated.
I added proper comment to the `AccountSettingChangeContext `

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
pkukielka: I added a comment and a better error message.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
pkukielka: I think we maybe should fix it right away.
I do not like mixing production and test code, and it would be useful to understand how it could be `null`. Possibly it's some kind of bug we are silently ignoring there.

I will take a deeper look.

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
mkondratek: ack, IMO it would be nice to mention the origin of the limitation in the same comment

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
```diff
@@ -22,9 +22,19 @@ class AllSuites {
...
+        // We extract polly.js errors to notify users about the missing recordings, if any
         val missingRecordings = errors.filter { it.error?.contains("`recordIfMissing` is") == true }
         missingRecordings.forEach { missing ->
-          System.err.println("Missing recording: ${missing.error}")
+          System.err.println(
+              """Recording is missing: ${missing.error}
+                |
+                |------------------------------------------------------------------------------------------
+                |To fix this problem please run `./gradlew :recordingIntegrationTest`.
+                |You need exported access tokens first, using script from the `sourcegraph/cody` repository:
```
mkondratek: nit: `You need to export access tokens first`

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
mkondratek: fine, so the variable was missing 😉

src/main/kotlin/com/sourcegraph/cody/edit/widget/LensSpinner.kt
mkondratek: awesome

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
pkukielka: I think previously it might have been caused by a race with an agent startup.
After some re-arrangements of the testing logic problem seems to not appear anymore and whole custom logic here can be removed.

src/main/kotlin/com/sourcegraph/cody/edit/FixupService.kt
```diff
@@ -70,23 +69,27 @@ class FixupService(val project: Project) : Disposable {
       logger.warn("Edit code invoked when Cody not enabled")
       return false
     }
-    if (CodyStatusService.getCurrentStatus(project) == CodyStatus.CodyAgentNotRunning) {
+
+    if (!CodyAgentService.isConnected(project)) {
```
pkukielka, line 73: This is a bit more explicit/correct, as it is based on real state and not the status reported on the UI.

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
pkukielka: Fixed, thanks!

mkondratek (approved):
go go go

PR #1767, opened 2024-06-10
Updated Java to version 17 in CI

odisseus:
This is a prerequisite for https://github.com/sourcegraph/cody-issues/issues/240.

## Test plan

Green CI

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,9 +12,9 @@ jobs:
           node-version: 20
       - uses: actions/setup-java@v4
         with:
-          distribution: "temurin"
-          java-version: 11
-          cache: "gradle"
+          distribution: temurin
+          java-version: 17
+          cache: gradle
       # See note about QEMU and binfmt requirement here https://github.com/vercel/pkg#targets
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -10,10 +10,10 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 20
-      - uses: actions/setup-java@v3
+      - uses: actions/setup-java@v4
         with:
           distribution: "temurin"
-          java-version: 11
```
mkondratek, line 16: so the plugin will be built with java 17 and published. will a user be able to install it in IntelliJ 2022.1?

odisseus: For compatibility with 2022.1, we need to produce Java 11 bytecode. However, this is [configured](https://github.com/sourcegraph/jetbrains/blob/main/gradle.properties#L18) in a different place.

mkondratek: ack, thanks

PR #1770, opened 2024-06-11
Upgraded Gradle to version 8.8

odisseus:
Gradle 8.2 or later is a prerequisite for https://github.com/sourcegraph/cody-issues/issues/240.

## Known problems

Three integration tests fail, but they also fail in main.

## Test plan

Green CI


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.1.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.2-bin.zip
```
mkondratek: can we upgrade to the latest release? why no?

odisseus: Looks like we can.

pkukielka (approved):
LGTM

PR #1777, opened 2024-06-12
Run all tests on CI

pkukielka:
## Changes

All tests should be run on CI, not only unittests.

Fixed problems:

- integrations tests were not triggered as part of the run
- missing redacted tokens for CI
- too tight ignore policy timeout for CI
- race between document change listeners and commands sent in tests (fixed by using `PlatformTestUtil.dispatchAllEventsInIdeEventQueue()`)
- CI does not allow on setting `+x` for executables - fallback mechanism was needed
- `DISABLE_FEATURE_FLAGS` was needed to avoid flakiness in recordings 
- `showWorkingGroup` was not called on the EDT in one place
- `TOPIC_PERFORM_ACCEPT` notification was sent on wrong thread causing a race
- Ignore `ProcessCanceledException` during the test `tearDown`
 
## Test plan

N/A

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -218,9 +218,12 @@ fun Test.sharedIntegrationTestConfig(buildCodyDir: File, mode: String) {
       "idea.test.execution.policy" to "com.sourcegraph.cody.test.NonEdtIdeaTestExecutionPolicy",
       "test.resources.dir" to resourcesDir.absolutePath)
 
+  val accessToken = System.getenv("SRC_DOTCOM_PRO_ACCESS_TOKEN")
+  if (accessToken != null) {
+    environment("SRC_ACCESS_TOKEN" to accessToken)
+  }
+
   environment(
-      "CODY_INTEGRATION_TEST_TOKEN" to System.getenv("SRC_DOTCOM_PRO_ACCESS_TOKEN"),
-      "SRC_ACCESS_TOKEN" to System.getenv("SRC_DOTCOM_PRO_ACCESS_TOKEN"),
       "CODY_RECORDING_MODE" to mode,
       "CODY_RECORDING_NAME" to "integration-test",
       "CODY_RECORDING_DIRECTORY" to resourcesDir.resolve("recordings").absolutePath,
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -215,6 +215,7 @@ fun Test.sharedIntegrationTestConfig(buildCodyDir: File, mode: String) {
       "cody.autocomplete.enableFormatting" to
           (project.property("cody.autocomplete.enableFormatting") as String? ?: "true"),
       "cody.integration.testing" to "true",
+      "cody.ignore.policy.timeout" to 500, // Increased to 500ms as CI tends to be slower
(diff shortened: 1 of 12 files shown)


build.gradle.kts
```diff
@@ -219,9 +219,12 @@ fun Test.sharedIntegrationTestConfig(buildCodyDir: File, mode: String) {
...
       "test.resources.dir" to resourcesDir.absolutePath)
 
+  val accessToken = System.getenv("SRC_DOTCOM_PRO_ACCESS_TOKEN")
+  if (accessToken != null) {
+    environment("SRC_ACCESS_TOKEN" to accessToken)
+  }
+
   environment(
-      "CODY_INTEGRATION_TEST_TOKEN" to System.getenv("SRC_DOTCOM_PRO_ACCESS_TOKEN"),
-      "SRC_ACCESS_TOKEN" to System.getenv("SRC_DOTCOM_PRO_ACCESS_TOKEN"),
```
pkukielka, line 223: Gradle was converting `null` to `"null"` string 😠

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -322,17 +321,19 @@ private constructor(
...
           Files.deleteIfExists(binaryTarget)
         }
-        logger.info("extracting Node binary to " + binaryTarget.toAbsolutePath())
+        logger.info("Extracting Node binary to " + binaryTarget.toAbsolutePath())
         Files.copy(binarySource, binaryTarget, StandardCopyOption.REPLACE_EXISTING)
         val binary = binaryTarget.toFile()
         if (binary.setExecutable(true)) {
           binary
         } else {
-          throw CodyAgentException("failed to make executable " + binary.absolutePath)
```
pkukielka, line 331: That was failing on CI, where apparently we cannot make processes executable.
That said I think fallback to the base node process is also sensible default for end users.

odisseus:
Can we run unit tests and integration tests as separate jobs? Both of them happen to be flaky, and I'd rather re-run part of the test suite than the whole.

pkukielka:
> Can we run unit tests and integration tests as separate jobs? Both of them happen to be flaky, and I'd rather re-run part of the test suite than the whole.

Integration tests in replay mode should not be flaky and takes around 5 seconds to execute.
Problem with making them separate steps is that pre-execution setup (e.g. java) is expensive, so it would make whole run significantly longer.
Instead of relaying on rerunning I think we should focus hard on fixing the flakiness.

odisseus:
Consider increasing `ASYNC_WAIT_TIMEOUT_SECONDS`. Hopefully this will reduce flakiness.

Otherwise looks good.

odisseus (approved):

pkukielka:
> Consider increasing `ASYNC_WAIT_TIMEOUT_SECONDS`. Hopefully this will reduce flakiness.
> 
> Otherwise looks good.

TBH on CI in replay mode those tests either pass super quickly or not at all (if there are any problems, which hopefully I just fixed):

image

mkondratek (approved):

PR #1778, opened 2024-06-12
Build: add Gradle settings to auto-download missing JDK

olafurpg:
Previously, I was unable to import the JetBrains plugin codebase into IntelliJ on my Windows computer because I didn't have a compatible installation of Java 11. This diff configures Gradle to automatically download the JDK so that I can import the build into IntelliJ on Windows without a custom setup.

Following docs here https://docs.gradle.org/8.1.1/userguide/toolchains.html#sub:download_repositories

## Test plan
Imported the build into IntelliJ on Windows.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,5 +1,10 @@
 rootProject.name = "Sourcegraph"
 
+plugins {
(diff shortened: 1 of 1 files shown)


pkukielka (approved):
LGTM

PR #1779, opened 2024-06-12
Removed the override of wrapper.gradleVersion

odisseus:
This is a follow-up to #1770.

The `wrapper.gradleVersion` [property](https://docs.gradle.org/8.1.1/userguide/gradle_wrapper.html#customizing_wrapper) affects only the `wrapper` task, which we run only once in a while when updating the wrapper. It is also [possible](https://stackoverflow.com/a/36385128) to specify the desired version as a command-line parameter, which is what we usually do anyway.

With that in mind, I actually don't see any point in defining this property here. 

## Test plan

N/A

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -384,8 +384,6 @@ tasks {
     withType<KotlinCompile> { kotlinOptions.jvmTarget = it }
   }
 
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
LGTM

PR #1780, opened 2024-06-13
Fixed compilation warnings.

odisseus:
## Test plan

Green CI

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/config/CodyAccountDetailsProvider.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/CodyAccountDetailsProvider.kt
@@ -23,7 +23,7 @@ class CodyAccountDetailsProvider(
   ): CompletableFuture<DetailsLoadingResult<CodyAccountDetails>> {
 
     return ProgressManager.getInstance()
-        .submitIOTask(indicator) {
+        .submitIOTask(indicator) { ind ->
           val token =
               accountsModel.newCredentials.getOrElse(account) {
                 accountManager.findCredentials(account)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/sourcegraph/cody/autocomplete/action/AutocompleteActionHandler.kt
```diff
@@ -1,27 +1,23 @@
...
 
 open class AutocompleteActionHandler : EditorActionHandler() {
-  private val logger = Logger.getInstance(AutocompleteActionHandler::class.java)
 
   override fun isEnabledForCaret(editor: Editor, caret: Caret, dataContext: DataContext?): Boolean {
     // Returns false to fall back to normal action if there is no suggestion at the caret.
-    val project = editor.project ?: return false
-    return CodyEditorUtil.isEditorInstanceSupported(editor) &&
-        hasAnyAutocompleteItems(project, caret)
+    if (editor.project == null) return false
```
pkukielka: Looks like that project check is not needed, or at least not there
(project is not used now).

src/main/kotlin/com/sourcegraph/cody/config/CodyAccountDetailsProvider.kt
```diff
@@ -23,15 +23,15 @@ class CodyAccountDetailsProvider(
   ): CompletableFuture<DetailsLoadingResult<CodyAccountDetails>> {
 
     return ProgressManager.getInstance()
-        .submitIOTask(indicator) { indicator ->
+        .submitIOTask(indicator) {
```
pkukielka: Was than an warning?
Previous version was more readable to me.

odisseus: The compiler was complaining because the argument of the lambda expression shadows the method argument.

pkukielka (approved):
LGTM

PR #1782, opened 2024-06-14
Updated key shortcuts to latest bindings from Design

steveyegge:
Updated the JetBrains keyboard shortcuts according to the instructions [here](https://linear.app/sourcegraph/document/keyboard-shortcuts-for-cody-on-jetbrains-2f8747a22530).

## Test plan

I wrote a unit test for the bindings. I would have liked to trigger the keystrokes programmatically in the test (i.e., an integration test), and then listened for the resulting actions to fire. However, without the UI-robot tests, I was not able to find a way to simulate a keypress to trigger an action.

So instead, I wrote a simpler and slightly weaker test, which looks up all the actions of interest, gets their registered key bindings, and compares them against the expected test values from Design. This unfortunately doesn't exercise all the code paths -- specifically it isn't really testing the routing for retry and for the undo/cancel/dismiss cluster, and it's not guaranteeing that the actions will trigger even if the bindings are registered. But it will definitely catch whether we have the wrong bindings for the top-level actions.

I have not manually tested this on Windows yet, because I still need to set up Parallels, and I'm headed out for PTO. But the unit test should work on Windows as well.


mkondratek (approved):
I did not test it myself but it looks good 👍 

pls make sure it works as expected on macOS and Windows

src/main/resources/META-INF/plugin.xml
```diff
@@ -163,13 +169,13 @@
         </action>
         <action id="cody.cycleForwardAutocompleteAction"
                 class="com.sourcegraph.cody.autocomplete.action.CycleForwardAutocompleteAction">
-            <keyboard-shortcut first-keystroke="alt OPEN_BRACKET"
-                               keymap="$default"/>
+            <keyboard-shortcut first-keystroke="alt OPEN_BRACKET" keymap="$default"/>
```
pkukielka: Minor thing, but I think we should add `replace-all="true"` there as in other places. 
That ensures that if we ever change the binding in the plugin configuration old one will be removed.

src/test/kotlin/com/sourcegraph/cody/KeyboardShortcutTest.kt
```diff
@@ -0,0 +1,74 @@
...
+          action.shortcutSet.shortcuts
+              .mapNotNull { shortcut ->
+                when (shortcut) {
+                  is KeyboardShortcut -> shortcut.firstKeyStroke
+                  else -> null
+                }
+              }
+              .toTypedArray()
+
+      val key = if (SystemInfoRt.isMac) macKey else windowsKey
```
pkukielka, line 61: what is expected behaviour on the Linux CI?

pkukielka (approved):
LGTM, I added minor comments.
Also - PR requires spotless reformatting.

src/main/resources/META-INF/plugin.xml
steveyegge: Fixed!

src/test/kotlin/com/sourcegraph/cody/KeyboardShortcutTest.kt
steveyegge: cc: @danielmarquespt -- I think we need to formally define the Linux bindings.

PR #1784, opened 2024-06-17
Replace use of Kotlin Coroutines with CompletableFuture

dominiccooney:
Fixes sourcegraph/cody-issues#437

This slightly changes the behavior of the repo autocompleter, it will wait until it has a complete set of results. This component is going away anyway when the sidebar switches to the webview.

Also removes the unused boolean result of `withAgent`.

## Test plan

- Tested the repo picker locally, including autocompletions and red squiggles on unknown repos
- CI


pkukielka (approved):
LGTM

PR #1788, opened 2024-06-18
Run `testUndo`

mkondratek:
Fixes https://linear.app/sourcegraph/issue/CODY-2437/fix-testundo.

## Test plan
1. greeen ci


pkukielka (approved):
Thanks for checking and reenabling it!

Issue #1763, opened 2024-06-09
bug: java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: No default chat model found


ernestoBizDirect:
### IDE Information

PyCharm 2023.3 (Professional Edition)
Build #PY-233.11799.259, built on December 3, 2023
Licensed to gurgles tumbles
You have a perpetual fallback license for this version.
Subscription is active until September 14, 2026.
Runtime version: 17.0.9+7-b1087.7 x86_64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.6
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 4
Metal Rendering is ON
Registry:
  ide.experimental.ui=true
Non-Bundled Plugins:
  atom-one (1.5.1)
  com.supermaven.intellij (1.25)
  com.intellij.plugins.vscodekeymap (233.11799.188)
  app.aicommit.plugin (3.0.8)
  net.firebits.odoo (1.2.1)
  at.wtioit.intellij.plugins.odoo (0.6.11)
  dev.ngocta.pycharm-odoo (2024.4.0.233)
  com.sourcegraph.jetbrains (6.0.4-nightly)
  cn.yiiguxing.plugin.translate (3.6.1)

### Bug Description

There must be some way to reload the models, sometimes the IDE loads and I have a bad Internet connection and therefore the models do not load, then even though the connection returns it is no longer possible to reload, I have to close the IDE and restart it to open. On other occasions, only the free Claude Sonnet model is loaded, even though he recognizes that my subscription is the Pro, in the same way in that case I must restart the IDE

### Additional context

Stacktrace:

```text
java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: No default chat model found
 
Error: No default chat model found
    at /Users/developer/Library/Application Support/JetBrains/PyCharm2023.3/plugins/Sourcegraph/agent/index.js:112466:19
 at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
 at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
 at com.sourcegraph.cody.chat.AgentChatSession.submitMessageToAgent$lambda$0(AgentChatSession.kt:121)
[...]

PR #1793, opened 2024-06-19
Do not restore chat sessions without connectionId

mkondratek:
Related to https://github.com/sourcegraph/jetbrains/issues/1763.

## Test plan
1. Start IDE
2. ensure `chat/restore` is not called


pkukielka (approved):
LGTM

PR #1795, opened 2024-06-20
Add telemetry v2 support

pkukielka:
Fixes CODY-913

## Changes

Add support for telemetry v2 and migrate few of the existing calls (not all of them need to be migrated to v2 as some v2 events now comes from agent).

## Test plan

1. Ask cody to generate custom code
2. Copy/paste it using the button, or insert it at cursor
3. Check appropriate looker board to see if event exists (I will paste a screenshot later)

### DIFF
--- a/src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
+++ b/src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
@@ -1,18 +1,9 @@
 package com.sourcegraph.cody.chat;
 
-import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.command.WriteCommandAction;
 import com.intellij.openapi.diagnostic.Logger;
-import com.intellij.openapi.editor.CaretModel;
-import com.intellij.openapi.editor.Document;
-import com.intellij.openapi.editor.Editor;
-import com.intellij.openapi.editor.EditorFactory;
-import com.intellij.openapi.editor.EditorSettings;
-import com.intellij.openapi.editor.event.DocumentEvent;
-import com.intellij.openapi.editor.event.DocumentListener;
-import com.intellij.openapi.editor.event.EditorMouseEvent;
-import com.intellij.openapi.editor.event.EditorMouseListener;
-import com.intellij.openapi.editor.event.EditorMouseMotionListener;
+import com.intellij.openapi.editor.*;
+import com.intellij.openapi.editor.event.*;
 import com.intellij.openapi.editor.ex.EditorEx;
 import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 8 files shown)


src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
```diff
@@ -179,9 +171,8 @@ private ActionListener copyCodeListener(EditorEx editor, JButton copyButton) {
       timer.start();
 
       lastCopiedText = text;
-      ApplicationManager.getApplication()
```
pkukielka, line 182: No need to run it on separate thread as telemetry calls just runs an agent call which is done on a separate thread anyway.

src/main/kotlin/com/sourcegraph/cody/chat/ui/SingleMessagePanel.kt
```diff
@@ -123,9 +122,7 @@ class SingleMessagePanel(
         AttributionSearchCommand(project).onSnippetFinished(lastPart.text, connectionId, listener)
       }
 
-      ApplicationManager.getApplication().executeOnPooledThread {
```
pkukielka, line 126: That one is already done by agent.

src/main/kotlin/com/sourcegraph/cody/telemetry/TelemetryV2.kt
```diff
@@ -0,0 +1,67 @@
...
+
+import com.intellij.openapi.application.ApplicationInfo
+import com.intellij.openapi.project.Project
+import com.sourcegraph.cody.agent.CodyAgentService
+import com.sourcegraph.cody.agent.protocol.TelemetryEvent
+import com.sourcegraph.cody.agent.protocol.TelemetryEventParameters
+
+class TelemetryV2 {
+  companion object {
+    private val intellijProductCodeMap =
```
mkondratek, line 11: nice

src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
```diff
@@ -21,9 +21,9 @@ class CodyDocumentListener(val project: Project) : BulkAwareDocumentListener {
     val pastedCode = event.newFragment.toString()
     if (pastedCode.isNotBlank() && CodeEditorFactory.lastCopiedText == pastedCode) {
       CodeEditorFactory.lastCopiedText = null
-      ApplicationManager.getApplication().executeOnPooledThread {
-        GraphQlLogger.logCodeGenerationEvent(project, "keyDown:Paste", "clicked", pastedCode)
-      }
+      GraphQlLogger.logCodeGenerationEvent(project, "keyDown:Paste", "clicked", pastedCode)
+      TelemetryV2.sendCodeGenerationEvent(
+          project, feature = "keyDown", action = "paste", pastedCode)
```
mkondratek, line 26: so we use both v1 and v2 at the moment, right?

mkondratek (approved):
lgtm

src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
pkukielka: Yes, but we won't be adding new V1 events and existing ones will be removed in the near future.

PR #1796, opened 2024-06-21
Improve invalid access token handling

### LINKED ISSUE #1790: JetBrains: Login screen appears for a moment when user switch accounts or relaunch the IDE.
### IDE Information

IntelliJ IDEA 2023.3.3 (Community Edition)
Build #IC-233.14015.106, built on January 25, 2024
Runtime version: 17.0.9+7-b1087.11 aarch64
macOS 14.3
Memory: 2048M
Cores: 8
Metal Rendering is ON
Registry:
ide.experimental.ui=true
Non-Bundled Plugins:
idea.plugin.protoeditor (233.13135.65)
DevKit (233.14015.152)
gherkin (233.11799.165)
cucumber-java (233.11799.196)
PythonCore (233.14015.106)
com.sourcegraph.jetbrains (6.0.5-nightly)
Kotlin: 233.14015.106-IJ

### Bug Description

Platform: macOS

**Steps to reproduce:**

1. Have Cody plugin installed in your JB IntelliJ editor
2. Have both accounts Free & Pro user added
3. Click on Cody icon from bottom, select "Manage Accounts" and select other account
4. Check the Cody screen
5. Close the IDE and relaunch it

### Observed behavior:
Cody login screen appears for a moment for both 1) switching between accounts 2) relaunching the IDE.

### Expected behavior:
Cody login screen should not appear.

### Additional context

**Video link:**

https://drive.google.com/file/d/1yFTai63BETl7v7EaZxIUj9bz-mEVB-Ci/view?usp=drive_link

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/1790.

Fixes https://linear.app/sourcegraph/issue/CODY-2453/jetbrains-login-screen-appears-for-a-moment-when-user-switch-accounts#comment-3527b6fe.

## Review 
Please review this PR commit by commit. They are logically separate changes (but they make most sense all together).

#### [Update pluginVersion in gradle.properties](https://github.com/sourcegraph/jetbrains/pull/1796/commits/29a8e5795c0286b1e43aeb92a3aefee462dad118)

Plugin update notification when running IDE from `runIde` is annoying. This fixes that.

#### [Remove isLoginUniqueChecker](https://github.com/sourcegraph/jetbrains/pull/1796/commits/2da48db2f54e17d9c0ac8de20bfd30d32af5f237)

IMO the check add a redundant complexity. It is a valid case to update the token for an existing account (e.g. when the token is expired). Let's allow it.

#### [Set the default server to the active account's server in SignInWithEn…](https://github.com/sourcegraph/jetbrains/pull/1796/commits/2011de65ccd8e39c12b75d62b4649166d5bf694a) 

When the token is expired let's add the initial value of the server.

#### [Refactor CodyOnboardingGuidancePanel](https://github.com/sourcegraph/jetbrains/pull/1796/commits/9c0ce5015cb6c5ed5e2f37583c5c6fec18df570c)

Let's create it once. Let's update it with the name. Also clean up CodyToolWindowContent a bit.

#### [Add previousToken to CodyAuthenticationManager::setActiveAccount](https://github.com/sourcegraph/jetbrains/pull/1796/commits/19f155f34843d9617358ec42b8cffbd9c23556b2)

This was error prone. We were not catching token changes properly before. 

#### [Add a check for access token validness and improve invalid access tok…](https://github.com/sourcegraph/jetbrains/pull/1796/commits/88b3c4b0b2b153c29eefceb632395ca11bc8f5fa) 

The actual fix to the issue. 

### [Show chat initially when there is an account](https://github.com/sourcegraph/jetbrains/pull/1796/commits/f9e9320fed7a0460e307bb3a88897199e9da6151)

The actual fix to the issue (part 2). 

### The original issue
1. Have Cody plugin installed in your JB IntelliJ editor
2. Have both accounts Free & Pro user added
3. Click on Cody icon from bottom, select "Manage Accounts" and select other account
4. Check the Cody screen
5. Close the IDE and relaunch it

#### Expected
Cody login screen should not appear (when switching, when restarting having an account).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@
 pluginGroup=com.sourcegraph.jetbrains
 pluginName=Sourcegraph
 # SemVer format -> https://semver.org
-pluginVersion=6.0-localbuild
+pluginVersion=7.0-localbuild
 # See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 # for insight into build numbers and IntelliJ Platform versions.
 pluginSinceBuild=221.1
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@
 pluginGroup=com.sourcegraph.jetbrains
 pluginName=Sourcegraph
 # SemVer format -> https://semver.org
-pluginVersion=7.0-localbuild
+pluginVersion=6.0-localbuild
 # See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 # for insight into build numbers and IntelliJ Platform versions.
 pluginSinceBuild=221.1
--- a/src/main/kotlin/com/sourcegraph/cody/auth/ui/LoadingAccountsDetailsProvider.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/auth/ui/LoadingAccountsDetailsProvider.kt
@@ -9,6 +9,7 @@ import com.intellij.util.ui.EmptyIcon
 import com.sourcegraph.cody.auth.Account
 import com.sourcegraph.cody.auth.AccountDetails
 import com.sourcegraph.cody.auth.SingleValueModel
+import com.sourcegraph.cody.config.notification.AccountSettingChangeContext.Companion.UNAUTHORIZED_ERROR_MESSAGE
 import java.awt.Image
 import java.util.concurrent.CompletableFuture
 import javax.swing.Icon
@@ -41,9 +42,13 @@ abstract class LoadingAccountsDetailsProvider<in A : Account, D : AccountDetails
           }
           .exceptionally {
             val error = CompletableFutureUtil.extractError(it)
-            val errorMessage =
-                error.localizedMessage.takeWhile { c -> c.isLetterOrDigit() || c.isWhitespace() }
-            DetailsLoadingResult(null, null, errorMessage, false)
+            if (error.message == UNAUTHORIZED_ERROR_MESSAGE) {
+              DetailsLoadingResult(null, null, error.localizedMessage, true)
+            } else {
+              val errorMessage =
+                  error.localizedMessage.takeWhile { c -> c.isLetterOrDigit() || c.isWhitespace() }
+              DetailsLoadingResult(null, null, errorMessage, false)
+            }
           }
     }
   }
--- a/src/main/kotlin/com/sourcegraph/cody/auth/ui/SignInWithEnterpriseInstanceAction.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/auth/ui/SignInWithEnterpriseInstanceAction.kt
@@ -4,6 +4,7 @@ import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.actionSystem.PlatformCoreDataKeys
 import com.intellij.openapi.wm.ToolWindowManager
 import com.sourcegraph.cody.CodyToolWindowFactory
+import com.sourcegraph.cody.config.CodyAuthenticationManager
 import com.sourcegraph.cody.config.CodyPersistentAccountsHost
 import com.sourcegraph.cody.config.signInWithSourcegraphDialog
 import com.sourcegraph.common.ui.DumbAwareEDTAction
@@ -16,11 +17,13 @@ class SignInWithEnterpriseInstanceAction(
   override fun actionPerformed(e: AnActionEvent) {
     val project = e.project ?: return
     val accountsHost = CodyPersistentAccountsHost(project)
+    val authManager = project.getService(CodyAuthenticationManager::class.java)
+    val serverUrl = authManager.getActiveAccount()?.server?.url ?: defaultServer
     val dialog =
         signInWithSourcegraphDialog(project, e.getData(PlatformCoreDataKeys.CONTEXT_COMPONENT))
             .apply {
               contentPane.minimumSize = Dimension(MIN_DIALOG_WIDTH, MIN_DIALOG_HEIGHT)
-              setServer(defaultServer)
+              setServer(serverUrl)
             }
     if (dialog.showAndGet()) {
       accountsHost.addAccount(
--- a/src/main/kotlin/com/sourcegraph/cody/auth/ui/SignInWithEnterpriseInstanceAction.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/auth/ui/SignInWithEnterpriseInstanceAction.kt
@@ -17,7 +17,7 @@ class SignInWithEnterpriseInstanceAction(
   override fun actionPerformed(e: AnActionEvent) {
     val project = e.project ?: return
     val accountsHost = CodyPersistentAccountsHost(project)
-    val authManager = project.getService(CodyAuthenticationManager::class.java)
+    val authManager = CodyAuthenticationManager.getInstance(project)
     val serverUrl = authManager.getActiveAccount()?.server?.url ?: defaultServer
     val dialog =
         signInWithSourcegraphDialog(project, e.getData(PlatformCoreDataKeys.CONTEXT_COMPONENT))
--- a/src/main/kotlin/com/sourcegraph/cody/auth/ui/SignInWithEnterpriseInstanceAction.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/auth/ui/SignInWithEnterpriseInstanceAction.kt
@@ -17,10 +17,7 @@ class SignInWithEnterpriseInstanceAction(
     val project = e.project ?: return
     val accountsHost = CodyPersistentAccountsHost(project)
     val dialog =
-        signInWithSourcegraphDialog(
-                project,
-                e.getData(PlatformCoreDataKeys.CONTEXT_COMPONENT),
-                accountsHost::isAccountUnique)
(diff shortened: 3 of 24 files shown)


gradle.properties
```diff
@@ -3,7 +3,7 @@
 pluginGroup=com.sourcegraph.jetbrains
 pluginName=Sourcegraph
 # SemVer format -> https://semver.org
-pluginVersion=6.0-localbuild
```
mkondratek: It was annoying that `runIde` runs always showed update available notification. This is the fix for that.

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -108,6 +124,44 @@ class CodyAuthenticationManager(val project: Project) : Disposable {
...
+      val executor =
+          SourcegraphApiRequestExecutor.Factory.instance.create(activeAccount.server, token)
+      val progressIndicator = EmptyProgressIndicator(ModalityState.NON_MODAL)
+      val submitIOTask =
+          service<ProgressManager>().submitIOTask(progressIndicator) {
+            SourcegraphApiRequests.CurrentUser(executor, progressIndicator).getDetails()
+          }
+
+      submitIOTask.exceptionally { error ->
+        if (error.cause?.message == UNAUTHORIZED_ERROR_MESSAGE) { // invalid or expired access token
```
mkondratek: This is the best we can do. This call omits the agent (this is probably a good thing). The agent checks the token in the same way (https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/AuthProvider.ts?L253%3A1-335%3A1).

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -129,14 +183,15 @@ class CodyAuthenticationManager(val project: Project) : Disposable {
     else null
   }
 
-  fun setActiveAccount(account: CodyAccount?) {
+  fun setActiveAccount(account: CodyAccount?, previousToken: String?) {
     if (!project.isDisposed) {
       val previousAccount = getActiveAccount()
-      val previousToken = previousAccount?.let { getTokenForAccount(it) }
```
mkondratek, line 135: It is a possible and valid case to update the access token for an existing account. Previously we did not catch the access token change since at this point the access token is already updated. This is why I pass it as an argument.

src/main/kotlin/com/sourcegraph/cody/statusbar/CodyStatusService.kt
```diff
@@ -102,7 +102,9 @@ class CodyStatusService(val project: Project) : Disposable {
 
     @JvmStatic
     fun notifyApplication(project: Project, status: CodyStatus) {
-      getInstance(project).onCodyAutocompleteStatus(status)
+      if (!project.isDisposed) {
```
mkondratek, line 110: We did not see any errors related to this yet but this check makes sense.

src/main/kotlin/com/sourcegraph/cody/auth/ui/LoadingAccountsDetailsProvider.kt
```diff
@@ -41,9 +42,13 @@ abstract class LoadingAccountsDetailsProvider<in A : Account, D : AccountDetails
...
           .exceptionally {
             val error = CompletableFutureUtil.extractError(it)
-            val errorMessage =
-                error.localizedMessage.takeWhile { c -> c.isLetterOrDigit() || c.isWhitespace() }
-            DetailsLoadingResult(null, null, errorMessage, false)
+            if (error.message == UNAUTHORIZED_ERROR_MESSAGE) {
+              DetailsLoadingResult(null, null, error.localizedMessage, true)
+            } else {
+              val errorMessage =
+                  error.localizedMessage.takeWhile { c -> c.isLetterOrDigit() || c.isWhitespace() }
```
pkukielka, line 49: What about `$` in the class names? What exactly you try to trim there?

src/main/kotlin/com/sourcegraph/cody/auth/ui/SignInWithEnterpriseInstanceAction.kt
```diff
@@ -16,14 +17,13 @@ class SignInWithEnterpriseInstanceAction(
   override fun actionPerformed(e: AnActionEvent) {
     val project = e.project ?: return
     val accountsHost = CodyPersistentAccountsHost(project)
+    val authManager = project.getService(CodyAuthenticationManager::class.java)
```
pkukielka: Please use `CodyAuthenticationManager.getInstance`

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -108,6 +124,44 @@ class CodyAuthenticationManager(val project: Project) : Disposable {
...
+    isTokenInvalid = isTokenInvalidFuture
+
+    isTokenInvalidFuture.thenApply { newAuthStatus ->
+      if (previousIsTokenInvalid != newAuthStatus) {
+        publisher.afterAction(AccountSettingChangeContext(isTokenInvalidChanged = true))
+      }
+    }
+
+    val activeAccount = getActiveAccount()
+    val token = activeAccount?.let(::getTokenForAccount)
```
pkukielka, line 142: Isn't that the same as just `activeAccount?.getTokenForAccount`?

src/main/kotlin/com/sourcegraph/cody/auth/ui/LoadingAccountsDetailsProvider.kt
mkondratek: This is the previous behaviour. I only improved the handling for unauthorized error.

When I switch the internet connection of I see this:
image

The code is supposed to extract the root cause of the problem (but it does not work properly in many cases).

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
pkukielka: Nitpick, but you could replace 5 lines with one:
```kt
isTokenInvalidFuture.complete(error.cause?.message == UNAUTHORIZED_ERROR_MESSAGE)
```

src/main/kotlin/com/sourcegraph/cody/auth/ui/LoadingAccountsDetailsProvider.kt
```diff
@@ -41,9 +42,13 @@ abstract class LoadingAccountsDetailsProvider<in A : Account, D : AccountDetails
           }
           .exceptionally {
             val error = CompletableFutureUtil.extractError(it)
-            val errorMessage =
-                error.localizedMessage.takeWhile { c -> c.isLetterOrDigit() || c.isWhitespace() }
-            DetailsLoadingResult(null, null, errorMessage, false)
+            if (error.message == UNAUTHORIZED_ERROR_MESSAGE) {
+              DetailsLoadingResult(null, null, error.localizedMessage, true)
```
mkondratek, line 46: image

When the user is unauthorized there is an action to login 🚀

src/main/kotlin/com/sourcegraph/cody/config/ui/AccountConfigurable.kt
```diff
@@ -98,8 +109,15 @@ class AccountConfigurable(val project: Project) :
       activeAccount = accountsModel.accounts.getFirstAccountOrNull()
     }
 
-    CodyAuthenticationManager.getInstance(project).setActiveAccount(activeAccount)
-    accountsModel.activeAccount = activeAccount
+    val accountChanged = initialActiveAccount != activeAccount
+    val tokenChanged =
+        initialToken != activeAccount?.let { authManager.getTokenForAccount(activeAccount) }
+
+    if (accountChanged || tokenChanged) {
```
pkukielka: It looks like that check should be inside `setActiveAccount`

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -129,14 +183,15 @@ class CodyAuthenticationManager(val project: Project) : Disposable {
     else null
   }
 
-  fun setActiveAccount(account: CodyAccount?) {
+  fun setActiveAccount(account: CodyAccount?, previousToken: String?) {
```
pkukielka: Can you explain wht that change was necessary?
It looks a bit backwards to me, as one could expect that `CodyAuthenticationManager` should know about previous token instead of being told explicitly what it was.
I see that e.g. you pass `null` there for the initial set, but I do not see why it would be needed vs the old code.

mkondratek: `CodyAccount` does not store the token value. `CodyAccount` is only a key. The credentials (the token) is already updated when we call `setActiveAccount`. `com.sourcegraph.cody.config.CodyAccountManager` is responsible for the token storage.

pkukielka (approved):
LGTM, but please make sure it is manually tested again before merge

PR #1800, opened 2024-06-24
Respect plugin version form gradle.properties

pkukielka:
## Test plan

1. Build plugin using `./gradlew buildPlugin`
2. Unpack the plugin jar and check if versions are set correctly inside plugin.xml

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -9,18 +9,6 @@
     <change-notes>
         <![CDATA[Update notes are available on <a href="https://github.com/sourcegraph/jetbrains/releases">GitHub</a>.]]></change-notes>
 
-    <!--
-        See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
-        for insight into build numbers and IntelliJ Platform versions.
-        - 2020.2 was the first version to have JCEF enabled by default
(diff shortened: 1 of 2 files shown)


odisseus (approved):

src/main/resources/META-INF/plugin.xml
```diff
@@ -9,18 +9,6 @@
...
-    <!--
-        See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
-        for insight into build numbers and IntelliJ Platform versions.
-        - 2020.2 was the first version to have JCEF enabled by default
-            -> https://plugins.jetbrains.com/docs/intellij/jcef.html
-        - Version 2020.2 and 2020.3 have issues with adding custom HTTP headers to requests from within the JCEF view
-             -> https://github.com/sourcegraph/sourcegraph/issues/37475#issuecomment-1171355831
-        - Version 2022.1 is the version introducing the Kotlin UI DSL API we are using at the moment
-          - if the need arises, we may try to implement backwards compatibility to 2021.3
-    -->
```
mkondratek, line 21: shouldn't we move that comment to `gradle.properties` then?

pkukielka: I think that comment is not adding any value anymore anymore. We won't ever consider going to such old versions, so reason why we dumped them in a past is just a historical trivia.

PR #1805, opened 2024-06-25
added Document Synchronization integration tests

steveyegge:
This PR introduces a new set of integration tests for document synchronization testing.

This PR is dependent on a [Cody-side PR](https://github.com/sourcegraph/cody/pull/4512), and the tests will not pass until that PR is merged.

## Test plan

This is not ready to merge yet; I'll update this section when it's ready to go.


src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTestFixture.kt
```diff
@@ -29,13 +27,23 @@ import java.nio.file.Paths
...
 
   override fun setUp() {
     super.setUp()
-    configureFixture()
+    val methodName = name
+    val method =
+        this.javaClass.getMethod(methodName)
+            ?: throw IllegalStateException(
+                "No method with name $methodName found in ${this.javaClass.name}")
+    if (method.isAnnotationPresent(TestFile::class.java)) {
```
pkukielka, line 48: Maybe we could use the default test file if no annotation is added?

pkukielka (approved):
I added one comment, otherwise LGTM

src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTestFixture.kt
steveyegge: Good call; I'll clean that up.

src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
```diff
@@ -18,16 +19,25 @@ import org.junit.runners.Suite
  * automatically after the platform version bump.
  *
  * Multiple recording files can be used, but each should have its own suite with tearDown() method
- * nad define unique CODY_RECORDING_NAME.
+ * and define unique CODY_RECORDING_NAME.
```
dominiccooney: ...while we're here...

```suggestion
 * and define a unique CODY_RECORDING_NAME.
```

src/integrationTest/kotlin/com/sourcegraph/cody/DocumentSynchronizationTest.kt
```diff
@@ -0,0 +1,285 @@
...
+            class Foo {
+              console.log("hello there^")
+            }
+        """
+
+    val expectedContent =
+        """
+            class Foo {
+              console.log("hello there!🎉🎂
+              🥳🎈")
```
dominiccooney, line 242: Resulting indentation is interesting...

dominiccooney (approved):
Looks *good*. Note, one bit of feedback from @pkukielka you said you'll address.

src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTestFixture.kt
pkukielka: Looks like you missed this one @steveyegge

src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTextFixture.kt
```diff
@@ -42,7 +42,8 @@ open class CodyIntegrationTextFixture : BasePlatformTestCase() {
 
   override fun tearDown() {
     try {
-      FixupService.getInstance(myFixture.project).getActiveSession()?.apply {
```
pkukielka, line 45: Shouldn't that whole file be removed and that line modified in the `CodyIntegrationTextFixture.kt`?
Then git should correctly show it as rename, and not like now as whole new file added.

src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTestFixture.kt
steveyegge: Yeah, there really isn't a good choice for a default right now, since the only test file is specific to `DocumentCodeTest`. 

But I went ahead and used that as the default for now, until we get more tests with disk files.

steveyegge: I had to roll this change back because it broke all the new tests, which don't have testResources files on disk.

src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTextFixture.kt
steveyegge: Good catch; fixed that.

PR #1806, opened 2024-06-25
Check if project is disposed in CodyAuthenticationManager before notifying the message bus

mkondratek:
## Test plan
The issue seems to be hard to reproduce. Possibly it can appear when you close and reopen the project in the right moment. The right moment is b/w focusing the IDE window and sending the receiving the graphql response from the server.

---

Let's try to merge it quickly to unblock others. I will try to review the docs related to disposables and review the intellij's code. I have a feeling that we are doing smth wrong. We should not have to add so many checks.


pkukielka (approved):
LGTH, thanks for fixing!

Issue #1798, opened 2024-06-24
bug:  Unable to navigate using left cursor when selecting a block and cursor at end of line, started with new inline shortcut hint feature (nightly 6.0.5+)


jafin:
### IDE Information

JetBrains Rider 2024.1.3
Build #RD-241.17011.166, built on June 8, 2024
Runtime version: 17.0.11+1-b1207.24 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Windows 11.0
.NET Core v8.0.1 x64 (Server GC)
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 32
Registry:
  ide.experimental.ui=true
  ide.mac.bigsur.window.with.tabs.enabled=false
  ide.new.project.model.index.case.sensitivity=true

### Bug Description

Only appeared since nightly 6.0.5/6.0.6  Previous nightly and stable are fine.

Attempt to describe the issue.
press `Shift-End` to select line,  
Then attempting to refine the selection using `Shift-Left arrow` results in no action., i.e. the selection scope is ignoring the key.  Shift-Up/Down work ok.

pressing `Shift-Left` then `Shift-Right` also works, so it appears only when cursor is at line end does it break the usage of `Shift-Left`.

Disabling Cody plugin fixes the problem (or reverting to a non 6.0.5/6.0.6 version)

~~Unsure why the shortcut key is being shown to the right of the line?~~ This I believe has only started appearing in the 6.0.5+ nightly.

**EDIT**: I see this is a new "feature", and is expected. **Hopefully it can be toggled on/off** for the user as I personally find it not useful to always be shown.

This is 6.0.6 with cursor at end of line, cannot navigate left using shift-left arrow.

![image](https://github.com/sourcegraph/jetbrains/assets/127927/0d250a3d-c1b4-42d1-9e6f-5322ad359772)

This is 6.0.4 with cursor at end of line, can navigate using shift-left.  Also no shortkey notification visible.
![image](https://github.com/sourcegraph/jetbrains/assets/127927/7551dc8e-69c3-46c2-8a2b-485bfa755c97)

jafin:
Perhaps related to this change in 6.0.5  https://github.com/sourcegraph/jetbrains/commit/aed272c4d6bbf55c42f535ee92f4a58fb10fe66d

Just an observation, the cursor is at the end of the hint when it breaks, but the EOL is actually after the semicolon.
[...]

PR #1814, opened 2024-06-25
Fix issues CODY-2598 and CODY-2600

steveyegge:
We were showing the hotkey help inlay in inappropriate situations:

- in random panes/views that allow selections, such as the console window
- when an inline edit is currently active

## Test plan

Updated manual test plan to check for these two cases.


abeatrix (approved):

gauravkumar37:
Related: https://github.com/sourcegraph/jetbrains/issues/1798

PR #1818, opened 2024-06-26
Add support for cody.keyDown.Copy:clicked telemetry event

pkukielka:
Fixes [CODY-2572](https://linear.app/sourcegraph/issue/CODY-2572/p1-implement-codykeydowncopyclicked-event)

## Changes 

Added support for cody.keyDown.Copy:clicked telemetry event

## Test plan

1. Ask cody to generate new code
2. Copy that code using "Ctrl+C" or "Command+C" shortcut.
3. After 2h check if new event is visible in looker.

### DIFF
--- a/src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
+++ b/src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
@@ -6,6 +6,7 @@ import com.intellij.openapi.editor.*;
 import com.intellij.openapi.editor.event.*;
 import com.intellij.openapi.editor.ex.EditorEx;
 import com.intellij.openapi.fileEditor.FileEditorManager;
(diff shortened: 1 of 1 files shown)


odisseus:
I'm not sure about this. A tool that sends the contents of your clipboard is a nightmare from the security point of view.

I understand that this functionality is meant to send only the text which has been copied from the LLM response. Could there be any corner cases where the LLM response contains sensitive information?

mkondratek (approved):

pkukielka:
> I'm not sure about this. A tool that sends the contents of your clipboard is a nightmare from the security point of view.
> 
> I understand that this functionality is meant to send only the text which has been copied from the LLM response. Could there be any corner cases where the LLM response contains sensitive information?

Look at the implementation of the `sendCodeGenerationEvent`, we are not sending any content, only generic data about text like number of character and lines.

PR #1845, opened 2024-06-28
Do not create AboutDialog off EDT

dominiccooney:
AboutDialog needs to be invoked from EDT in 2024.1.

## Test plan

Manual test:

1. Set JetBrains to connect to a spawned agent.
2. Set a breakpoint in the spawned agent, causing startup to timeout.
3. Check that the error report link generation callstack does NOT appear in IDE errors.


pkukielka (approved):
LGTM

PR #1858, opened 2024-07-01
Merge PersistentActiveAccountHolder into CodyAuthenticationManager

mkondratek:
Fixes https://linear.app/sourcegraph/issue/CODY-2666/merge-codyprojectactiveaccountholder-and-codyauthenticationmanager.

## Test plan

Authentication 
- log in for the first time 
- having multiple accounts
- logging into the same account twice
- switching accounts b/w pro, free, enterprise


src/main/kotlin/com/sourcegraph/cody/auth/SelectOneOfTheAccountsAsActive.kt
```diff
@@ -3,24 +3,19 @@ package com.sourcegraph.cody.auth
 import com.intellij.openapi.project.Project
 import com.sourcegraph.cody.CodyToolWindowContent
 import com.sourcegraph.cody.config.CodyAuthenticationManager
-import com.sourcegraph.cody.config.CodyProjectActiveAccountHolder
 import com.sourcegraph.cody.config.getFirstAccountOrNull
 import com.sourcegraph.cody.initialization.Activity
 
 class SelectOneOfTheAccountsAsActive : Activity {
```
pkukielka: Cannot we now move that whole code to `CodyAuthenticationManager::loadState`?

mkondratek: good point, lemme test it...

pkukielka (approved):
LGTM, thanks!
Also `+84 −155` ❤️

src/main/kotlin/com/sourcegraph/cody/auth/SelectOneOfTheAccountsAsActive.kt
mkondratek: needed to override noStateLoaded to get it working without the active account set initially (loadState executes only if there is a state)

docs:
https://plugins.jetbrains.com/docs/intellij/persisting-state-of-components.html#customizing-the-xml-format-of-persisted-values

Issue #117, opened 2023-11-13
Ask Cody to fix inspection


olafurpg:
As a user, I want Cody to apply a fix to errors detected in my project.

* A fix can be triggered by opening a Code Inspection and selecting a new option to "Ask Cody to Fix"
* Cody runs an "Edit" command in the background, with instructions to fix the error.
* In an open project file, Cody automatically applies a code fix to the selected code.
* Users can accept, edit and retry, see a diff view of the vix, or undo the fix.
* Executing the "fix" command fires a telemetry event aligned with the current VSC implementation.

  <img src="https://uploads.linear.app/15b96a4d-ffc0-481a-ba9c-9d840020f7bc/5efb7518-2135-4f43-b3af-b6c622e28b06/638176b6-c937-460d-afda-aeabde77c1dc?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzE1Yjk2YTRkLWZmYzAtNDgxYS1iYTljLTlkODQwMDIwZjdiYy81ZWZiNzUxOC0yMTM1LTRmNDMtYjNhZi1iNmM2MjJlMjhiMDYvNjM4MTc2YjYtYzkzNy00NjBkLWFmZGEtYWVhYmRlNzdjMWRjIiwiaWF0IjoxNzE4NzE3OTQ4LCJleHAiOjMzMjg5Mjc3OTQ4fQ.bR52ycIfoEMIsIraZFF0XXAnNHNHj5Q61oBvXP8IstE " alt="Screenshot 2024-06-18 at 7.38.51 AM.png" width="527" height="564" />

taylorsperry:
Closing as a dupe of https://github.com/sourcegraph/jetbrains/issues/103

PR #1859, opened 2024-07-01
Fix problems action

odisseus:
Partially implements #117.

### To do before merging
* Implement the `editCommands/fix` request on the agent side

### To do in a follow-up PR
* Code inspection and "Quick fix" annotation

### Out of scope 
* Hotkey and icon for the new action

## Test plan

TODO

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -294,6 +294,11 @@
             <override-text place="GoToAction" text="Cody: Generate Unit Tests"/>
         </action>
 
+        <action id="cody.fixProblemAction"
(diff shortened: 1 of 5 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -230,6 +230,11 @@
             <keyboard-shortcut replace-all="true" first-keystroke="ctrl alt G" keymap="Mac OS X 10.5+"/>
         </action>
 
+        <action id="cody.fixProblemAction"
```
pkukielka, line 297: It will also need a button on the Commands tab.

odisseus:
Superseded by #1890.

PR #1861, opened 2024-07-02
Add telemetry for rate limits notifications and upsell

pkukielka:
Closes https://linear.app/sourcegraph/issue/CODY-2570/p1-add-telemetry-events-usage-limits

## Test plan

**Scenario 1**

1. Switch to rate limited free account
2. Try to do an autocomplete
3. Verify that upsell notification is shown
4. Verify that `cody.upsellUsageLimitCTA:shown` telemetry event was fired
5. Click 'Upgrade' button
6. Verify that `cody.upsellUsageLimitCTA:clicked` event is fired

**Scenario 2**

Do the same for the chat.

**Scenario 3**

Do the same for both auto-complete and chat, but instead of clicking on the notification dismiss it, and open it again using status bar:
image
Verify that `cody.upsellUsageLimitStatusBar:shown` and `cody.upsellUsageLimitStatusBar:clicked` fires accordingly.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/HtmlViewer.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/HtmlViewer.kt
@@ -1,16 +1,21 @@
 package com.sourcegraph.cody.ui
 
+import com.intellij.openapi.project.Project
 import com.intellij.ui.BrowserHyperlinkListener
 import com.intellij.util.ui.HTMLEditorKitBuilder
 import com.intellij.util.ui.JBInsets
 import com.intellij.util.ui.SwingHelper
 import com.sourcegraph.cody.chat.ChatUIConstants
+import com.sourcegraph.cody.telemetry.TelemetryV2
+import com.sourcegraph.common.CodyBundle
 import java.awt.Insets
+import java.net.URL
 import javax.swing.JEditorPane
+import javax.swing.event.HyperlinkEvent
 
 object HtmlViewer {
   @JvmStatic
-  fun createHtmlViewer(): JEditorPane {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/sourcegraph/cody/ui/HtmlViewer.kt
```diff
@@ -21,7 +26,16 @@ object HtmlViewer {
                 ChatUIConstants.TEXT_MARGIN,
                 ChatUIConstants.TEXT_MARGIN,
                 ChatUIConstants.TEXT_MARGIN))
-    jEditorPane.addHyperlinkListener(BrowserHyperlinkListener.INSTANCE)
+    jEditorPane.addHyperlinkListener(MyBrowserHyperlinkListener(project))
     return jEditorPane
   }
+
+  class MyBrowserHyperlinkListener(val project: Project) : BrowserHyperlinkListener() {
```
mkondratek, line 33: Maybe `MyBrowserHyperlinkListener` could be a project level service? 🤔 that would allow us to use it without passing project via `CodyOnboardingGuidancePanel` and `createHtmlViewer`.

mkondratek: Or maybe TelemetryV2 should be a project level service. It seems to make sense

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/ui/HtmlViewer.kt
pkukielka: unfortunately it would not change anything because you would still need provide project for `getInstance`.

PR #1862, opened 2024-07-02
Manual Stable Release

mkondratek:
Our current setup produces a lot of stable releases that often are not unhidden eventually. That makes the QA and awaiting JB approval process parallel. Based on the recent experience only 1/8 releases goes public. That is a waste of JB approval team's time and CI time.

Let's change the setup. The nightly version handling does not change but the default release script does not longer publish the stable version. Instead the separate workflow can be triggered on the specific tag that publishes the stable version.

## Test plan
1. push a tag for a nightly release
2. trigger the stable release manually

to be tested once merged

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -29,7 +29,6 @@ jobs:
           echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)" >> $GITHUB_ENV
       - run: echo "Publishing version $RELEASE_VERSION"
       - name: Publish nightly version
-        if: "!endsWith(env.RELEASE_VERSION, '-nightly')"
         run: |
           echo "Publishing nightly version ${RELEASE_VERSION}-nightly"
           ./gradlew "-PpluginVersion=${RELEASE_VERSION}-nightly" publishPlugin
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,0 +1,37 @@
+name: Nightly Release
+on:
+  push:
+    tags: [ "*" ]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -117,28 +117,33 @@ After doing that:
 
 ## Publishing a New Release
```
mkondratek, line 118: Preview: 
https://github.com/sourcegraph/jetbrains/blob/848542ee2ad069caec2774fedcf40444be922c97/CONTRIBUTING.md#publishing-a-new-release

.github/workflows/nightly-release.yml
```diff
@@ -28,9 +28,6 @@ jobs:
       - run: |
           echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)" >> $GITHUB_ENV
       - run: echo "Publishing version $RELEASE_VERSION"
-      - run: ./gradlew "-PpluginVersion=$RELEASE_VERSION" publishPlugin
-        env:
-          PUBLISH_TOKEN: ${{ secrets.PUBLISH_TOKEN }}
       - name: Publish nightly version
         if: "!endsWith(env.RELEASE_VERSION, '-nightly')"
```
odisseus: Is this condition still needed?

scripts/next-release.sh
```diff
@@ -1,10 +1,4 @@
 #!/usr/bin/env bash
-# This script implements the time-based version scheme from RFC 795
```
dominiccooney, line 2: Why delete this, particularly since client releases are moving to be more aligned with Sourcegraph releases, not less?

dominiccooney (approved):
LGTM

Agree with @odisseus that the -nightly suffix check can be simplified

Some more feedback inline

scripts/next-release.sh
mkondratek: The script does not work like the comment says. Also, we wanted "prettier" version numbers for GA.

odisseus (approved):

PR #1867, opened 2024-07-02
Show the login screen when the token is missing

mkondratek:
## Test plan
Delete the token from the keychain externally 
You can run a separate instance of IDE and remove the account there.

1. Log in to the account
2. Open another IDE, log in to the same account, apply, then remove the account, apply
3. Switch back to the first ide
4. Close and reopen the project
5. Login panel visible


odisseus (approved):

pkukielka (approved):
LGTM

PR #1872, opened 2024-07-03
Improve action discoverability by adding Cody: prefix in Search Everywhere

pkukielka:
Closes https://linear.app/sourcegraph/issue/CODY-2130/improve-how-actions-appear-in-shiftshift-menu

## Changes

I moved action properties like `text` and `description` to `plugin.xml`.
It is just easier to edit and inspect this way.
To be honest, Geminit 1.5 did most of the work:

> Given the following plugin.xml file and properties file, please modify plugin.xml so each action have additional propertes `text` and `description` with values from properties file.
> Additionally for every action please add such tag: `<override-text place="GoToAction" text="Cody: $TEXT"/>` where `$TEXT` is taken from properties file as well.
> For some actions like: `<action id="Cody.Accounts.LogInToSourcegraphAction" class="com.sourcegraph.cody.config.LogInToSourcegraphAction"/>` you may have to change them to form such as:
> ```
> <action id="Cody.Accounts.LogInToSourcegraphAction" class="com.sourcegraph.cody.config.LogInToSourcegraphAction" text="Log In to Sourcegraph">
>   <override-text place="GoToAction" text="Cody: Log In to Sourcegraph"/>
> </action>
>```
>
> Please output whole content of the modified plugin.xml file.

## Test plan

1. Hit `Shift shift`
2. Type 'Cody:'

image

### DIFF
--- a/src/main/resources/CodyBundle.properties
+++ b/src/main/resources/CodyBundle.properties
@@ -75,7 +75,6 @@ UpgradeToCodyProNotification.content.upgrade=\
     (Already upgraded to Pro? Restart your IDE for changes to take effect)\
   </html>
 UpgradeToCodyProNotification.content.explain=To ensure that Cody can stay operational for all Cody users, please come back tomorrow for more chats, commands, and autocompletes.
-context-panel.button.edit-repositories=Edit Remote Repositories
 context-panel.button.reindex=Reindex Local Project
 context-panel.button.help=Help
 context-panel.in-progress=Running Cody 'Keyword Search' indexer...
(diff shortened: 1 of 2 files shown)


odisseus:
I see that the action names are now specified in plugin.xml rather than in the properties file. What's the reason behind that change?

src/main/resources/CodyBundle.properties
```diff
@@ -142,49 +141,6 @@ group.CodyStatusBarActions.text=Cody
 group.CodyEditorActions.text=Cody
 group.InternalsStatusBarActions.text=?? Internals
 
-# Authentication Actions
-action.Cody.Accounts.LogInToSourcegraphAction.text=Log In to Sourcegraph
```
odisseus, line 146: I think it is better to keep these strings in a properties file for the sake of [internationalisation](https://plugins.jetbrains.com/docs/intellij/internationalization.html#message-bundles), but this is more of a theoretical concern.

odisseus (approved):

src/main/resources/CodyBundle.properties
pkukielka: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it
:)

mkondratek: Maybe the i18n is not a good reason but keeping all user-facing strings in one place seems to be a rational objective.

PR #1888, opened 2024-07-05
Add 'Authenticate in Browser' in sign-in dialog

mkondratek:
Resolves https://linear.app/sourcegraph/issue/CODY-2308/enterprise-users-unable-to-use-browser-redirects-for-authentication-on.

## Test plan
1. Go to Settings
2. `Log In with Token to Sourcegraph Enterprise`
3. Try different server inputs: `sourcegraph.com`, `sourcegraph.sourcegraph.com`, invalid server url, empty server url 
4. `Authenticate automatically`

## Demo

~https://github.com/sourcegraph/jetbrains/assets/19799111/0f64b145-96ae-4952-a168-5f811e0a0ec1~

https://github.com/user-attachments/assets/a2f0e950-5b4d-4c32-acdb-8b5971c171da

## Final look

image


dominiccooney (commented):
Thank you for working on this.

Technical aspects look good to me, but the design needs work. We should defer to @danielmarquespt , but here's my feedback:

- Let's not call it "Authenticate automatically" because it isn't automatic... you still need to click things. Just call it "Authenticate".
- Let's make the new path the default, that is, remove the token box.
- Let's use extra language for the manual flow, for example, maybe there's a link label where the token box was that says "Manually enter a token..." or something like that.

src/main/kotlin/com/sourcegraph/cody/config/BaseLoginDialog.kt
```diff
@@ -10,20 +9,54 @@ import com.intellij.openapi.project.Project
...
 
   protected val loginPanel = CodyLoginPanel(executorFactory)
 
+  private val authenticateAutomatically =
+      object : DialogWrapperAction("Authenticate automatically") {
+
+        override fun doAction(e: ActionEvent?) {
+          val info = loginPanel.getServerPathValidationInfo()
+          if (info != null) {
+            if (info.component != null && info.component!!.isVisible) {
```
pkukielka: It is a really minor nit, but kotlin forces you to use `!!` because it cannot quarantine `null` safety with code written this way.

Instead you could do:

```kotlin
        override fun doAction(e: ActionEvent?) {
          val info = loginPanel.getServerPathValidationInfo()
          val component = info?.component
          if (component != null) {
            if (component.isVisible) {
              IdeFocusManager.getInstance(project).requestFocus(component, true)
            }

            updateErrorInfo(listOf(info))
            this.isEnabled = info.okEnabled
            startTrackingValidation()
            if (!info.okEnabled) {
              return
            }
          }
          authenticate()
        }
```

src/main/kotlin/com/sourcegraph/cody/config/BaseLoginDialog.kt
```diff
@@ -10,20 +9,54 @@ import com.intellij.openapi.project.Project
...
+              IdeFocusManager.getInstance(null).requestFocus(info.component!!, true)
```
pkukielka: Shouldn't it be `IdeFocusManager.getInstance(project)?`

danielmarquespt:
Hi @mkondratek thank you for working on this! I propose a slightly different flow and promotes the browser path:

Clicking [Advanced...] closes the simplified dialog, and opens the old dialog with some keyword changes

login flow

src/main/kotlin/com/sourcegraph/cody/config/CodyAccountListModel.kt
```diff
@@ -35,11 +35,11 @@ class CodyAccountListModel(private val project: Project) :
             .login(
                 parentComponent,
                 CodyLoginRequest(
-                    login = account.name,
+                    title = "Edit Sourcegraph Account",
```
pkukielka, line 38: Should be in the bundle? :)

src/main/kotlin/com/sourcegraph/cody/config/SourcegraphInstanceLoginDialog.kt
```diff
@@ -0,0 +1,248 @@
...
+              JBLabel("Logging in, check your browser").apply {
+                icon = AnimatedIcon.Default.INSTANCE
+                foreground = getInactiveTextColor()
+              })
+        }
+        .visibleIf(isAcquiringToken.selected)
+    row("Instance URL:") {
+          textField()
+              .applyToComponent {
+                emptyText.text = "https://sourcegraph.yourcompany.com"
```
odisseus: This is an actual link! Currently it is just a parked domain, but it might [end up](https://www.bbc.com/news/technology-20627460) pointing to NSFW content sometime in the future.

src/main/kotlin/com/sourcegraph/cody/config/BaseLoginDialog.kt
mkondratek: n/a

src/main/kotlin/com/sourcegraph/cody/config/SourcegraphInstanceLoginDialog.kt
mkondratek: fixed

pkukielka: I think it looks worse now but I guess everyone have different taste ;)
FWIW it was not clickable, so there was no risk it will redirect someone anywhere.

src/main/resources/CodyBundle.properties
```diff
@@ -169,3 +169,27 @@ settings.migration.llm-upgrade-notification.body=\
...
+
+# Authentication
+login.dialog.title=Add Sourcegraph Account
+login.dialog.authorize-in-browser=Authorize in Browser
+login.dialog.add-account=Add Account
+login.dialog.show-advanced=Show Advanced...
+login.dialog.hide-advanced=Hide Advanced...
+login.dialog.check-browser=Logging in, check your browser
+login.dialog.instance-url.label=Instance URL:
+login.dialog.instance-url.comment=Enter the address of your sourcegraph instance. For example https://sourcegraph.<yourcompany>.org
```
odisseus: Looks like IntelliJ parses HTML tags in these strings. Try replacing the angle brackets with `&lt;` and `&gt;`.

mkondratek: done, thanks

src/main/kotlin/com/sourcegraph/cody/config/BaseLoginDialog.kt
```diff
@@ -13,17 +12,20 @@ import com.intellij.openapi.util.Disposer
 import com.sourcegraph.cody.api.SourcegraphApiRequestExecutor
 import com.sourcegraph.cody.auth.SsoAuthMethod
 import java.awt.Component
+import javax.swing.Action
 import javax.swing.JComponent
 
 abstract class BaseLoginDialog(
-    project: Project?,
-    parent: Component?,
+    val project: Project?,
```
odisseus: Can these fields be made protected or private?

src/main/kotlin/com/sourcegraph/cody/config/SourcegraphInstanceLoginDialog.kt
```diff
@@ -0,0 +1,260 @@
...
+import java.awt.event.ActionEvent
+import java.net.UnknownHostException
+import java.util.concurrent.CompletableFuture
+import javax.swing.Action
+import javax.swing.JCheckBox
+
+class SourcegraphInstanceLoginDialog(
+    project: Project?,
+    parent: Component?,
+    private val defaultInstanceUrl: String = ""
```
odisseus, line 50: What is the purpose of this field? An empty string doesn't seem to be a reasonable default URL for anything.

src/main/resources/CodyBundle.properties
```diff
@@ -169,3 +169,27 @@ settings.migration.llm-upgrade-notification.body=\
...
+
+# Authentication
+login.dialog.title=Add Sourcegraph Account
+login.dialog.authorize-in-browser=Authorize in Browser
+login.dialog.add-account=Add Account
+login.dialog.show-advanced=Show Advanced...
+login.dialog.hide-advanced=Hide Advanced...
+login.dialog.check-browser=Logging in, check your browser
+login.dialog.instance-url.label=Instance URL:
+login.dialog.instance-url.comment=Enter the address of your sourcegraph instance. For example https://sourcegraph.&lt;yourcompany&gt;.org
```
odisseus: Sourcegraph should probably be capitalised.

src/main/kotlin/com/sourcegraph/cody/config/SourcegraphInstanceLoginDialog.kt
mkondratek: if the default is empty we show the empty text (e.g. `sourcegraph.<yourcompany>.com`)

odisseus (approved):

taylorsperry:
@mkondratek is this ready to merge?

mkondratek:
> @mkondratek is this ready to merge?

@taylorsperry, sorry for the delay - done

PR #1890, opened 2024-07-08
Support for the "Fix problems" API

odisseus:
Partially implements #117. See also #1859.

## Test plan

TODO


pkukielka:
@odisseus should we close this one?

PR #1894, opened 2024-07-09
feat(chore): Adding tags and usage for LLM Models

jamesmcnamara:
[CODY-2746](https://linear.app/sourcegraph/issue/CODY-2746/confirm-chat-dropdown-is-available-on-jetbrains-as-well)

This update is to align with a [refactor](https://github.com/sourcegraph/cody/pull/4780) in the vscode repo around model types to remove certain hard coded fields and replace them with ModelTags.

This is very rough around the edges because I don't know how to access the generated code in that PR (nor am I aware of the context on this repo) so any feedback is welcome!

## Test plan
Green CI

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
@@ -28,7 +28,7 @@ class LlmDropdown(
     val parentDialog: EditCommandPrompt?,
     val chatModelProviderFromState: ChatModelsResponse.ChatModelProvider?,
 ) : ComboBox<ChatModelsResponse.ChatModelProvider>(MutableCollectionComboBoxModel()) {
-  var hasServerSentModels = false 
+  var hasServerSentModels = false
 
   init {
     renderer = LlmComboBoxRenderer(this)
@@ -50,7 +50,7 @@ class LlmDropdown(
 
   private fun subscribeToFeatureUpdates() {
     val currentConfigFeatures: CurrentConfigFeatures =
-    project.getService(CurrentConfigFeatures::class.java)
+        project.getService(CurrentConfigFeatures::class.java)
     currentConfigFeatures.attach(::handleConfigUpdate)
   }
 
@@ -64,7 +64,7 @@ class LlmDropdown(
 
   @RequiresEdt
   private fun updateModelsInUI(models: List<ChatModelsResponse.ChatModelProvider>) {
-    if (project.isDisposed) return 
+    if (project.isDisposed) return
     this.removeAllItems()
 
     val availableModels = models.filterNot { it.isDeprecated() }
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
@@ -49,8 +49,8 @@ class LlmDropdown(
   private fun updateModelsInUI(models: List<ChatModelsResponse.ChatModelProvider>) {
     if (project.isDisposed) return
 
-    val availableModels = models.filterNot { it.deprecated }
-    availableModels.sortedBy { it.codyProOnly }.forEach(::addItem)
+    val availableModels = models.filterNot { it.isDeprecated() }
+    availableModels.sortedBy { it.isCodyProOnly() }.forEach(::addItem)
 
     val selectedFromState = chatModelProviderFromState
     val selectedFromHistory = HistoryService.getInstance(project).getDefaultLlm()
@@ -59,7 +59,8 @@ class LlmDropdown(
             ?: availableModels.find { it.model == selectedFromHistory?.model }
 
     selectedItem =
-        if (selectedModel?.codyProOnly == true && isCurrentUserFree()) availableModels.getOrNull(0)
+        if (selectedModel?.isCodyProOnly() == true && isCurrentUserFree())
+            availableModels.getOrNull(0)
         else selectedModel
 
     val isEnterpriseAccount =
@@ -83,7 +84,7 @@ class LlmDropdown(
     if (project.isDisposed) return
     val modelProvider = anObject as? ChatModelsResponse.ChatModelProvider
     if (modelProvider != null) {
-      if (modelProvider.codyProOnly && isCurrentUserFree()) {
+      if (modelProvider.isCodyProOnly() && isCurrentUserFree()) {
         BrowserOpener.openInBrowser(project, "https://sourcegraph.com/cody/subscription")
         return
       }
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
@@ -63,13 +63,10 @@ class LlmDropdown(
             availableModels.getOrNull(0)
         else selectedModel
 
-    val isEnterpriseAccount =
-        CodyAuthenticationManager.getInstance(project).account?.isEnterpriseAccount() ?: false
-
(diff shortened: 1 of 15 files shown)


src/main/kotlin/com/sourcegraph/cody/history/state/LLMState.kt
```diff
@@ -13,12 +13,18 @@ class LLMState : BaseState() {
 
   @get:OptionTag(tag = "provider", nameAttribute = "") var provider: String? by string()
 
+  @get:OptionTag(tag = "tags", nameAttribute = "") var tags: MutableList<String> by list()
+
+  @get:OptionTag(tag = "usage", nameAttribute = "") var usage: MutableList<String> by list()
+
```
jamesmcnamara, line 19: I don't know the implications of this. Where is it getting this state from? Is this going to break immediately?

odisseus:
In order for the tests to work, you'll need to update the `cody.commit` field in `gradle.properties`.

src/main/kotlin/com/sourcegraph/cody/history/state/LLMState.kt
pkukielka: This is state saved in a IntelliJ xml settings files.
You will need to add migration which will convert old format to the new one .
For the reference you can look at the `DeprecatedChatLlmMigration`.
 
In practice you will most likely have to:
1. Keep `ChatModelProvider::deprecated` field but mark it as `@Deprecated`.
2. In the migration function (see `DeprecatedChatLlmMigration::migrate` for reference):
- read all the old chats
- for each chat read the content from `deprecated` field
- set `tags` appropriately
- reset `deprecated` to be `null` (as it won't be used anymore, we will remove it in future)

src/test/kotlin/com/sourcegraph/cody/config/SettingsMigrationTest.kt
```diff
@@ -190,22 +198,20 @@ class SettingsMigrationTest : BasePlatformTestCase() {
   fun `test DeprecatedChatLlmMigration`() {
```
pkukielka, line 195: That is incorrect I think.
That is migration function, so it specifically migrates data form one old format to another.
It needs to work after your changes.
Then your migration function can pick up state after all existing migrations, and migrate it further (so you will have to add new test for it)

jamesmcnamara: Understood. I reverted those changes and added a new test for my migration.

src/main/kotlin/com/sourcegraph/cody/history/state/LLMState.kt
jamesmcnamara: Alright. I created `ChatTagsLllmMigration` that I believe accomplishes this.

src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
```diff
@@ -59,8 +59,8 @@ class LlmDropdown(
             ?: availableModels.find { it.model == selectedFromHistory?.model }
 
     selectedItem =
-        if (selectedModel?.codyProOnly == true && isCurrentUserFree())
-            availableModels.find { it.default }
+        if (selectedModel?.isCodyProOnly() == true && isCurrentUserFree())
+            availableModels.getOrNull(0)
```
pkukielka, line 81: Out of curiosity, if we do not have `default` anymore isn't there any tag which would replace it?
Like recommended LLM?

src/main/kotlin/com/sourcegraph/cody/history/state/LLMState.kt
pkukielka: Thank you! It looks good to me now, I just added one question about the `default` but otherwise I think it is mergeable.

pkukielka (approved):
LGTM, but with two caveats:
1. Isn't `default` replace by any tag? Don't we have any LLMs which we recommend to use?
2. Are those changes already in the Cody?
We bind cody version we are using by setting `cody.commit` there: https://github.com/sourcegraph/jetbrains/blob/main/gradle.properties#L27
Please make sure version from this commit already contains changes to the model. If not, you need to bump that commit in this PR (you can just set it to the hash of the last commit on Cody `main`)

src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
jamesmcnamara: I've just replaced it with the fallback convention of the [head of the list](https://github.com/sourcegraph/cody/pull/4780/files#diff-fe82b352761565cc5b6eb38a881a5bda7c83ad2c6b9231fa4bb25364154cc407R272-R277). The model structs are immutable having a mutable field tracking the user's selection was a bit tricky and also easy to get wrong (multiple default's was possible). And we always fell back to just grabbing the head of the list if we couldn't find one.

I could see adding a default tag, but I would think to only use it for what the server originally sent as the default, not as the user's current selection.

jamesmcnamara:
@pkukielka 
1) default is now the head of the list always
2) I will update `cody.commit` when [this PR](https://github.com/sourcegraph/cody/pull/4780) merges.

pkukielka:
> @pkukielka
> 
> 1. default is now the head of the list always
> 2. I will update `cody.commit` when [this PR](https://github.com/sourcegraph/cody/pull/4780) merges.

Thanks! BTW. Looks like it's merged now :)

src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
```diff
@@ -59,8 +59,8 @@ class LlmDropdown(
...
         else selectedModel
 
     val isEnterpriseAccount =
```
pkukielka, line 84: @jamesmcnamara 
1. We need to remove this line and just set `isVisible = true` always.
2. Also, I noticed that there is now `enterprise` tag present, shouldn't we filter out LLMs displayed for enterprise users based on if that tag is present? Or maybe for an enterprise account we are guaranteed to get a correct list?

pkukielka: I pushed first change to your branch already.

I'm not sure if we need the second.

jamesmcnamara: I updated this because we don't want it to be visible for enterprise users that don't have multiple models enabled. However I'm still struggling with the integration tests failing and I'm not sure how to debug them. Do you know any good tricks for figuring out why they keep timing out? I tried opening the IDE to the test file and trying the document code action but it seemed to work fine.

To your second point, I think the current idea is that an enterprise user can use any model that the server provided (but most likely they will all be enterprise models).

Also, unrelated, but do you know how to access the generated kotlin code from the Cody repo? I don't see it referenced anywhere.

pkukielka: Problem with integration tests was that updated recordings were missing (requires run of the `./gradlew :recordingIntegrationTest`). I updated them.
Interestingly it seems we were not reporting that fact properly, I will make sure we fixed that.

> Also, unrelated, but do you know how to access the generated kotlin code from the Cody repo? I don't see it referenced anywhere.

Right now they are just working as a manual reference, BUT it will change very soon as there is @RXminuS [PR in flight ](https://github.com/sourcegraph/jetbrains/pull/1898)which adress that.

pkukielka:
I updated recordings for integration tests + reformatted the code to make spotless happy.
I also tested it manually with pro account and two enterprise accounts (demo and gs02) and now it seems to be working perfectly fine.

Thanks @jamesmcnamara and @abeatrix for fixing bug with the auth and models!

PR #1895, opened 2024-07-09
Several fixes for Hotkey Inlay Hints

steveyegge:
Based on user feedback from GTM and Product, discussions with @danielmarquespt , and my own testing, I've made the following round of fixes to the hotkey hint feature, which shows an inlay with "Ctrl + Alt + ↵ to Edit" at the end of your selections in the editor window:

1. There is a new Preferences option for disabling the feature:
image

2. The hint no longer draws if the selection is only on one line.
  - this takes care of many annoying cases, such as Find/Replace lighting it up everywhere
 
3. The hint now draws below the last line of the selection.

This makes it less intrusive, IMO, as it's out of the way of the cursor.

I think it's also more intuitive. Several users claim that it "used to work that way". It never actually worked that way; it always drew at the end of the last line of the selection. But now it does work that way.

The one downside is that the hint can sometimes draw rather far from the selection, appearing to be two lines below. We will see how people respond to it.

Coincidentally and usefully, this change also fixes CODY-2541. Getting the hint inlay out of the way of the cursor removed the interference with fine-tuning the selection.

These changes also appear have to fixed CODY-2748, because I reported it and I can no longer reproduce it. It used to be 100% reproducible, so I'm going to mark it closed until it resurfaces.

fixes CODY-2541
fixes CODY-2748

## Test plan

This requires UI testing to verify with actual integration tests.
I reviewed the test plan, and these changes do not affect the plan.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
@@ -11,6 +11,7 @@ import com.intellij.openapi.editor.markup.TextAttributes
 import com.intellij.openapi.keymap.KeymapManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Disposer
+import com.sourcegraph.cody.config.CodyApplicationSettings
 import com.sourcegraph.cody.edit.FixupService
 import java.awt.Font
 import java.awt.Graphics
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/listeners/CodySelectionInlayManager.kt
```diff
@@ -41,17 +41,22 @@ class CodySelectionInlayManager(val project: Project) {
     val startOffset = event.newRange.startOffset
     val endOffset = event.newRange.endOffset
     if (startOffset == endOffset) {
-      // Don't show if there's no selection.
-      return
+      return // Don't show if there's no selection.
```
pkukielka, line 45: @steveyegge Shouldn't `isCodyUIHintsEnabled` be also checked somewhere in this code block?
I fail to see any place where it is actually used, but perhaps I miss something obvious

steveyegge: Damn, good catch. I had it at one point, and a merge killed it. Fixed.

pkukielka (approved):
LGTM

PR #1898, opened 2024-07-10
Prepare for Eventual Protocol Migrations.

RXminuS:
Fixes CODY-2830

At the moment we rely on manually written protocol messages and server/client implementations. This is a first step towards making sure that we can migrate towards auto generated bindings. 

More importantly this was done now because for the "Fix It" feature a few new protocols need to be implemented. So rather than digging a deeper hole this now allows to implement those in a way that will make migrating easier once all the other messages have been converted as well. 

So to summarize, this isn't so much about migrating everything now; rather making sure that everything new added doesn't make the work to migrate larger. I created a [placeholder project](https://linear.app/sourcegraph/project/jetbrains-auto-generated-bindings-129aac5aaa30/issues) for it so I can collect issues for the actual migration there.

For now I've only "migrated" a few low-hanging messages that were easy to do more as an example than anything else. No functionality should have changed. Couldn't quite verify though given the e2e test flake. Also added a simple helper to make working with Enums a bit nicer.

Right now there's no rigid testing / verification yet that the copied protocols are of the exact version mentioned etc. Instead there's a ad-hoc task/script that you can run with `./scripts/copy-protocol.sh` that will copy the protocols from CODY_DIR (or by default `<git_root>/../cody/` )

## Test plan

- No functionality should have changed. Existing tests pass
- Added a test to ensure that for the "SubsetOfGeneratedProtocolMessages" that we have migrated they are kept identical to the generated protocol.


build.gradle.kts
```diff
@@ -341,12 +344,74 @@ tasks {
...
   }
 
+  fun copyProtocol(): Unit {
+    // This is just a temporary solution to make it easier to use certain generated protocol
+    // messages.
+    // Ultimately this should entirely be replaced by use of the generated protocol.
+    if (!isForceProtocolCopy) {
+      return
+    }
+    val fromEnvironmentVariable = System.getenv("CODY_DIR")
```
pkukielka: You can use `downloadCody()` instead.
It uses cody from `CODY_DIR` if defined, if not then downloads cody specified by [cody.commit](https://github.com/sourcegraph/jetbrains/blob/main/gradle.properties#L27) (or uses one from cache, if exists).
This way it will also work without specifying `CODY_DIR`.

src/main/kotlin/com/sourcegraph/cody/agent/protocol_helpers/ClientInfo.kt
```diff
@@ -0,0 +1,25 @@
+package com.sourcegraph.cody.agent.protocol_helpers
+
+import com.sourcegraph.cody.agent.protocol_generated.ClientInfo as ProtocolClientInfo
+
+data class ClientInfo(
```
pkukielka: @RXminuS I think it is a bit confusing that helper classes have the same name as the generated one. I feel like a bit idiomatic for this use case would be to simply have a construction method in the `protocol_generated.ClientInfo` extension.

build.gradle.kts
pkukielka: This way when we will have a class with exactly the same name it will be clear it is needed because the one automatically generated is somehow incorrect (or, for the time being, not migrated yet).

src/main/kotlin/com/sourcegraph/cody/agent/protocol_helpers/ClientInfo.kt
RXminuS: Everything in protocl_generated is not allowed to be modified and also contains headers to warn against doing so. If we are missing helpers that we can make an issue to extend the generation system. 

 But I understand what you mean about the naming duplication being confusing. That should mostly go away though as we either use the new protocol classes directly. And for the other 10% of cases I think we're probably best going with a Factory design pattern instead of modifications to classes or wrapper classes.

build.gradle.kts
RXminuS: Good point! 👍

build.gradle.kts
```diff
@@ -341,12 +346,73 @@ tasks {
...
+    targetDir.deleteDirectoryContents()
+    sourceDir.copyRecursively(targetDir, overwrite = true)
+    // in each file replace the package name
+    for (file in targetDir.walkTopDown()) {
+      if (file.isFile && file.extension == "kt") {
+        val content = file.readText()
+        val newContent =
+            """
+        |/*
+        | * Generated file - DO NOT EDIT MANUALLY
```
olafurpg, line 385: 👍🏻

olafurpg: We can btw add a similar comment to the top of the file in the code generator. My thinking was that we can just do `cp -r ...` in a bash script and ideally avoid custom gradle tasks like this one

build.gradle.kts
```diff
@@ -341,12 +346,73 @@ tasks {
...
+        |
+    """
+                .trimMargin() +
+                // TODO: these are several "mods" until we can fix the code generation
+                content
+                    // TODO: CODY-2865 replacing ints with longs
+                    .replace(": Int", ": Long")
+                    // replacing classpath
+                    .replace(
+                        "com.sourcegraph.cody.protocol_generated",
```
olafurpg, line 398: Let's remove these workarounds, it only takes a couple minutes to incorporate these changes in the codegen https://github.com/sourcegraph/cody/pull/4915

src/main/kotlin/com/sourcegraph/cody/agent/EnumTypeAdapterFactory.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import com.google.gson.*
+import com.google.gson.annotations.SerializedName
+import com.google.gson.reflect.TypeToken
+import com.google.gson.stream.JsonReader
+import com.google.gson.stream.JsonWriter
+import java.io.IOException
+import java.lang.reflect.Field
+
+class EnumTypeAdapterFactory : TypeAdapterFactory {
```
olafurpg, line 11: Does this fix the issue in lsp4j where it writes numbers for string enums?

olafurpg (approved):
I'm very excited about this! I hit on a bug yesterday in Eclipse where deserializing `WebviewMessage` crashed because the bindings expect a string for a `URI` where a JSON serialized object appears. I'm hoping to not have to deal with these issues in `WebviewMessage` because we're only forwarding string-encoded messages for 99% of the features

src/main/kotlin/com/sourcegraph/cody/agent/EnumTypeAdapterFactory.kt
RXminuS: Yes

build.gradle.kts
RXminuS: Yes, I just wanted to have a single PR ready to ship the feature as it's time critical. But there's some time left so let's do so.

I might just want to add so we always generate an empty companion object too as it makes extending easier and I can get rid of the factory classes

olafurpg: The bindings have been unadopted for ~5 months, is it really time critical?

RXminuS: haha; No not in that sense. But my use of the auto bindings in a time critical feature just because I didn't want to dig a deeper hole meant I tried to be cautious about poking at adjacent stuff I didn't fully understand the impact of.

But you have showed me the way.  And properly fixing this in codegen is a lot easier than I thought 😊

(This is a stacked PR)

src/main/kotlin/com/sourcegraph/cody/agent/EnumTypeAdapterFactory.kt
RXminuS: We could probably just fix this in codegen too by generating serializers for all the enums

olafurpg: I'd prefer not to do that because we generate a lot of enums. I have a different workaround in lsp4j. If this adapter does the job then it's fine

RXminuS:
Superseded by https://github.com/sourcegraph/jetbrains/pull/1921

PR #1905, opened 2024-07-12
Do not show inlay hint when no account signed in

mkondratek:
Fixes https://linear.app/sourcegraph/issue/CODY-2840/confusion-with-cody-plugin-suggesting-key-combination-and-no-response

## Test plan
1. Having no account signed-in
2. Select some code

Expected: no inlay hint


pkukielka (approved):
LGTM

PR #1907, opened 2024-07-13
Stop spamming notifications when selecting code in Cody-ignored files

steveyegge:
fixes CODY-2863

I noticed this annoying behavior when working on debugging WSL -- all the files are considered "ignored", and selecting code in them went wild with notifications. So I've turned them off entirely for code selection. You will get a notification when you try an inline edit or other Cody action.

## Test plan

Tested locally by simulating the code paths. We don't have any Cody-ignore tests in JetBrains yet, unfortunately.


pkukielka (approved):
LGTM

PR #1911, opened 2024-07-15
Repeat mode for integration tests

odisseus:
To use it, add the option `-PrepeatTests` to the command line. Example:
```
./gradlew integrationTest -PrepeatTests=20
```

## Test plan

1. Introduce non-deterministic failures, like this:
   ```kotlin
   val x = Random.nextInt() % 10
   if (x < 2) {
       fail()
   } else if (x < 4) {
       1 / 0
   } else if (x < 6) {
       throw IllegalStateException("$x")
   }
   ```
3. Run the tests with high enough repetition count
4. Verify that different stack traces are present in the test report

Screenshot 2024-07-16 at 11 17 46

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -508,6 +508,7 @@ tasks {
     description = "Runs the integration tests."
     sharedIntegrationTestConfig(buildCodyDir, "replay")
     dependsOn("processIntegrationTestResources")
+    project.properties["repeatTests"]?.let { systemProperty("repeatTests", it) }
   }
 
   register<Test>("passthroughIntegrationTest") {
--- a/src/integrationTest/kotlin/com/sourcegraph/cody/edit/DocumentCodeTest.kt
+++ b/src/integrationTest/kotlin/com/sourcegraph/cody/edit/DocumentCodeTest.kt
@@ -18,9 +18,15 @@ import com.sourcegraph.cody.edit.widget.LensLabel
 import com.sourcegraph.cody.edit.widget.LensSpinner
 import com.sourcegraph.cody.util.CodyIntegrationTextFixture
 import junit.framework.TestCase
+import org.junit.Ignore
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
 
+@RunWith(JUnit4::class)
 class DocumentCodeTest : CodyIntegrationTextFixture() {
 
+  @Test
   fun testGetsFoldingRanges() {
     runAndWaitForNotifications(DocumentCodeAction.ID, TOPIC_FOLDING_RANGES)
 
@@ -39,6 +45,8 @@ class DocumentCodeTest : CodyIntegrationTextFixture() {
         selection.startOffset == caret && selection.endOffset == caret)
   }
 
+  @Ignore
+  @Test
   fun skip_testGetsWorkingGroupLens() {
     val assertsExecuted = AtomicInteger(0)
     val showWorkingGroupSessionStateListener =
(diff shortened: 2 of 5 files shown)


src/integrationTest/kotlin/com/sourcegraph/cody/edit/DocumentCodeTest.kt
```diff
@@ -139,6 +149,7 @@ class DocumentCodeTest : CodyIntegrationTextFixture() {
     assertNoActiveSession()
   }
 
+  @Test
   fun skip_testUndo() {
```
mkondratek: `skip_` was a prefix to force the skip (like ignore annotation). Please remove it from the method name.

src/integrationTest/kotlin/com/sourcegraph/cody/edit/DocumentCodeTest.kt
```diff
@@ -39,6 +45,8 @@ class DocumentCodeTest : CodyIntegrationTextFixture() {
         selection.startOffset == caret && selection.endOffset == caret)
   }
 
+  @Ignore
+  @Test
   fun skip_testGetsWorkingGroupLens() {
```
mkondratek: pls remove `skip_` prefix

build.gradle.kts
```diff
@@ -508,6 +508,7 @@ tasks {
     description = "Runs the integration tests."
     sharedIntegrationTestConfig(buildCodyDir, "replay")
     dependsOn("processIntegrationTestResources")
+    project.properties["repeatTests"]?.let { systemProperty("repeatTests", it) }
```
pkukielka, line 511: If you need it to be a system property you don't need to do that, but instead you can run gradle with `-DrepeatTests` instead of `-PrepeatTests`

src/integrationTest/kotlin/com/sourcegraph/cody/util/CustomJunitClassRunner.kt
```diff
@@ -0,0 +1,13 @@
+package com.sourcegraph.cody.util
+
+import org.junit.runners.BlockJUnit4ClassRunner
+import org.junit.runners.model.FrameworkMethod
+
+class CustomJunitClassRunner(klass: Class<*>?) : BlockJUnit4ClassRunner(klass) {
```
pkukielka, line 6: Maybe also something with `Repeatable` or `Repeat` in name? `Custom` doesn't bring too much context

pkukielka:
Could you somehow attach example off test report or at least screenshot of it?

build.gradle.kts
odisseus: According to [Stackoverflow](https://stackoverflow.com/a/21406600), the `-D` argument won't necessarily pass through to the tests.

src/integrationTest/kotlin/com/sourcegraph/cody/util/CustomJunitClassRunner.kt
odisseus: This class doesn't really repeat anything or make anything repeatable. I don't mind renaming it, but I can't think of any good name.

mkondratek (commented):
Is it easily doable to replace
```
DocumentCodeTest. testGetsFoldingRanges
DocumentCodeTest. testGetsFoldingRanges
DocumentCodeTest. testGetsFoldingRanges
...
```
with
```
DocumentCodeTest. testGetsFoldingRanges (0)
DocumentCodeTest. testGetsFoldingRanges (1)
DocumentCodeTest. testGetsFoldingRanges (2)
...
```
❓

odisseus:
Not without writing a lot more code. I will try to do that in a follow-up PR.

mkondratek (approved):

PR #1912, opened 2024-07-16
CI workflow to measure stability of integration tests

odisseus:
This is a follow-up to #1911.

Screenshot 2024-07-16 at 13 49 54

## Test plan

Tested on a personal fork

### DIFF
--- a/.github/workflows/integration-stability.yml
+++ b/.github/workflows/integration-stability.yml
@@ -1,0 +1,60 @@
+name: Measure stability of integration tests
+on:
+  workflow_dispatch:
+    inputs:
+      repetitions:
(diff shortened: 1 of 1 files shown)


mkondratek (approved):
How do we want to use repeated tests? Should we run them automatically for each PR? Or should we just use this workflow? When?

(you could describe the motivation and instruction in CONTRIBUTING.md or smth)

odisseus:
I'm not sure. I don't think we need to run the new workflow for every commit or pull request. It is worth running it manually on a pull request if we expect it to affect the stability of integration tests. Also, we should run it occasionally on main to get a baseline.

PR #1915, opened 2024-07-17
Do not fail the build when integration tests fail in repetition mode

odisseus:
This is a follow-up to #1912.

I believe that build failure is not particularly informative when it happens every single time. I would prefer having a way to distinguish at first glance between a build that finished normally (with the inevitable test failures) and a build that was aborted (e.g. due to an OOM error).

## Test plan

Verified on a personal fork

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -508,7 +508,12 @@ tasks {
     description = "Runs the integration tests."
     sharedIntegrationTestConfig(buildCodyDir, "replay")
     dependsOn("processIntegrationTestResources")
-    project.properties["repeatTests"]?.let { systemProperty("repeatTests", it) }
+    project.properties["repeatTests"]?.let {
+      systemProperty("repeatTests", it)
+      // We expect some of the repetitions to fail,
+      // so we don't treat this situation as a build failure
+      ignoreFailures = true
+    }
   }
 
   register<Test>("passthroughIntegrationTest") {


build.gradle.kts
```diff
@@ -508,7 +508,12 @@ tasks {
     description = "Runs the integration tests."
     sharedIntegrationTestConfig(buildCodyDir, "replay")
     dependsOn("processIntegrationTestResources")
-    project.properties["repeatTests"]?.let { systemProperty("repeatTests", it) }
+    project.properties["repeatTests"]?.let {
+      systemProperty("repeatTests", it)
+      // We expect some of the repetitions to fail,
```
pkukielka, line 513: Is there any practical benefit of that change?
I would say we **temporarily** expect them to fail because we know they are flaky now, but in general we want all them to succeed no matter how many times we rerun it.

odisseus: I expect this temporary expectation to stay with us for a long time (but if I'm wrong, we can easily revert this change).

In the meantime, the practical benefit looks like [this](https://github.com/odisseus/cody-jetbrains/actions/workflows/integration-stability.yml). Can you tell which [runs](https://github.com/sourcegraph/jetbrains/actions/workflows/integration-stability.yml) have completed normally, and which ones have crashed? (There's more than one of each type!)
Screenshot 2024-07-17 at 16 25 12

odisseus:
The memory issues for the integration tests have been resolved by #1922. Although there is still some flakiness and I expect at least one test failure on every run of the workflow, the crashes should be very rare now, so this PR is probably not needed.

PR #1919, opened 2024-07-18
Fix command-enter shortcut for Edit Code

mkondratek:
Fixes https://linear.app/sourcegraph/issue/CODY-2868.
Fixes https://github.com/sourcegraph/cody-issues/issues/307

## Test plan
1. In particular: Having IC 2024.1 and com.intellij.lang.jsgraphql (241.14494.150) plugin installed
2. Edit Code...
3. Trigger the edit with the shortcut

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -288,8 +288,7 @@
                 <action id="cody.inlineEditEditCode"
                         text="Edit Code"
                         class="com.sourcegraph.cody.edit.InlineEditPromptEditCodeAction" >
-                    <keyboard-shortcut first-keystroke="ctrl ENTER" keymap="$default"/>
-<!--                    replace-all="true"-->
+                    <keyboard-shortcut replace-all="true" first-keystroke="ctrl ENTER" keymap="$default"/>
                 </action>
 
                 <!-- also now dismisses error lens, if showing -->
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -52,6 +52,9 @@
 
         <applicationService
                 serviceImplementation="com.sourcegraph.cody.config.CodyPersistentAccounts"/>
+
+        <actionPromoter implementation="com.sourcegraph.cody.edit.CodyActionPromoter"/>
+
         <notificationGroup id="cody.auth" displayType="BALLOON"/>
         <notificationGroup id="Sourcegraph errors" displayType="BALLOON"/>
         <notificationGroup id="Sourcegraph: URL sharing" displayType="BALLOON"/>
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -285,6 +285,13 @@
                     <override-text place="GoToAction" text="Cody: Dismiss"/>
                 </action>
 
+                <action id="cody.inlineEditEditCode"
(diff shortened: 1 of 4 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -52,6 +52,9 @@
 
         <applicationService
                 serviceImplementation="com.sourcegraph.cody.config.CodyPersistentAccounts"/>
+
+        <actionPromoter implementation="com.sourcegraph.cody.edit.CodyActionPromoter"/>
```
mkondratek, line 56: this is the actual fix for this one - thank you JetBrains community Slack for help 😃

src/main/resources/META-INF/plugin.xml
```diff
@@ -285,6 +288,12 @@
                     <override-text place="GoToAction" text="Cody: Dismiss"/>
                 </action>
 
+                <action id="cody.inlineEditEditCode"
+                        text="Edit Code"
+                        class="com.sourcegraph.cody.edit.InlineEditPromptEditCodeAction" >
+                    <keyboard-shortcut replace-all="true" first-keystroke="ctrl ENTER" keymap="$default"/>
```
pkukielka, line 294: I see such comment from @steveyegge in this file:

```
    <!--
        N.B. You have to declare all the Mac keystrokes with ctrl or alt separately,
        even if their definition is identical to $default, because otherwise it will
        result in the actual keybindings switching to command/meta and/or shift.
    -->
```

@mkondratek  could you please check?

pkukielka: Also, please add `<override-text place="GoToAction" text="Cody: Confirm Code Edition"/>`
or something along those lines (it is for search everywhere).

src/main/resources/META-INF/plugin.xml
```diff
@@ -285,6 +288,12 @@
                     <override-text place="GoToAction" text="Cody: Dismiss"/>
                 </action>
 
+                <action id="cody.inlineEditEditCode"
+                        text="Edit Code"
```
pkukielka, line 292: I think text should not be `Edit Code`. We already have other `Edit Code` and I think both will be displayed in the key bindings settings with the same name (for button on the edit dialog we can have custom name, if needed).

mkondratek: We have `Edit Code...` action (not the three dots!) - it's a common practice in JB that the actions that open dialogs or need some confirmation have `...` suffix. IMHO it's a good enough to distinguish them (we clearly indicate what the shortcuts do and display them - it's super clear).

src/main/resources/META-INF/plugin.xml
mkondratek: `text="Cody: Confirm Code Edition"` that would be different from the text in the button and hence confusing imho

pkukielka: Problem is that we have two "Edit Code" now. One to open edit code prompt and second for confirming the edit. they somehow need to have different names.

pkukielka: I missed your second comment about `...`.
In this case: `<override-text place="GoToAction" text="Cody: Edit Code"/>`
Still we need to add `Cody: ` part.

mkondratek: done

pkukielka (approved):
LGTM

PR #1922, opened 2024-07-18
Shut down agent after each test

odisseus:
Since the `CodyIntegrationTextFixture` starts a fresh agent for every test case, we need to shut down those agents during cleanup. Otherwise the leftover agent processes clog the RAM (each Node process takes 50-250 MB), and the JVM threads which listen to them keep running forever.

## Test plan

Manually run the stability check in CI
[1000 repetitions](https://github.com/sourcegraph/jetbrains/actions/runs/10073107716) (5000 individual test cases): 2 hours 40 minutes, 118 failures

### DIFF
--- a/src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTextFixture.kt
+++ b/src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTextFixture.kt
@@ -49,6 +49,13 @@ open class CodyIntegrationTextFixture : BasePlatformTestCase() {
           logger.warn("Error shutting down session", x)
         }
       }
+      CodyAgentService.getInstance(myFixture.project).apply {
+        try {
+          stopAgent(project)
+        } catch (x: Exception) {
+          logger.warn("Error shutting down agent", x)
+        }
+      }
     } finally {
       super.tearDown()
     }


mkondratek (approved):
lgtm

src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTextFixture.kt
```diff
@@ -49,6 +49,13 @@ open class CodyIntegrationTextFixture : BasePlatformTestCase() {
           logger.warn("Error shutting down session", x)
         }
       }
+      CodyAgentService.getInstance(myFixture.project).apply {
+        try {
+          dispose()
```
pkukielka: This is not shutting down agent, but finishing current fixup session.
IMHO it makes sense to wait for such finish, but if it is not able to finish itself in a given time window it should be a test failure.
So instead of disposing it we should wait and check there.

src/integrationTest/kotlin/com/sourcegraph/cody/util/CodyIntegrationTextFixture.kt
```diff
@@ -259,7 +266,7 @@ open class CodyIntegrationTextFixture : BasePlatformTestCase() {
   }
 
   companion object {
-    const val ASYNC_WAIT_TIMEOUT_SECONDS = 10L
+    const val ASYNC_WAIT_TIMEOUT_SECONDS = 30L
```
pkukielka: Why is that needed? In the replay mode whole test should finish in well under 10s, not to mention single call.
Or maybe is it needed for the recording mode?
If so, maybe we should have configurable timeouts in the gradle config?

odisseus: Reverted

PR #1921, opened 2024-07-18
Code Actions Support (Ask Cody to Fix)

RXminuS:
Depends on #1931 

Add the ability for the Agent to provide CodeActions. In practice this means "Ask Cody to Fix" is now supported as a QuickFix on errors shown in JetBrains.

https://github.com/user-attachments/assets/f3bad3fd-cd67-427c-baa6-449a4c0dfd48

## FAQ

**Q: What errors are supported?**
A: Any errors shown in the JetBrains UI. Cody doesn't do any code analysis, we simply pass the issues that JetBrains is showing along. These Issues (or Inspections as they're known) can come both from JetBrain's language support, external Plugins as well as custom Inspections.

**Q: Can Cody explain the error like in VSCode?**
A: Not yet, but the foundational support is there now so it will be easy enough to add soon.

Fixes CODY-2788, CODY-2830

## Test plan
- Verified that existing unit & integration tests work with Protocol migrated protocol classes
- Manually verified that "Ask Cody to Fix" command is correctly shown and invoked. Most of the logic remains on the Agent.
- Isolated changes on FixupSession to FixupSessionV2 to not introduce bugs in FixupSessionV1 dependent codepaths which currently have low test coverage.

PR #1931, opened 2024-07-22
Generated Protocol Bindings

RXminuS:
Adds generated protocol bindings for use in #1921  

## Test plan

- Doesn't use generated protocol bindings. Verified that build still works


build.gradle.kts
```diff
@@ -341,6 +345,57 @@ tasks {
     return buildCodyDir
   }
 
+  fun copyProtocol(): Unit {
+    // This is just a temporary solution to make it easier to use certain generated protocol
+    // messages.
+    // Ultimately this should entirely be replaced by use of the generated protocol.
+    if (!isForceProtocolCopy) {
```
pkukielka, line 352: What is the purpose of this flag if it needs to always be set for task to have any effect?
Shouldn't force be necessary only if the protocol files already exists? (is it even an option that no such files would be present)?
If that flag is always needed then equally good we can just not have it.

pkukielka (approved):
LGTM except one other comment.

mkondratek (approved):
ok but pls note the Peter's comment

build.gradle.kts
RXminuS: I'll address this when we set up CI & automatic generation on builds etc. For now it's all manual while we iron out the wrinkles

PR #1932, opened 2024-07-22
Make UI hints disabled by default

pkukielka:
Fixes https://linear.app/sourcegraph/issue/CODY-3022/remove-the-edit-hints

## Test plan

1. Select multiline text in IntelliJ
2. Make sure edit hint is not displayed


danielmarquespt:
Will this only impact new installations, or all of our users (which is intended) ?

pkukielka:
> Will this only impact new installations, or all of our users (which is intended) ?

It will impact all users which upgrade the plugin.
The way those settings files works in IntelliJ is that those defaults from code are used, unless user specifically changed the settings. If that happens change is written down to the xml file (and after restart of the IDE loaded from it) - in this case `cody_application_settings.xml`.

mkondratek (approved):

PR #1938, opened 2024-07-23
Drop post-sign-up survey which breaks in Sourcegraph experimental UI

dominiccooney:
## Test plan

Manual test:

- Log in to Sourcegraph.com, flip the "experimental" mode on (top right).
- Open JetBrains settings and sign out of all accounts.
- In the side bar, use the social buttons to sign in. You should be able to sign in.
- Repeat the above test, but with the "experimental" mode off.


mkondratek (approved):

pkukielka (approved):
LGTM

mkondratek:
just verified - works

PR #1970, opened 2024-08-01
Got Cody working with WSL projects on Windows

### LINKED ISSUE #138: WSL PhpStorm [UriError]: If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//")
## Installation information
PhpStorm 2023.2.3
Build #PS-232.10072.32, built on October 13, 2023
Licensed to XXXXXXXXXXXX
Subscription is active until May 3, 2024.
XXXXXXXXXX
Runtime version: 17.0.8.1+7-b1000.32 amd64
Windows 11.0
Memory: 2048M
Cores: 16
Registry:
    debugger.watches.in.variables=false
    run.processes.with.pty=TRUE

Non-Bundled Plugins:
    com.github.shubham076.darkthemes (3.0.2)
    com.intellij.ideolog (222.1.0.0)
    CMD Support (1.0.5)
    Intellij_GitHub_Dark_Theme (1.1.1)
    net.vektah.codeglance (1.5.4)
    com.intellij.kubernetes (232.10072.33)
    PlantUML integration (7.3.0-IJ2023.2)
    com.alicannklc.laravel.snippets (1.1.1)
    ru.adelf.idea.dotenv (2023.2)
    org.endpointdeps.EndpointDependencies (1.3)
    de.espend.idea.laravel (0.15.4)
    net.seesharpsoft.intellij.plugins.csv (3.2.3-232)
    com.intellij.ml.llm (232.9826)
    com.nekofar.milad.intellij.nuxtjs (2.0.8)
    tv.twelvetone.intellij.plugins.intellivue (1.1.43)
    com.sourcegraph.jetbrains (5.2.7836-nightly)
    com.intellij.jira (4.6.3)
    mobi.hsz.idea.gitignore (4.5.2)

## (optional) Error message or idea.log output
WARN - #com.sourcegraph.cody.agent.CodyAgent - Error: Error: [UriError]: If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//")

## Steps to reproduce
Using phpstorm for windows and having the repository on WSL. IT seems to not recognize the path to the files. If i set the repository on windows it works as expected. 

## Expected behavior
Normal use

steveyegge:
fixes: CODY-2862

Cody couldn't start up if your repo/project was on a WSL drive/path, e.g., \\wsl.localhost\Ubuntu\home\person\foo -- we didn't handle the new URI syntaxes required here.

I've updated the code to normalize the paths properly. There was also a minor build issue that's fixed in `build.gradle.kts`.

## Test plan

I've tested it manually on Mac, Windows/normal, and Windows/WSL.

I wrote new unit tests for the primary change, which is how we normalize paths.

I will talk to our folks and see if we should add a manual QA pass for WSL projects.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -198,29 +198,32 @@ private constructor(
       return normalizeUriOrPath(uriString)
     }
 
-    // Fix up Windows paths.
+    @JvmStatic
     fun normalizeUriOrPath(uriString: String): String {
       val hasScheme = uriString.startsWith("file://")
       val path =
           (if (hasScheme) uriString.removePrefix("file://") else uriString).replace("\\", "/")
 
-      // Normalize WSL paths.
+      // Normalize WSL paths
       val wslPrefix = "//wsl$"
       if (path.startsWith(wslPrefix)) {
         val newPath = "//wsl.localhost${path.removePrefix(wslPrefix)}"
         return if (hasScheme) "file://$newPath" else newPath
       }
 
-      // Normalize drive letters for Windows.
+      // Normalize drive letters for Windows
       val driveLetterPattern = """^(\w):/""".toRegex()
       val normalizedPath =
           driveLetterPattern.replace(path) { matchResult ->
             val driveLetter = matchResult.groupValues[1].lowercase(Locale.getDefault())
-            // Ensure the drive letter path is formatted correctly
-            if (hasScheme) "file:///${driveLetter}:/" else "${driveLetter}:/"
+            "${driveLetter}:/"
           }
 
-      return if (hasScheme) "file:///$normalizedPath" else normalizedPath
+      if (!hasScheme) return normalizedPath
+      if (path.startsWith("/") && path.length > 1 && path[1] != '/') {
+        return "file://$normalizedPath"
+      }
+      return "file:///$normalizedPath"
     }
   }
 }
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
@@ -2,6 +2,7 @@ package com.sourcegraph.cody.agent.protocol
 
 import com.intellij.codeInsight.codeVision.ui.popup.layouter.bottom
 import com.intellij.codeInsight.codeVision.ui.popup.layouter.right
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.event.CaretEvent
 import com.intellij.openapi.editor.event.DocumentEvent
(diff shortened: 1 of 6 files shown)


steveyegge (commented):
I just noticed on the plane that this is wrong. It doesn't always normalize the slashes. Will fix on the plane in a bit.

Update August 6th:  I have inspected the code and it does not have the error I thought it had. It's ready to go.

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
```diff
@@ -186,12 +194,36 @@ private constructor(
...
-        val driveLetter =
-            it.groups[1]?.value?.lowercase(Locale.getDefault()) ?: return@replace it.value
-        "file:///$driveLetter%3A/"
+      val uriString = FileSystems.getDefault().getPath(file.path).toUri().toString()
+      return normalizeUriOrPath(uriString)
+    }
+
+    @JvmStatic
+    fun normalizeUriOrPath(uriString: String): String {
+      val hasScheme = uriString.startsWith("file://")
```
pkukielka, line 203: Do we care only about `file` schemes? If so, why?

pkukielka (commented):
LGTM except one question I added

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
pkukielka: I assume we probably do not support other protocols right now anyway, although I think we should support things like `jar://` which are common if one browses files from external libraries (and which still could be useful as open files for chat context).

src/test/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocumentTest.kt
```diff
@@ -223,4 +219,69 @@ class ProtocolTextDocumentTest : BasePlatformTestCase() {
...
+    assertEquals(expected, ProtocolTextDocument.normalizeUriOrPath(input))
+  }
+
+  fun testWindowsPathUpperCase() {
+    val input = "D:/home/person"
+    val expected = "d:/home/person"
+    assertEquals(expected, ProtocolTextDocument.normalizeUriOrPath(input))
+  }
+
+  fun testWslPathWithFileScheme() {
```
mkondratek, line 264: 💟

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
mkondratek: idk if that's directly related to the thing Peter is mentioning here but I could be worth to try to open a file like this:
```
/Users/mkondratek/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/2022.1/c91a19824ffcb28872ecd2761c83b250f703ff21/ideaIC-2022.1/lib/app.jar!/com/intellij/openapi/project/Project.class
```

(simply `runIde` with JetBrains project and go to the definition of some library class)

In particular `Explain Code` should work properly in it. 

I'd love to verify it myself but I have no access to my Windows PC till next week.

mkondratek (commented):
awesome 🚀 can't wait to have it released and notify the users about the happy news

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocument.kt
steveyegge: Verified that it works; screenshot attached. 
image

mkondratek (approved):

PR #1975, opened 2024-08-02
Fix integration tests

pkukielka:
### Changes

This PR fixes few problems:

* We were restarting agent without waiting for it to finish
* We were not properly refreshing external file changes to VFS which sometimes was leading to having old files in memory

It's also worth noting that `BasePlatformTestCase` seems to be in face reusing the same project for all tests, even if it does some minor cleanup on it. I think this is not ideal for our case, and instead we should try to either:

* reuse the same project and do not restart agent every time
* have full project isolation

Such in-the-middle solution as we have now does not seem to be beneficial to me. But we can try to change that later, especially in case this PR would not some all integration tests instabilities.

## Test plan

N/A

### DIFF
--- a/src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
+++ b/src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
@@ -1,55 +1,8 @@
 package com.sourcegraph.cody
 
-import com.sourcegraph.cody.agent.CodyAgentService
 import com.sourcegraph.cody.edit.DocumentCodeTest
-import com.sourcegraph.cody.util.CodyIntegrationTextFixture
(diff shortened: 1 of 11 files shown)


src/integrationTest/kotlin/com/sourcegraph/cody/AllSuites.kt
```diff
@@ -21,35 +17,4 @@ import org.junit.runners.Suite
  * Multiple recording files can be used, but each should have its own suite with tearDown() method
  * nad define unique CODY_RECORDING_NAME.
  */
```
mkondratek: The comment above seems to be inaccurate now

mkondratek (approved):

PR #1930, opened 2024-07-22
Add integration tests for the chat (context files)

mkondratek:
This is a follow up PR for https://github.com/sourcegraph/jetbrains/pull/1928. 

## TODO
- [ ] cody commit update 

## Test plan
1. Green CI

PR #1983, opened 2024-08-06
A few minor fixes

mkondratek:
## Test plan
1. #1930 passing (so we are sure the fix to the enhanced context works)

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
@@ -169,7 +169,7 @@ class CodyAgentService(private val project: Project) : Disposable {
               throw (CodyAgentException(msg))
             }
 
-        val agent = future.get(45, TimeUnit.SECONDS)
+        val agent = future.get(secondsTimeout, TimeUnit.SECONDS)
         if (!agent.isConnected()) {
           val msg = "Failed to connect to agent Cody agent"
           logger.error(msg)
--- a/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextItem.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextItem.kt
@@ -8,7 +8,8 @@ import com.google.gson.JsonObject
 import com.google.gson.JsonSerializer
 import com.intellij.openapi.vfs.StandardFileSystems
 import com.intellij.openapi.vfs.VfsUtil
-import com.sourcegraph.cody.agent.protocol_extensions.*
+import com.sourcegraph.cody.agent.protocol_extensions.intellijRange
+import com.sourcegraph.cody.agent.protocol_extensions.toSearchRange
 import com.sourcegraph.cody.agent.protocol_generated.Range
 import java.io.File
 import java.lang.reflect.Type
@@ -86,9 +87,8 @@ data class ContextItemFile(
           "@${uri.path.removePrefix(projectPath ?: "")}"
         } else {
           val repo = repoName?.split("/")?.lastOrNull()
-          val file = title
           return if (repo != null && title != null) {
-            "$repo $file"
+            "$repo $title"
           } else return uri.path
         }
 
--- a/src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
@@ -210,7 +210,7 @@ class CodyAuthenticationManager(val project: Project) :
       val previousUrl = previousAccount?.server?.url
       val previousTier = previousAccount?.isDotcomAccount()
 
-      this.account = newAccount
+      account = newAccount
       tier = null
       isTokenInvalid = null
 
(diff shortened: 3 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -169,7 +169,7 @@ class CodyAgentService(private val project: Project) : Disposable {
               throw (CodyAgentException(msg))
             }
 
-        val agent = future.get(45, TimeUnit.SECONDS)
+        val agent = future.get(secondsTimeout, TimeUnit.SECONDS)
```
mkondratek, line 172: this is my fault 😄

src/main/kotlin/com/sourcegraph/cody/agent/protocol/ContextItem.kt
```diff
@@ -86,9 +87,8 @@ data class ContextItemFile(
           "@${uri.path.removePrefix(projectPath ?: "")}"
         } else {
           val repo = repoName?.split("/")?.lastOrNull()
-          val file = title
           return if (repo != null && title != null) {
-            "$repo $file"
+            "$repo $title"
```
mkondratek, line 91: unrel improvement

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -210,7 +210,7 @@ class CodyAuthenticationManager(val project: Project) :
       val previousUrl = previousAccount?.server?.url
       val previousTier = previousAccount?.isDotcomAccount()
 
-      this.account = newAccount
+      account = newAccount
```
mkondratek, line 213: unrel

src/main/kotlin/com/sourcegraph/config/ConfigUtil.kt
```diff
@@ -140,7 +140,7 @@ object ConfigUtil {
           val jsonObj = JsonParser.parseString(text).asJsonObject
           jsonObj.keySet().associateWith { jsonObj[it] }
         } catch (e: Exception) {
-          logger.warn("Failed to read settings file", e)
+          logger.info("No user defined settings file found. Proceeding with empty custom config")
```
mkondratek, line 143: Let's limit spam in the logs

src/main/kotlin/com/sourcegraph/cody/context/RemoteRepoUtils.kt
```diff
@@ -27,7 +27,7 @@ object RemoteRepoUtils {
             Graphql_GetRepoIdsParams(codebaseNames.map { it.value }, codebaseNames.size.toLong())
         val repos = agent.server.graphql_getRepoIds(param).get()
         result.complete(
-            repos?.repos?.map { reposParams -> Repo(reposParams.id, reposParams.name) }
+            repos?.repos?.map { reposParams -> Repo(name = reposParams.name, id = reposParams.id) }
```
mkondratek, line 30: @RXminuS 👋

mkondratek: I was trying to get https://github.com/sourcegraph/jetbrains/pull/1930 working - happily, it turned out that my single test for chat is testing something and I was able to find this one 😅

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
RXminuS: It is our job to be creative chaos monkeys. Mistakes slipping through is either the fault of lacking DX or lacking process. If the system can't guard against oversights the system isn't done yet. I'll actually make a 😉 -krakathon-you-know-check for these kinds of mistakes.

src/main/kotlin/com/sourcegraph/cody/context/RemoteRepoUtils.kt
RXminuS: Not sure what you wanted me to check here 👋  @mkondratek ?

mkondratek: nothing, just pinging you as you are the author 😃

pkukielka: @RXminuS this is actually an issue caused by introducing generated protocol files.
Parameters order was switched in a old and new class, but since both types were `String` we never realised it's a bug.
But it breaks the logic.

That is something which may happen again in future and it's hard to guard agains unless we use named params everywhere. So just something to be cautious about :)

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/context/RemoteRepoUtils.kt
RXminuS: Gotcha.

To live by my previous ["statement of wisdom 🧌"](https://github.com/sourcegraph/jetbrains/pull/1983#discussion_r1705480898) I think what we can do for migration is make a quick check to print & ensure field orders are maintained. Just so we can flag wherever they break. ~I'll put up an issue for it~ [CODY-3150](https://linear.app/sourcegraph/issue/CODY-3150/)

PR #1996, opened 2024-08-07
Call `getAuthenticationState` on a pooled thread instead of EDT

mkondratek:
No need to call this method on EDT (at least not there).

## Test plan
1. `runIde`
2. verify that the error does not appear in the logs


pkukielka (approved):
LGTM

PR #1995, opened 2024-08-07
CodeActions: Skip UI specific paths

RXminuS:
Before the Cody CodeHighlight pass would apply to all editors, including those used in a "hacky" way to create UI elements such as the repo-input field used in the Cody sidebar.

<div class='graphite__hidden'>
          <div>🎥 Video uploaded on Graphite:</div>
            <a href="https://app.graphite.dev/media/video/onbAacT3LuCgMVcIIUAU/8364ca3b-a40d-4e24-87a8-149095bd0957.mp4">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/onbAacT3LuCgMVcIIUAU/8364ca3b-a40d-4e24-87a8-149095bd0957.mp4">CleanShot 2024-08-07 at 11.49.10.mp4</video>

Now we explicitly ignore these fake file editors. Note that as a follow up we might need to consider that other extensions similarly "mis-use" editors in the same way we did and that ignoring our specific paths isn't a long-term solution.

Fixes [CODY-3166](https://linear.app/sourcegraph/issue/CODY-3166/sourcegraphjetbrains1991-jetbrains-ask-cody-to-fix-and-ask-cody-to) / [#1991](https://github.com/sourcegraph/jetbrains/issues/1991)

## Test plan
Manually verified that HighlightPass no longer applies to the Repo Editor

Issue #1991, opened 2024-08-06
JetBrains: "Ask Cody to Fix" and "Ask Cody to Explain" appear for Add repository window. 


jay-fibi:
### IDE Information

IntelliJ IDEA 2023.3.3 (Community Edition)
Build #IC-233.14015.106, built on January 25, 2024
Runtime version: 17.0.9+7-b1087.11 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 14.3
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 8
Metal Rendering is ON
Registry:
ide.experimental.ui=true
Non-Bundled Plugins:
idea.plugin.protoeditor (233.13135.65)
DevKit (233.14015.152)
gherkin (233.11799.165)
cucumber-java (233.11799.196)
PythonCore (233.14015.106)
com.sourcegraph.jetbrains (6.0.22-nightly)
Kotlin: 233.14015.106-IJ

### Bug Description

Platform: macOS

**Steps to reproduce:**

1. Have Cody plugin installed in your JB IntelliJ editor
2. Login to Cody with Enterprise user
3. Click on `Add repositories` button
4. Enter some invalid URL
5. Click on red globe icon

### Observed behavior:
`Ask Cody to Fix` and `Ask Cody to Explain` appear in the list

### Expected behavior:
These options should not appear inside Add repository window.

### Additional context

**<img width="1470" alt="Screenshot 2024-08-07 at 12 50 07 AM" src="https://github.com/user-attachments/assets/a54028f0-a3f0-4696-b0b7-62e6d775d848">**

PR #1998, opened 2024-08-07
Disable highlight pass on mock files

pkukielka:
## Changes 

Replaces https://github.com/sourcegraph/jetbrains/pull/1995

Fixes Fixes [CODY-3166](https://linear.app/sourcegraph/issue/CODY-3166/sourcegraphjetbrains1991-jetbrains-ask-cody-to-fix-and-ask-cody-to) / https://github.com/sourcegraph/jetbrains/issues/1991

Mock files are used for in-memory buffers. Normally all editable code files are not mocks/in-memory files (even scratches are normal files) so we can assume if file is mock it is most likely some special component for which we do not want to add cody error highlighting.

## Test plan

Manually verified that HighlightPass no longer applies to the Repo Editor


RXminuS (approved):
I like it ... much better :-)

PR #2009, opened 2024-08-08
Update latest supported IntelliJ platform version to 242.2

pkukielka:
## Test plan

Full QA on 2024.2 IntelliJ version.


mkondratek:
I have verified the locally built plugin with that version - it works. lgtm

mkondratek (approved):

PR #2010, opened 2024-08-08
Run plugin verifier for 2024.2 and add INTERNAL_API_USAGES to the skipped failure levels

mkondratek:
Removing the internal api usages is nontrivial task. It seems to be better to run the verifier skipping the internal api usages complaints than not running it at all for 242. Let's run it and skip the internal api usages errors. Eventually, we should try to eliminate them and remove the skipped failure level.

The follow-up task: https://linear.app/sourcegraph/issue/CODY-3202/jetbrains-do-not-use-internal-api

## Test plan
1. Green CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,7 +80,7 @@ jobs:
           plugin-location: '*.zip'
           ide-versions: |
             ideaIC:2021.1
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
LGTM

PR #2023, opened 2024-08-09
Apply chore improvements; update latest ide version

mkondratek:
~We need the GitHub releases and their links to look nice. A stable release should not include `-nightly` in the link. This PR will automatically push a new tag without the suffix so we will be able to use it to generate the release notes.~

We cannot push from CI (without creating a token and setting it as a secret). Let's stick to the present process (manually push stable tag when needed).

Btw, I added a few minor improvements.

## Test plan
1. ~To test these changes we will need to do a stable release 😃~
1. N/A - these are only chore changes and improvements

### DIFF
--- a/.github/workflows/stable-release.yml
+++ b/.github/workflows/stable-release.yml
@@ -27,6 +27,7 @@ jobs:
       - run: |
           echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)" >> $GITHUB_ENV
       - run: echo "Publishing version $RELEASE_VERSION"
+      - run: git tag -a "RELEASE_VERSION" -m "RELEASE_VERSION" && git push origin "$RELEASE_VERSION"
       - run: ./gradlew "-PpluginVersion=$RELEASE_VERSION" publishPlugin
         env:
           PUBLISH_TOKEN: ${{ secrets.PUBLISH_TOKEN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/stable-release.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       - run: |
           echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)" >> $GITHUB_ENV
       - run: echo "Publishing version $RELEASE_VERSION"
+      - run: git tag -a "RELEASE_VERSION" -m "RELEASE_VERSION" && git push origin "$RELEASE_VERSION"
```
pkukielka: I don't think it will work on CI

mkondratek:
TODO: review this - https://stackoverflow.com/questions/51716044/how-do-i-push-to-a-repo-from-within-a-gitlab-ci-pipeline

pkukielka (changes requested):
I don't think it will work on CI

.github/workflows/stable-release.yml
mkondratek: you are right. let's stick to the current approach then

.github/workflows/stable-release.yml
```diff
@@ -26,7 +26,6 @@ jobs:
       - run: yarn global add pnpm@8.6.7
       - run: |
           echo "RELEASE_VERSION=$(./scripts/version-from-git-tag.sh)" >> $GITHUB_ENV
-      - run: echo "Publishing version $RELEASE_VERSION"
```
pkukielka, line 29: Why that change? :)

pkukielka (approved):
LGTM

.github/workflows/stable-release.yml
mkondratek: this step was redundant

PR #2061, opened 2024-08-19
Fix copy text in webviews

mkondratek:
Cody part: https://github.com/sourcegraph/cody/pull/5236

## Test plan
1. Ask chat to generate a piece of code
2. Click "Copy Code" on the generated code block
3. Verify that the code has been copied (try to paste it)

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
@@ -1,6 +1,7 @@
 package com.sourcegraph.cody.ui
 
 import com.google.gson.Gson
+import com.google.gson.GsonBuilder
 import com.google.gson.JsonArray
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.runInEdt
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
```diff
@@ -66,12 +86,40 @@ interface NativeWebviewProvider {
...
+    private val logger = Logger.getInstance(WebUIServiceWebviewProvider::class.java)
+  }
+
   override fun createPanel(params: WebviewCreateWebviewPanelParams) =
       WebUIService.getInstance(project).createWebviewPanel(params)
 
-  override fun receivedPostMessage(params: WebviewPostMessageStringEncodedParams) =
-      WebUIService.getInstance(project)
-          .postMessageHostToWebview(params.id, params.stringEncodedMessage)
+  override fun receivedPostMessage(params: WebviewPostMessageStringEncodedParams) {
```
mkondratek, line 96: Can we send a pain message (not string encoded)? If not, what is the proper approach? 

Maybe there is a way to handle copy-to-clipboard wholly in Code? Can we alter `navigator.clipboard.writeText(preText)` from [vscode/webviews/chat/ChatMessageContent/create-buttons.ts](https://sourcegraph.com/github.com/sourcegraph/cody@mkondratek/fix/copy-code/-/blob/vscode/webviews/chat/Chat MessageContent/create-buttons.ts?L173%3A1-222%3A1) to do so?

pkukielka:
Replaced by https://github.com/sourcegraph/jetbrains/pull/2063

PR #2063, opened 2024-08-19
Properly handle copy code event

pkukielka:
Fixes CODY-3064

## Changes 

Add explicit handling for a copy event + change manual JSON parsing to reading from JSON object.

## Test plan

1. Ask chat to generate a piece of code
2. Click "Copy Code" on the generated code block
3. Verify that the code has been copied (try to paste it)

Check for both IJ 2022 and IJ 2024.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
@@ -2,11 +2,13 @@ package com.sourcegraph.cody.ui
 
 import com.google.gson.Gson
 import com.google.gson.JsonArray
+import com.google.gson.JsonParser
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.runInEdt
 import com.intellij.openapi.components.Service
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
```diff
@@ -415,19 +435,26 @@ class WebUIProxy(private val host: WebUIHost, private val browser: JBCefBrowserB
...
+    val queryObject = JsonParser.parseString(query).asJsonObject
+    val queryWhat = queryObject["what"]?.asString
+
+    when (queryWhat) {
+      "postMessage" -> {
+        val queryValue = queryObject["value"] ?: return
+        host.postMessageWebviewToHost(queryValue.toString())
+
+        val messageObject = if (queryValue.isJsonObject) queryValue.asJsonObject else return
+        if (messageObject["command"]?.asString == "copy" &&
```
mkondratek, line 447: works for now but eventually - if we want to support more different messages - it would be nice to have the handlers in separate methods and keep it cleaner

mkondratek (approved):
lgtm

PR #2065, opened 2024-08-19
Avoid white flash when creating new panel by enabling offscreen prerendering

pkukielka:
Fixes CODY-2844

## Changes

PR title really says it all

## Test plan

1. Start a new chat, ask a question
2. Open new chat window

There should be no flash of white while new chat is loading.


mkondratek (approved):
Nice 😄

PR #2072, opened 2024-08-20
.gitignore `protocol_generated` and `copyProtocol` before `kotlinCompile` 

mkondratek:
The former approach allowed us to have a protocol that does not correspond to the commit pointed by `cody.commit`. Usually, we had an outdated protocol. `cody.commit` corresponds to the specific protocol and it seems to be a good idea to keep the protocol strictly up to date. 

This PR removes the generated protocol from git and runs `copyProtocol` just before `kotlinCompile` task. It guarantees us the consistency. It allows us to notice some important changes earlier (in test CI).

## Test plan
1. Green CI


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -129,6 +129,7 @@ private constructor(
                               webview = ClientCapabilities.WebviewEnum.Native,
                               webviewNativeConfig =
                                   WebviewNativeConfigParams(
+                                      view = WebviewNativeConfigParams.ViewEnum.Multiple,
```
mkondratek, line 132: speaking of the devil... our protocol was outdated. 

From now on, we will be able to notice similar scenarios in a PR bumping the cody commit.

pkukielka (approved):
LGTM

PR #2074, opened 2024-08-20
Fix input prompt not working after switching to offscreen rendering

pkukielka:
Fixes https://linear.app/sourcegraph/issue/CODY-3369/backspace-key-not-functioning

## Test plan

Scenario 1:

1. Open chat in IntelliJ older than 2024.
2. Make sure backspace works in chat input prompt.

Scenario 2:

1. Open IJ 2024. 
2. Make sure there is no flickering when switching to new chat window

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
@@ -3,6 +3,7 @@ package com.sourcegraph.cody.ui
 import com.google.gson.Gson
 import com.google.gson.JsonArray
 import com.google.gson.JsonParser
+import com.intellij.openapi.application.ApplicationInfo
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.runInEdt
 import com.intellij.openapi.components.Service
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
```diff
@@ -334,7 +335,9 @@ class WebUIProxy(private val host: WebUIHost, private val browser: JBCefBrowserB
       val browser =
           JBCefBrowserBuilder()
               .apply {
-                setOffScreenRendering(true)
+                val isOffScreenRenderingSupported =
+                    ApplicationInfo.getInstance().build.baselineVersion >= 224
```
mkondratek, line 339: did you verify the versions in between? does it change exactly in 224?

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
pkukielka: Looks like so, 2024.1 works, and e.g. 2023.3 does not.

PR #2078, opened 2024-08-20
Fix CodeSearch download

mkondratek:
## Test plan
- green ci


pkukielka (approved):
LGTM, thanks for fixing it!

PR #2083, opened 2024-08-21
Intercept the "edit code" command from WebView and run the native JetBrains UI

dominiccooney:
Out of the box, the WebView runs the Cody "edit" command which tries to open a QuickPick. VSCode shims does not support QuickPick, and has an independent implementation of the edits UI that is not integrated at the extension TypeScript level.

For now, intercept the WebView's request to run "edit" and invoke the native JetBrains edit command directly.

Fixes CODY-3238.

## Test plan

Manually tested:

1. Open a file.
2. Open the sidebar, commands tab, click Edit Code.
3. Verify that the Cody Edit dialog appears.

Note, the OK button doesn't work because of an unrelated regression in JetBrains edits in general, probably introduced in https://github.com/sourcegraph/cody/commit/698239e099a153d54364960e89e6e5402843e820 . When that is fixed, this will work end-to-end.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
@@ -271,18 +271,26 @@ class WebUIHostImpl(
   override fun postMessageWebviewToHost(stringEncodedJsonMessage: String) {
     // Some commands can be handled by the client and do not need to round-trip client -> Agent ->
     // client.
-    when (stringEncodedJsonMessage) {
-      // TODO: Remove this redirection when sign in moves to Web UI.
-      "{\"command\":\"command\",\"id\":\"cody.auth.signin\"}",
-      "{\"command\":\"command\",\"id\":\"cody.auth.signout\"}" -> {
+    val stringsOfInterest = listOf("cody.auth.signin", "cody.auth.signout", "cody.action.command")
+    val decodedJson =
+        if (stringsOfInterest.any { stringEncodedJsonMessage.contains(it) }) {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
```diff
@@ -275,6 +280,27 @@ class WebUIHostImpl(
               .showSettingsDialog(project, AccountConfigurable::class.java)
         }
       }
+      // TODO: Delete this intercept when Cody edits UI is abstracted so JetBrains' native UI can be invoked from the
+      // extension TypeScript side through Agent.
+      "{\"command\":\"command\",\"id\":\"cody.action.command\",\"arg\":\"edit\"}" -> {
```
pkukielka: This looks a bit fragile, as all parsing jsons by hand.
Are there any downsides to parse it like I did in `handleCefQuery`?

pkukielka (approved):
Giving you an approve if you want to merge, but I would consider proper JSON parsing - it won't be any harder but is potentially safer.

Also, you need to reformat with spotless ;)

src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
dominiccooney: The serialization comes from JSON.stringify so this is not particularly fragile, and this is succinct.

dominiccooney:
- Run spotless. Spotless comment line breaking is odd.
- Switch to JSON parsing, did not bother to check whether any new keys have appeared because this is already getting verbose.

PR #2084, opened 2024-08-21
Fix missing model ID for Edit commands

valerybugakov:
- Migrated the `ChatModelsResponse` class to use the generated `Model` class internally as a follow-up to https://github.com/sourcegraph/cody/pull/5267 to fix the missing model ID for Edit commands.
- Slack thread https://sourcegraph.slack.com/archives/C06R69BC8UW/p1724228969152979
- Part of https://linear.app/sourcegraph/issue/CODY-3409/self-hosted-models-openaicompatible-autocomplete-provider-supports

## Test plan

CI

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/history/state/LLMState.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/history/state/LLMState.kt
@@ -7,7 +7,7 @@ import com.sourcegraph.cody.agent.protocol.ChatModelsResponse
 
 @Tag("llm")
 class LLMState : BaseState() {
-  @get:OptionTag(tag = "model", nameAttribute = "") var model: String? by string()
+  @get:OptionTag(tag = "id", nameAttribute = "") var id: String? by string()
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/history/state/LLMState.kt
```diff
@@ -7,7 +7,7 @@ import com.sourcegraph.cody.agent.protocol.ChatModelsResponse
 
 @Tag("llm")
 class LLMState : BaseState() {
-  @get:OptionTag(tag = "model", nameAttribute = "") var model: String? by string()
```
mkondratek: are we going to keep intellij's persistance? or is it to be handled by the agent? if we want to keep it we would need a migration to migrate the old state

valerybugakov: Oh, I see. Then, I'd keep this part unchanged. Let me push an update.

PR #2086, opened 2024-08-21
Fix missing llm in Inline Edit

mkondratek:
This PR is a reattempt of https://github.com/sourcegraph/jetbrains/pull/2084. 

API changed in Cody. `model` becomes `id`. 

## Test plan
1. Edit Code works


src/main/kotlin/com/sourcegraph/cody/history/HistoryService.kt
```diff
@@ -151,5 +153,21 @@ class HistoryService(private val project: Project) :
 
   companion object {
     @JvmStatic fun getInstance(project: Project): HistoryService = project.service<HistoryService>()
+
+    fun convertModelsToChatModelProviders(models: List<Model>) =
```
pkukielka: `Model` and `ChatModelProvider` are in fact the same data structure, one autogenerated and one crafted by us, no?
Why not to replace it completely?

mkondratek: done, ~however we need to leave the class for the sake of some old migrations that could have the separate deprecated and isCodyProOnly fields in their state~ 

nah, we can remove it - we did not persist old fashioned field, we have the tags only

pkukielka (approved):
Thank you! LGTM now

pkukielka:
```
> Task :test
CompileCommand: exclude com/intellij/openapi/vfs/impl/FilePartNodeRoot.trieDescend bool exclude = true

SettingsMigrationTest > test ChatTagsLlmMigration FAILED
    junit.framework.AssertionFailedError at SettingsMigrationTest.kt:351
```

PR #2087, opened 2024-08-21
Fix focus and panel movement issues

pkukielka:
Fixes CODY-2842
Fixes CODY-3367
Fixes CODY-3068

# Changes

1. [JetBrains JCEF has faulty focus implementation ](https://youtrack.jetbrains.com/issue/IJPL-158952/Focus-issue-when-using-multiple-JCEF-instances) sometime causing infinite cycle of components stealing focus from each other. I have not found a way to fix this elegantly, but dirty method works:
i. remove existing focus handler using reflection (:()
ii. add new focus handler with circuit breaker

2. Our implementation of panel disposal stopped working, as apparently if panel is moved old editor can be removed before new on is created (which breaks assumption in the old code). Again, I found no way to implement this in a nice way (like existing listener on "move" file between editors) and instead I had to implement a delayed, scheduled disposal if needed.

## Test plan

**Scenario for bug #1:** 

1. Open few chats in new standalone panels
2. Make sure last one have a focus
3. Close this tab

Expected result: There should be no flickering between right Cody panel and last chat open as standalone panel.

**Scenarios for bug #2:** 

_Scenario 1_
1. Open chat in new standalone panel
2. Click on the tab name and select `Split and Move Right`

Expected result: It should not crash. Window should be split and functioning panel should be moved to the right.

_Scenario 2_
1. Open few chats in new standalone panels mixed with normal files
2. Drag and drop tabs changing their order
3. Click on one of the chat panel tabs and select "Split Right"
4. Click on one of the chat panel tabs and select "Split Down"
5. Drag and drop chat tabs between split views

Expected result: It should not crash. All chat split & move operations should work as one would normally expect from other files.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
@@ -34,6 +34,7 @@ import com.sourcegraph.cody.config.ui.AccountConfigurable
 import com.sourcegraph.cody.sidebar.WebTheme
 import com.sourcegraph.cody.sidebar.WebThemeController
 import com.sourcegraph.common.BrowserOpener
+import java.awt.Component
 import java.awt.datatransfer.StringSelection
 import java.io.IOException
 import java.net.URI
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
Looks good.

I saw some cycle busting solution in IJ platform forums, linked from the Linear issue. But if the Reflection solution is better, that's fine.

dominiccooney:
The test plan would be better if it included dragging and dropping *and* splitting *and* split and move, not just split and move.

We could consider splitting up the focus and dispose changes.

pkukielka:
> Looks good.
> 
> I saw some cycle busting solution in IJ platform forums, linked from the Linear issue. But if the Reflection solution is better, that's fine.

I think the solution proposed in the thread you linked was applied in the same place (`onSetFocus`) but there are two differences:
* reflection is needed because JetBrains implementation (`JBCefBrowser`) adds this method already, but incorrectly; so we need to remove it, and we need reflection to be able to do that
* they were not switching focus if any other browser window was already in focus... which I do no think if fully correct

Anyway we can always tinker with the exact behaviour in future if we will find any corner cases where it is not perfect.

PR #2089, opened 2024-08-21
Increase minimum supported IntelliJ version to 2023.2

pkukielka:
Fixes CODY-3383

This also partially address CODY-3193.
We still need to switch to platform version 2.0 but now it will not be technically blocked.

## Test plan

Try to install plugin in IntelliJ `2023.1` or lower - it should not be allowed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,7 +79,7 @@ jobs:
         with:
           plugin-location: '*.zip'
           ide-versions: |
-            ideaIC:2022.1
+            ideaIC:2023.2
             ideaIC:2024.2
           failure-levels: |
             COMPATIBILITY_WARNINGS
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,7 +6,7 @@ pluginName=Sourcegraph
 pluginVersion=6.0-localbuild
 # See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 # for insight into build numbers and IntelliJ Platform versions.
-pluginSinceBuild=232.*
+pluginSinceBuild=232
 pluginUntilBuild=242.*
 # IntelliJ Platform Properties -> https://github.com/JetBrains/gradle-intellij-plugin#intellij-platform-properties
 platformType=IC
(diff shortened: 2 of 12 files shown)


dominiccooney (approved):
LGTM

mkondratek (approved):
wow 🥹

.github/workflows/test.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         with:
           plugin-location: '*.zip'
           ide-versions: |
-            ideaIC:2022.1
```
mkondratek, line 82: there is also a reference to 2022 in README.md

gradle.properties
```diff
@@ -6,16 +6,16 @@ pluginName=Sourcegraph
 pluginVersion=6.0-localbuild
 # See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 # for insight into build numbers and IntelliJ Platform versions.
-pluginSinceBuild=221.1
```
mkondratek, line 9: If I understand correctly, we will be able to move to the new intellj platform plugin and resolve the FailureLevel.COMPATIBILITY_PROBLEMS issues (remove this failure level), right?

https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html#requirements

gradle.properties
```diff
@@ -6,16 +6,16 @@ pluginName=Sourcegraph
...
+pluginSinceBuild=232.*
 pluginUntilBuild=242.*
 # IntelliJ Platform Properties -> https://github.com/JetBrains/gradle-intellij-plugin#intellij-platform-properties
 platformType=IC
-platformVersion=2022.1
+platformVersion=2023.2
 # Plugin Dependencies -> https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html
 # Example: platformPlugins = com.intellij.java, com.jetbrains.php:203.4449.22
 platformPlugins=Git4Idea,PerforceDirectPlugin,java
 # Java language level used to compile sources and to generate the files for - Java 11 is required for 2020.3 <= x < 2022.2
```
mkondratek: I guess the comment is no longer needed

mkondratek:
works

gradle.properties
pkukielka: Yep

PR #2095, opened 2024-08-22
Do not run edit OK action if edit dialog is not visible

dominiccooney:
Fixes CODY-3264

## Test plan

Manually tested:

1. Right click on a file, edit, perform an edit such as "replace strings with a different human language"
2. Start a chat, enter a prompt such as "what does this file do?" and hit cmd-enter
3. Confirm this does not trigger a second edit


pkukielka (approved):
LGTM

PR #2108, opened 2024-08-23
added(agent): `chat/import` to migrate historical chats to agent storage for webview

jamesmcnamara:
Adds a migration to import local chat history into the agent to be used in the webview.

Tie in for https://github.com/sourcegraph/cody/pull/5304

## Test plan
Tested manually. Need to add a unit test.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
@@ -54,17 +54,17 @@ class SettingsMigration : Activity {
       organiseChatsByAccount(project)
     }
     RunOnceUtil.runOnceForApp("CodyApplicationSettingsMigration") { migrateApplicationSettings() }
-    RunOnceUtil.runOnceForApp("ToggleCodyToolWindowAfterMigration") {
+    RunOnceUtil.runOnceForProject(project, "ToggleCodyToolWindowAfterMigration") {
       toggleCodyToolbarWindow(project)
     }
-    RunOnceUtil.runOnceForApp("CodyAccountsIdsRefresh") { refreshAccountsIds(project) }
-    RunOnceUtil.runOnceForApp("CodyAssignOrphanedChatsToActiveAccount") {
+    RunOnceUtil.runOnceForProject(project, "CodyAccountsIdsRefresh") { refreshAccountsIds(project) }
+    RunOnceUtil.runOnceForProject(project, "CodyAssignOrphanedChatsToActiveAccount") {
       migrateOrphanedChatsToActiveAccount(project)
     }
 
     DeprecatedChatLlmMigration.migrate(project)
     ChatTagsLlmMigration.migrate(project)
-    RunOnceUtil.runOnceForApp("CodyMigrateChatHistory") {
+    RunOnceUtil.runOnceForProject(project, "CodyMigrateChatHistory") {
       ChatHistoryMigration.migrate(project)
     }
   }
--- a/src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
@@ -64,6 +64,9 @@ class SettingsMigration : Activity {
 
     DeprecatedChatLlmMigration.migrate(project)
     ChatTagsLlmMigration.migrate(project)
+    RunOnceUtil.runOnceForApp("CodyMigrateChatHistory") {
+      ChatHistoryMigration.migrate(project)
+    }
   }
 
   private fun migrateOrphanedChatsToActiveAccount(project: Project) {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
```diff
@@ -64,6 +64,9 @@ class SettingsMigration : Activity {
 
     DeprecatedChatLlmMigration.migrate(project)
     ChatTagsLlmMigration.migrate(project)
+    RunOnceUtil.runOnceForApp("CodyMigrateChatHistory") {
```
pkukielka: I think this should be `runOnceForProject`.
I think few other using `runOnceForApp` are incorrect too, could you maybe fix them by the way?
The only one which should use `runOnceForApp` is - I believe - `CodyApplicationSettingsMigration` as you can see it's not using `project` at all.

jamesmcnamara: Ah nice. Updated.

mkondratek:
It works now 🚀

src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
```diff
@@ -54,16 +54,19 @@ class SettingsMigration : Activity {
...
     }
     RunOnceUtil.runOnceForApp("CodyApplicationSettingsMigration") { migrateApplicationSettings() }
-    RunOnceUtil.runOnceForApp("ToggleCodyToolWindowAfterMigration") {
+    RunOnceUtil.runOnceForProject(project, "ToggleCodyToolWindowAfterMigration") {
       toggleCodyToolbarWindow(project)
     }
-    RunOnceUtil.runOnceForApp("CodyAccountsIdsRefresh") { refreshAccountsIds(project) }
-    RunOnceUtil.runOnceForApp("CodyAssignOrphanedChatsToActiveAccount") {
+    RunOnceUtil.runOnceForProject(project, "CodyAccountsIdsRefresh") { refreshAccountsIds(project) }
+    RunOnceUtil.runOnceForProject(project, "CodyAssignOrphanedChatsToActiveAccount") {
```
mkondratek, line 61: I see the change from `runOnceForApp` to `runOnceForProject` (and that `CodyMigrateChatHistory` is for project too). 

AFAIK Cody state is persisted between projects. In particular, the logged in account remains unchanged when switching project while still using the same IDE (e.g. IntelliJ Ultimate). 

On the other hand `CodyAuthenticationManager` is a project level service. This is somewhat inconsistent. 

Anyway, I think this change is harmless. Eventually, all of these should work.

src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
```diff
@@ -54,16 +54,19 @@ class SettingsMigration : Activity {
...
       migrateOrphanedChatsToActiveAccount(project)
     }
 
     DeprecatedChatLlmMigration.migrate(project)
     ChatTagsLlmMigration.migrate(project)
+    RunOnceUtil.runOnceForProject(project, "CodyMigrateChatHistory") {
```
mkondratek, line 67: What if 

1. I have an account with a chat, e.g.:
```
<accountId value="VXNlcjoxNDIxNTQ=" />
<chats>
  <list>
    <chat>
      <internalId value="ac348d02-8fbc-41d3-a7c0-d73a571fc719" />
      <llm>
        <llm>
          <model value="openai/gpt-4-turbo" />
          <provider value="OpenAI" />
          <title value="GPT-4 Turbo" />
        </llm>
      </llm>
      <messages>
        <list>
          <message>
            <speaker value="HUMAN" />
            <text value="What model are you?" />
          </message>
          <message>
            <speaker value="ASSISTANT" />
            <text value="I am GPT-4 Turbo created by OpenAI."/>
          </message>                   
  </list>
</chats>
```
2. I open a project JetBrains with my IntelliJ Ultimate
3. The chat is migrated to the agent storage
4. I restore the chat and write a follow up question, wait for the response 
5. I open another project e.g. cody with the same IntelliJ Ultimate (the same app but different project)

What will happen? Will the chat be overrode by the second migration? Will my follow up question be "removed"? Maybe it should be `runOnceForApp`?

src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
pkukielka: FYI conversations are also saved per project.

src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
pkukielka: History is per project.
`HistoryService.getInstance(project).getChatHistoryFor(account.id)`
Doing migration per app would be incorrect, and would cause only one project (the first opened) to be migrated.

pkukielka:
LGTM.
@mkondratek please revert your last commit and we can merge this first:
https://github.com/sourcegraph/jetbrains/pull/2131

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
jamesmcnamara: Also to your question, the use of `_.merge` means data won't be lost. The resulting history will have both histories.

PR #2128, opened 2024-08-27
Add action to toggle on/off offscreen rendering

pkukielka:
## Test plan

1. Hit `shift shift`
2. Start typing "offscreen", you should see "Cody: Enable offscreen rendering"
3. Hit space to turn it on / off

image

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
@@ -6,7 +6,6 @@ import com.google.gson.JsonParser
 import com.intellij.openapi.actionSystem.ActionManager
 import com.intellij.openapi.actionSystem.AnActionEvent
 import com.intellij.openapi.actionSystem.CommonDataKeys
-import com.intellij.openapi.application.ApplicationInfo
 import com.intellij.openapi.application.ApplicationManager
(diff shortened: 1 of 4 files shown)


dominiccooney (approved):
Looks good, should we put `(Requires restart)` in the label or something?

pkukielka:
> Looks good, should we put `(Requires restart)` in the label or something?

It turned out to not be required, if you spawn new panel it already respects new setting.

src/main/kotlin/com/sourcegraph/cody/ui/CodyToolWindowFactory.kt
```diff
@@ -419,9 +419,7 @@ class WebUIProxy(private val host: WebUIHost, private val browser: JBCefBrowserB
       val browser =
           JBCefBrowserBuilder()
               .apply {
-                val isOffScreenRenderingSupported =
-                    ApplicationInfo.getInstance().build.baselineVersion >= 232
```
mkondratek, line 423: What's the background of this change? Isn't it the matter of the app version anymore?

pkukielka: This branch works only on `>= 232` now, so this check is redundant.
Background: we want to be able to test various UI behaviours with and without this flag, and be able to switch on the fly easily.

PR #2140, opened 2024-08-28
Fix triggering commands from webview

pkukielka:
Fixes CODY-3531
Fixes CODY-3533

## Test plan

1. Trigger Document command from webview ui `Prompts & Commands` tab
2. It should work :)


mkondratek (approved):
robi sens

PR #2144, opened 2024-08-28
Enable JCEF in integration tests

mkondratek:
## Test plan
1. IntelliJ does not complain about the env variable being "false"

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -302,8 +302,7 @@ fun Test.sharedIntegrationTestConfig(buildCodyDir: File, mode: String) {
       "CODY_TEMPERATURE_ZERO" to "true",
       "CODY_TELEMETRY_EXPORTER" to "testing",
       // Fastpass has custom bearer tokens that are difficult to record with Polly
-      "CODY_DISABLE_FASTPATH" to "true",
-      "ide.browser.jcef.headless.enabled" to "true")
+      "CODY_DISABLE_FASTPATH" to "true")
 
   useJUnit()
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -226,7 +226,7 @@ fun Test.sharedIntegrationTestConfig(buildCodyDir: File, mode: String) {
       "CODY_TELEMETRY_EXPORTER" to "testing",
       // Fastpass has custom bearer tokens that are difficult to record with Polly
       "CODY_DISABLE_FASTPATH" to "true",
-  )
+      "ide.browser.jcef.headless.enabled" to "true")
```
mkondratek: it still complains... 🤔 

```
[   6779]   WARN - #c.i.u.j.JBCefApp - JCEF is manually disabled in headless env via 'ide.browser.jcef.headless.enabled=false'
```

dominiccooney: I don't think these are environment variables, but Registry properties?

`com. intellij. openapi. util. registry.Registry.get("ide.browser.jcef.headless.enabled").setValue("true")` something like that...

dominiccooney: Ping @mkondratek

mkondratek: Thanks Dominic. I tried setting this in the test setup. Something goes wrong in the test execution in CI. I have no time to focus on it now. Hopefully, I will get back to it later this week.

mkondratek:
won't fix. we are going to switch to the other integration tests anyway

PR #2148, opened 2024-08-28
Make offscreen rendering a default

pkukielka:
## Test plan

1. Run IDE
2. Check if problems mentioned there: https://linear.app/sourcegraph/issue/CODY-3545/decide-if-we-should-go-with-offscreenrendering-enabled-or-disabled in the section **OffScreenRendering disabled** are gone


danielmarquespt (approved):
Approved

PR #2154, opened 2024-08-29
Make getIdeaInstallDir more generic to support other platform types

mkondratek:
## Test plan
1. runIde with pycharm and with idea

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -409,12 +409,23 @@ tasks {
           "cody.autocomplete.enableFormatting" to
               (project.property("cody.autocomplete.enableFormatting") ?: "true"))
 
-  fun getIdeaInstallDir(ideaVersion: String, ideaType: String): File? {
+  fun getPlatformName(ideaType: String) =
+      when (ideaType) {
+        "PC" -> "pycharm"
+        "IC",
+        "IU" -> "idea"
+        else -> throw IllegalArgumentException("Unsupported ideaType: $ideaType")
+      }
+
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -409,12 +409,23 @@ tasks {
           "cody.autocomplete.enableFormatting" to
               (project.property("cody.autocomplete.enableFormatting") ?: "true"))
 
-  fun getIdeaInstallDir(ideaVersion: String, ideaType: String): File? {
+  fun getPlatformName(ideaType: String) =
```
pkukielka, line 412: https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/main/src/main/kotlin/org/jetbrains/intellij/platform/gradle/IntelliJPlatformType.kt

mkondratek: Not yet available with 1.17.4, we must move to the new gradle plugin (v2).

pkukielka: Maybe just after launch it will be good time to do that anyway?

dominiccooney (approved):
Being able to test more IDEs sounds great.

- Please update the CONTRIBUTING.md documenting how to use this.
- Why did you stop at just pycharm? Could you add Goland, we seem to get a bunch of issues from Goland users. Ideally a bunch of 'em.
- Upgrading the build infra sounds good.

mkondratek:
It looks like we do not need these changes with the new intellij gradle platform plugin 2. closing!

PR #2175, opened 2024-08-31
fix/Webview: Read/write providers, views consistently.

dominiccooney:
Fixes a race where we collect two pieces of data from two maps with no coordination.

The race manifest by running the "commit" step when both pieces of data are available twice, hitting an assertion. However it is probably also possible to have a different outcome where nobody does the commit step.

The fix simply guards access to both maps under the monitor for the first map.

Fixes CODY-3569

## Test plan

Tested manually:

- Start the extension and open the sidebar.
- Spam the "new chat" action (Ctrl-Alt-0 on Windows for example.)


mkondratek (approved):

pkukielka (approved):
LGTM

PR #2176, opened 2024-08-31
CODY-3538: Fixed chat-id bug in migration

jamesmcnamara:
Tie in for https://github.com/sourcegraph/cody/pull/5407. Properly constructs chat transcripts and during import.

## Test plan
Updated unit tests

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/config/migration/ChatHistoryMigration.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/config/migration/ChatHistoryMigration.kt
@@ -1,9 +1,5 @@
 package com.sourcegraph.cody.config.migration
 
-import java.time.LocalDateTime
-import java.time.ZoneOffset
-import java.time.format.DateTimeFormatter
-import java.util.Locale
 import com.intellij.openapi.project.Project
 import com.sourcegraph.cody.agent.CodyAgentService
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/sourcegraph/cody/config/migration/ChatHistoryMigration.kt
```diff
@@ -33,16 +33,16 @@ object ChatHistoryMigration {
     return chats
         .map { (account, chats) ->
           val serializedChats = chats.mapNotNull(::toSerializedChatTranscript)
-          val byId = serializedChats.associateBy { it.id }
```
mkondratek, line 36: To be considered: 
- we could change the migration key in 
```
    RunOnceUtil.runOnceForProject(project, "CodyMigrateChatHistory") {
      ChatHistoryMigration.migrate(project)
    }
```

Instead of `CodyMigrateChatHistory` use e.g. `CodyMigrateChatHistory-v2`. So that whoever installs the new version will get the migration even though it happened before (but in the wrong way).

pkukielka:
@jamesmcnamara I noticed that if I open old chat and ask a new question in that conversation, the date of the chat in the history does not change, and the chat is not moved on the top of the list.

I'm not sure if it's expected or not, but I think we can live with it (at least for now).

pkukielka (approved):
LGTM

PR #2190, opened 2024-09-02
Change CodyAuthenticationManager from project to application level service

pkukielka:
## Changes

After discussion with @danielmarquespt we decided account settings should be global, shared across all IntelliJ projects.

To reflect that fact I'm changing `CodyAuthenticationManager` to be APP level service instead of project level.
Notifications about account state change are pushed to all open projects.
All projects use shared storage, so changes done when project is closed are also properly loaded upon opening that project.

All relevant changes are in the `CodyAuthenticationManager`, all the rest is removal of unneeded `project` parameter.

## Test plan

**Scenario 1**
1. Sign-in with any account
2. Restart the IntelliJ
3. Make sure you are still logged in

**Scenario 2**
1. Sign-in with any account, add a few more
2. Restart the IntelliJ
3. Make sure you are still logged in, with the same account

**Scenario 3**
1. Sign-in with any account, add a few more
2. Switch account
3. Restart the IntelliJ
4. Make sure you are still logged in, with the same account you switched to

**Scenario 4**
1. Sign-in with any account, add a few more
2. Sign-out
3. Restart the IntelliJ
4. Make sure you are not logged in

**Scenario 5**
1. Sign-in with any account, add a few more
2. Switch account
3. Open a second IntelliJ project
4. Make sure in the second IntelliJ window you are logged in, with the same account you switched to

**Scenario 6**
1. Open two IntelliJ projects side by side
2. Open `Account` panel in both of them
3. Make sure that switching account in one of them causes switch to the same account on the second

**Scenario 7**
1. Open two IntelliJ projects side by side
2. Open `Account` panel in both of them and make sure they have active the same account
3. Close one of the instances
4. Switch account in the remaining instance
5. Close the second instance and open back the first one
6. Make sure account is properly switched

### DIFF
--- a/src/main/java/com/sourcegraph/config/CodyAuthNotificationActivity.java
+++ b/src/main/java/com/sourcegraph/config/CodyAuthNotificationActivity.java
@@ -25,7 +25,7 @@ public class CodyAuthNotificationActivity implements Activity {
 
   @Override
   public void runActivity(@NotNull Project project) {
-    CodyAccount activeAccount = CodyAuthenticationManager.getInstance(project).getAccount();
+    CodyAccount activeAccount = CodyAuthenticationManager.getInstance().getAccount();
     CodyAccountManager service =
         ApplicationManager.getApplication().getService(CodyAccountManager.class);
 
--- a/src/main/java/com/sourcegraph/find/browser/JSToJavaBridgeRequestHandler.java
+++ b/src/main/java/com/sourcegraph/find/browser/JSToJavaBridgeRequestHandler.java
@@ -42,7 +42,7 @@ public class JSToJavaBridgeRequestHandler {
     try {
       switch (action) {
         case "getConfig":
-          return createSuccessResponse(ConfigUtil.getConfigAsJson(project));
+          return createSuccessResponse(ConfigUtil.getConfigAsJson());
         case "getTheme":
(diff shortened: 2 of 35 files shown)


mkondratek:
Works properly according to the test plan. ✅ 

Personally I do not like **Scenario 6** and the constraint that I must have the same account in all the instances. I could have some conversation in one project and change the account in the other one. My discussion will disappear. Clearly, I can switch back to the original account and retrieve the chat from the history. However, it feels bad for me. I'd expect that account changes in one project does not affect my work in others. 

Why is it needed to have the account switching logic so linked b/w the projects?

mkondratek (approved):
I do not like this behavior but I am almost sure that the scenario of having multiple accounts is very rare. Probably we are the ones playing with it the most. 

Feel free to merge it if you are convinced.

pkukielka:
> Why is it needed to have the account switching logic so linked b/w the projects?

Both options have pros and cons. 
If we won't do that, then we should also require logging-in into each project separately, as mixing those models leads to confusion (e.g. what will happen when you sign-out)?
I think for most users 'login once - use everywhere' would be an expected behaviour. 
Using different accounts on different IJ instances indeed seems a rare and a bit bizarre scenario.

PR #2206, opened 2024-09-03
Fix windows build

pkukielka:
## Changes

Finally fixes windows gradle/pnpm build.

Couple of additional deps I had to install manually (but it could be due to my `node` version or some other setup detail):
`npx install gulp`
`npx install gulp-cli`
`npx install cross-env`
`npx update-browserlist-db@latest` // possibly not needed

Also, windows comandline can be more picky, e.g.
`.\gradlew :runIDE -PplatformRuntimeVersion='2024.1.2'`
(notice '' around a version)

## Test plan

N/A


mkondratek (approved):
Nice

PR #2225, opened 2024-09-05
Register widgets for disposal

mkondratek:
## Test plan
1. n/a

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/edit/LensesService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/edit/LensesService.kt
@@ -1,6 +1,5 @@
 package com.sourcegraph.cody.edit
 
-import com.intellij.openapi.application.runInEdt
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.components.service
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/edit/LensesService.kt
```diff
@@ -57,46 +57,45 @@ class LensesService(val project: Project) {
...
-              }
-              .toMap()
-
-      lensGroups[uri]?.values?.forEach { Disposer.dispose(it) }
-      newLensGroups.forEach { (taskId, rangeAndLensGroup) ->
-        val (range, lensGroup) = rangeAndLensGroup
-        val isNewTask = !lensGroups.containsKey(taskId)
-        lensGroup.show(range, shouldScrollToLens = isNewTask)
-      }
+    val vf = CodyEditorUtil.findFileOrScratch(project, uri) ?: return
```
mkondratek, line 62: only removed `runInEdt {}` wrapping here as we do it where we call the method

pkukielka (approved):
LGTM

PR #2227, opened 2024-09-05
fix/Webview: Make Cody: Restart Cody close Webview panels and restore Webview views

dominiccooney:
Fixes CODY-3546.

Hook the "stop agent" method to clean up webviews.

Some tidy ups:

- Move webviews into their own package.
- Break large webview files apart, and organize them by layer.
- Add a README.md describing this structure.
- Be explicit about UNIX-style line endings with Spotless.

## Test plan

1. Open some chats, chat in the sidebar
2. Shift-shift, Cody: Restart Cody action
3. Verify that the chat panels close and the sidebar reloads

## Changelog

The webview-based chat now works after restarting Cody.


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentService.kt
```diff
@@ -119,9 +120,12 @@ class CodyAgentService(private val project: Project) : Disposable {
     return codyAgent
   }
 
-  fun stopAgent(project: Project?): CompletableFuture<Unit>? {
+  fun stopAgent(project: Project?): CompletableFuture<Void>? {
     try {
-      return codyAgent.getNow(null)?.shutdown()
+      val shutdownFuture = codyAgent.getNow(null)?.shutdown()
+      return (shutdownFuture ?: CompletableFuture.completedFuture(null)).thenRun {
```
pkukielka: That makes it impossible to wait for original `shutdownFuture` to finish.
Why not just `shutdownFuture.thenRun { ... }`?

BTW. Do we care if webview if restarted before, after, or just roughly in the same time as agent? I think it needs working agent anyway, so the above should do no harm.

dominiccooney: > That makes it impossible to wait for original shutdownFuture to finish.

@pkukielka could you elaborate?

If `shutdownFuture` is non-`null`, then `(shutdownFuture ?: CompletableFuture.completedFuture(null)).thenRun { ... }` &rArr; `shutdownFuture.thenRun { ... }`. Waiting for the result of that obviously also waits for the `shutdownFuture`.

If `shutdownFuture` is `null`, then there *is* no `shutdownFuture` to wait for.

> Do we care if webview if restarted before, after, or just roughly in the same time as agent? I think it needs working agent anyway, so the above should do no harm.

Depends exactly what you mean by "restart". But it is cleanest if agent stops, we clean up a bit of webview state like which tool windows are occupied, and then the next agent session starts.

dominiccooney: > Why not just shutdownFuture.thenRun { ... }?

Because `shutdownFuture` might be null.

> That makes it impossible to wait for original shutdownFuture to finish.

Sorry, this was clowny, 🤡  I have improved it. PTAL.

> BTW. Do we care if webview if restarted before, after, or just roughly in the same time as agent? I think it needs working agent anyway, so the above should do no harm.

Ideally we stop agent, then clean up a bit of web UI state like which Tool Windows are occupied, and then start the next agent.

Stopping the webviews first isn't great, because agent might be busy making new ones.

pkukielka (approved):
LGTM, thank for the fix!

Issue #1988, opened 2024-08-06
JetBrains: Crash occurred while deleting suggestion generated by autocomplete


jay-fibi:
### IDE Information

IntelliJ IDEA 2023.3.3 (Community Edition)
Build #IC-233.14015.106, built on January 25, 2024
Runtime version: 17.0.9+7-b1087.11 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 14.3
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 8
Metal Rendering is ON
Registry:
  ide.experimental.ui=true
Non-Bundled Plugins:
  idea.plugin.protoeditor (233.13135.65)
  DevKit (233.14015.152)
  gherkin (233.11799.165)
  cucumber-java (233.11799.196)
  PythonCore (233.14015.106)
  com.sourcegraph.jetbrains (6.0.22-nightly)
Kotlin: 233.14015.106-IJ

### Bug Description

Platform: macOS

**Steps to reproduce:**

1. Have Cody plugin installed in your JB IntelliJ editor
2. Login to Cody with Free/Pro user
3. Type something in a source code file and trigger Autocompletion
4. Clear the auto completed code

### Expected behavior:
Crash should not occur.

### Observed behavior:
Crash occurred with following stacktrace.

Stacktrace:
```text
java.lang.IndexOutOfBoundsException: Wrong offset: 194. Should be in range: [0, 189]
	at com.intellij.openapi.editor.impl.LineSet.findLineIndex(LineSet.java:164)
	at com.intellij.openapi.editor.impl.DocumentImpl.getLineNumber(DocumentImpl.java:1001)
	at com.sourcegraph.cody.agent.intellij_extensions.DocumentKt.codyRange(Document.kt:16)
	at com.sourcegraph.cody.inspections.CodyFixHighlightPass.doCollectInformation(CodyFixHighlightPass.kt:53)
	at com.intellij.codeHighlighting.TextEditorHighlightingPass.collectInformation(TextEditorHighlightingPass.java:55)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$doRun$1(PassExecutorService.java:406)
	at com.intellij.platform.diagnostic.telemetry.helpers.TraceKt.runWithSpanIgnoreThrows(trace.kt:76)
[...]

PR #2252, opened 2024-09-09
Fix IndexOutOfBounds from the highlight pass

mkondratek:
## Test plan
Try these examples:
- https://github.com/sourcegraph/jetbrains/issues/2110
- https://github.com/sourcegraph/jetbrains/issues/1988

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/intellij_extensions/Document.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/intellij_extensions/Document.kt
@@ -4,6 +4,7 @@ import com.intellij.openapi.editor.Document
 import com.sourcegraph.cody.agent.protocol_extensions.Position
 import com.sourcegraph.cody.agent.protocol_generated.Position
 import com.sourcegraph.cody.agent.protocol_generated.Range
+import kotlin.math.min
 
 fun Document.codyPosition(offset: Int): Position {
   val line = this.getLineNumber(offset)
@@ -13,7 +14,7 @@ fun Document.codyPosition(offset: Int): Position {
 }
 
 fun Document.codyRange(startOffset: Int, endOffset: Int): Range {
-  val startLine = this.getLineNumber(startOffset)
+  val startLine = this.getLineNumber(min(startOffset, this.textLength))
   val lineStartOffset1 = this.getLineStartOffset(startLine)
   val startCharacter = startOffset - lineStartOffset1
 
--- a/src/main/kotlin/com/sourcegraph/cody/agent/intellij_extensions/Document.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/intellij_extensions/Document.kt
@@ -18,7 +18,7 @@ fun Document.codyRange(startOffset: Int, endOffset: Int): Range {
   val lineStartOffset1 = this.getLineStartOffset(startLine)
   val startCharacter = startOffset - lineStartOffset1
 
-  val endLine = this.getLineNumber(endOffset)
+  val endLine = this.getLineNumber(min(endOffset, this.textLength))
   val lineStartOffset2 =
       if (startLine == endLine) {
         lineStartOffset1
--- a/src/main/kotlin/com/sourcegraph/cody/agent/intellij_extensions/Document.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/intellij_extensions/Document.kt
@@ -4,7 +4,6 @@ import com.intellij.openapi.editor.Document
 import com.sourcegraph.cody.agent.protocol_extensions.Position
 import com.sourcegraph.cody.agent.protocol_generated.Position
 import com.sourcegraph.cody.agent.protocol_generated.Range
-import kotlin.math.min
 
 fun Document.codyPosition(offset: Int): Position {
   val line = this.getLineNumber(offset)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/intellij_extensions/Document.kt
```diff
@@ -13,6 +13,14 @@ fun Document.codyPosition(offset: Int): Position {
 }
 
 fun Document.codyRange(startOffset: Int, endOffset: Int): Range {
+  if (startOffset < 0 ||
+      startOffset > this.textLength ||
+      endOffset > this.textLength ||
+      startOffset > endOffset) {
+    throw IllegalArgumentException(
```
mkondratek, line 20: Initially I wanted to bound the values here. But maybe we should not do that? We should be aware of what arguments we pass here.

src/main/kotlin/com/sourcegraph/cody/inspections/CodyFixHighlightPass.kt
```diff
@@ -97,6 +101,12 @@ class CodyFixHighlightPass(val file: PsiFile, val editor: Editor) :
                 null
               }
             }
+            .toList()
+
+    if (protocolDiagnostics.isEmpty()) {
```
mkondratek, line 83: it was possible that we have been sending an empty list 😅 that was completely redundant

src/main/kotlin/com/sourcegraph/utils/CodyFormatter.kt
```diff
@@ -35,9 +36,8 @@ class CodyFormatter {
                 .createFileFromText("TEMP", file.fileType, appendedString)
 
         val codeStyleManager = CodeStyleManager.getInstance(project)
-        // This will fail if cursor < range.startOffset + completionText.length
-        // TODO change the signature of this method or at least validate the arguments
-        codeStyleManager.reformatText(psiFile, cursor, range.startOffset + completionText.length)
+        val endOffset = max(cursor, range.startOffset + completionText.length)
+        codeStyleManager.reformatText(psiFile, cursor, endOffset)
```
mkondratek, line 40: Unrelated to highlight pass but I managed to get it once

src/main/kotlin/com/sourcegraph/cody/inspections/CodyFixHighlightPass.kt
```diff
@@ -106,15 +116,18 @@ class CodyFixHighlightPass(val file: PsiFile, val editor: Editor) :
...
           val range = document.codyRange(highlight.startOffset, highlight.endOffset)
 
           if (myRangeActions.containsKey(range)) {
             continue
           }
-          val location =
-              ProtocolLocation(
-                  uri = protocolTextDocument.uri,
-                  range = document.codyRange(highlight.startOffset, highlight.endOffset))
+          val location = ProtocolLocation(uri = protocolTextDocument.uri, range = range)
```
pkukielka: Perhaps we can get rid of that whole `protocolTextDocument`?
It seems we need it only for uri, and we can get is using `ProtocolTextDocument::uriFor`.
It is not marked as requiring EDT so there is a good chance we will also not need that future, etc..

mkondratek: btw, doCollectInformation is called from the background thread by IntelliJ

image
(we are extending HighlightPass and collectInformation calls doCollectInformation)

src/main/kotlin/com/sourcegraph/cody/inspections/CodyFixHighlightPass.kt
```diff
@@ -106,15 +116,18 @@ class CodyFixHighlightPass(val file: PsiFile, val editor: Editor) :
           if (progress.isCanceled) {
             break
           }
+          if (highlight.startOffset > document.textLength ||
```
pkukielka: We are now reading document in outside of the EDT.
Your check probably highly decreases chance of the issue, but I doubt it eliminates it as state can change even between this and the next instruction.
I think `document.codyRange` should be called on EDT.

src/main/kotlin/com/sourcegraph/cody/inspections/CodyFixHighlightPass.kt
mkondratek: good point about the URI - fixing!

src/main/kotlin/com/sourcegraph/cody/inspections/CodyFixHighlightPass.kt
mkondratek: fixed

src/main/kotlin/com/sourcegraph/cody/inspections/CodyFixHighlightPass.kt
```diff
@@ -137,6 +131,12 @@ class CodyFixHighlightPass(val file: PsiFile, val editor: Editor) :
     for (highlight in myHighlights) {
       highlight.unregisterQuickFix { it.familyName == CodeActionQuickFix.FAMILY_NAME }
 
+      if (highlight.startOffset > document.textLength ||
+          highlight.endOffset > document.textLength ||
+          highlight.startOffset > highlight.endOffset) {
+        break
+      }
+
       val range = document.codyRange(highlight.startOffset, highlight.endOffset)
```
pkukielka, line 142: Is that running in the EDT?

mkondratek: It runs in EDT. I added `@RequiresEdt` on it.

pkukielka (approved):
LGTM

PR #325, opened 2024-01-22
Fix Sourcegraph actions (search selection, open selection, etc)

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/252.
Fixes https://github.com/sourcegraph/jetbrains/issues/162.
Fixes https://github.com/sourcegraph/jetbrains/issues/324.

## Test plan
- [x] test Sourcegraph actions with Cody **enabled**
- [x] test Sourcegraph actions with Cody **disabled**

PR #2257, opened 2024-09-09
Use Agent for parsing Git repo names (CODY-3132)

jamesmcnamara:
[CODY-3132](https://linear.app/sourcegraph/issue/CODY-3132/codys-convertgitcloneurltocodebasename-isnt-able-to-parse-some-git)

There was a duplication of effort around the convertGitURLToCodebaseName as it was implemented in the agent and here. There was even already an agent method for this, but it was unused. Thus when fixing the above on the agent side, I thought it made sense to switch it here. 
  
## Test plan
I'm not sure the best way to test this, so input would be appreciated! Do we have tests for agent integration?


src/main/java/com/sourcegraph/vcs/RepoUtil.java
```diff
@@ -128,7 +132,25 @@ private static String doReplacements(
 
     if (vcsType == VCSType.GIT && repository != null) {
       String cloneURL = GitUtil.getRemoteRepoUrl((GitRepository) repository, project);
-      return convertGitCloneURLToCodebaseNameOrError(cloneURL).getValue();
+      AtomicReference<String> repoURl = new AtomicReference<>();
```
dominiccooney: This isn't right. `withAgent` runs the callback on a separate thread, so you will often--but not always--have null in `repoURl`.

You need to:

- Make `getRemoteRepoUrl` return a CompletableFuture<String?>
- Convert the callers to handle an asynchronous result. This may take care and attention to not block EDT, not starve a threadpool, ensure that if you started on EDT you continue on EDT if the caller requires EDT, etc.
- Convert the method body to handle the asynchronous result: Create the future, run the code with Agent, complete the result in the callback.

By the way, let's pay attention to details like `repoURL` or `repoUrl` but not `repoURl`

src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
```diff
@@ -1,50 +0,0 @@
...
-    return CodebaseName("$host/$owner/${repo.replace(".git$".toRegex(), "")}")
-  }
-
-  var uri = URI(cloneURL)
-  if (uri.scheme == null) {
-    uri = URI("http://$cloneURL")
-  }
-
-  // Handle Azure DevOps URLs
-  if (uri.host?.contains("dev.azure") == true && !uri.path.isNullOrEmpty()) {
```
dominiccooney, line 23: Are all of these cases handled in unit tests on the Agent side?

dominiccooney (changes requested):
Good intuition to de-dup something. In particular, the Kotlin version of this function doesn't handle server-side repo renaming.

Unfortunately the implementation doesn't look right--see the comment inline for the approach we need here.

src/main/java/com/sourcegraph/vcs/ConvertUtil.kt
jamesmcnamara: Yes [in this file](https://github.com/sourcegraph/cody/blob/fb5f4c66c21ec0057ae02a6602a0bca9b483289b/lib/shared/src/utils.test.ts). In fact it looks like the tests here were carbon copied from there.

src/main/java/com/sourcegraph/vcs/RepoUtil.java
jamesmcnamara: Ah for some reason I thought it was blocking. I've updated it in [this commit](https://github.com/sourcegraph/jetbrains/pull/2257/commits/a40f65eb9dcedeb28addc9f3b2e3cdbe1ca1ba9a?diff=split&w=0). Do you know if there are any unit tests of agent integration? Would this be a good use case for an integration test?

pkukielka:
This function was in agent long time ago, and was copied to jetbrains because it turned out to be problematic.
In particular I think it was used in paths where agent was not started yet and we didn't want to start it yet.
I do not know if it's still true.

Also actions like `OpenRevisionAction` were now working even with Cody disabled, which won't be the case after this change. 
So it at least requires careful testing and consideration.

jamesmcnamara:
@pkukielka @dominiccooney agreed on the testing required. Do either of you have a suggestion of how to test this? Do we have any tests for agent communication? Should this be an integration test?

pkukielka:
@jamesmcnamara
https://github.com/sourcegraph/jetbrains/pull/2442

I removed part of the code which was not needed and fixed the duplicated code.
I know it's not ideal, but not this method is used only to generate sourcegraph links, and this functionality is supposed to work even if cody is disabled, so we cannot use agent for it.

mkondratek (commented):
image

https://github.com/sourcegraph/jetbrains/pull/325

The option - `Enable Cody` - decides whether the agent runs or not. 
Users can have `Enable Cody` turned off and still want to use Sourcegraph Actions.

image

pkukielka:
Closing this as:
* we fixed this on the JetBrains side
* it would be causing part of our features to not work properly

PR #2303, opened 2024-09-16
Move from the deprecated `horizontalAlign`

mkondratek:
A prerequisite for https://github.com/sourcegraph/jetbrains/pull/2304. 

## Test plan
1. Verify Settings UI

PR #2304, opened 2024-09-16
Add CodyInlineCompletionProvider as an alternative for inlay completions

mkondratek:
Fixes https://linear.app/sourcegraph/issue/CODY-3742.

# Based on PR #2303

## Full chain of PRs as of 2024-09-16

* PR #2304:
  `mkondratek/feat/completion-provider` ➔ `mkondratek/chore/ui-fixes`
* PR #2303:
  `mkondratek/chore/ui-fixes` ➔ `main`

## Test plan

AUTOMATIC
1. Run IDE in Remote Development mode 
2. Open a project, open a file
3. Hit Enter to trigger autocomplete 
Expected: autocompletion suggested

INVOKE
1. `shift + option + \` on **macOS**
Expected: autocompletion suggested

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -507,8 +507,7 @@ tasks {
 
   buildPlugin {
     dependsOn(project.tasks.getByPath("buildCody"))
-    val get = composedJar.get()
-    get.exclude("com/intellij/codeInsight/inline/completion/**")
+    composedJar.get().exclude("com/intellij/codeInsight/inline/completion/**")
     from(
         fileTree(buildCodyDir) {
           include("*")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -478,7 +478,7 @@ tasks {
           "cody-agent.panic-when-out-of-sync" to
               (System.getProperty("cody-agent.panic-when-out-of-sync") ?: "true"),
           "cody-agent.fullDocumentSyncEnabled" to
-              (System.getProperty("cody-agent.fullDocumentSyncEnabled") ?: "true"),
+              (System.getProperty("cody-agent.fullDocumentSyncEnabled") ?: "false"),
           "cody.autocomplete.enableFormatting" to
               (project.property("cody.autocomplete.enableFormatting") ?: "true"))
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -171,14 +171,6 @@ kotlin {
   }
 }
 
-sourceSets {
-  create("integrationTest") {
-    kotlin.srcDir("src/integrationTest/kotlin")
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyInlineCompletionProvider.kt
```diff
@@ -0,0 +1,206 @@
...
+    val completions =
+        fetchCompletions(project, editor, InlineCompletionTriggerKind.AUTOMATIC, cancellationToken)
+            .get() ?: return emptyList()
+
+    // Create a list to store InlineCompletionElement objects
+    val result = mutableListOf<InlineCompletionElement>()
+
+    // Add completions to the result
+    completions.items.forEach { completion ->
+      result.add(InlineCompletionElement(completion.insertText))
```
mkondratek: The problem is that `InlineCompletionElement` in platform version 232 does not support offsets. Completions need to be trimmed otherwise they look like this:
image

build.gradle.kts
```diff
@@ -580,19 +590,6 @@ tasks {
 
   test { dependsOn(project.tasks.getByPath("buildCody")) }
 
-  configurations {
```
pkukielka, line 583: Was that not needed?

build.gradle.kts
```diff
@@ -499,6 +507,8 @@ tasks {
 
   buildPlugin {
     dependsOn(project.tasks.getByPath("buildCody"))
+    val get = composedJar.get()
```
pkukielka: `get` does not sound like nice name ^^
Maybe just inline it?

src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
```diff
@@ -25,6 +25,7 @@ class CodyDocumentListener(val project: Project) : BulkAwareDocumentListener {
     }
   }
 
+  // todo: documentChanged
```
pkukielka: Remove or make it more meaningful please.

pkukielka (approved):
I left few comments, otherwise LGTM

build.gradle.kts
mkondratek: It looks so. the test are passing without it. Note that configuration creation is redundant. SourceSets persist.

build.gradle.kts
mkondratek: fixed

src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
mkondratek: fixed

PR #2305, opened 2024-09-16
Rewrite EditCommandPrompt implementation from custom JFrame to JBPopup

pkukielka:
Fixes CODY-3645

## Changes

Rewrite EditCommandPrompt implementation from custom JFrame to JBPopup.

JBPopup already contains some features we were re-implementing by hand, and it is laso guaranteed to be correctly handled by Gateway (as JetBrains uses it in a lot of places).

## Test plan

1. Hit `shift shift`
2. Search for `Edit Code...`
3. Type your edit command and run it
4. Make sure edit was applied correctly. Note that most likely you will see some side error due to faulty lenses implementation


dominiccooney (approved):
Great! Thank you for the quick work on this.

mkondratek (commented):
lgtm

PR #2313, opened 2024-09-17
Fixed a compile error caused by a simple name change

steveyegge:
Agent-side changed one of the message names; no other changes were needed

## Test plan

Locally tested.


dominiccooney (changes requested):
This appears broken in CI, you might need to roll the Cody agent version at the same time...

dominiccooney:
I am rolling the Cody commit to pull in the upstream change this fix depends on.

dominiccooney (approved):
LGTM once we roll the Cody commit.

PR #2319, opened 2024-09-18
fix/Remote: Make chat work in remote mode

dominiccooney:
## Test plan

Tested manually:

```
./gradlew :customRunIde -PplatformRuntimeVersion=2024.2 -PplatformRuntimeType=IU -PsplitMode=true
```

1. Sign in to Cody and do some chats.
2. Change themes, check that the theme updates.

Also works in Remote:

Screenshot

## Changelog

- Chat works in remote mode.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
@@ -3,6 +3,7 @@ package com.sourcegraph.cody
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
+import com.intellij.ui.components.JBLabel
 import com.intellij.util.concurrency.annotations.RequiresEdt
 import com.intellij.util.ui.UIUtil
 import com.sourcegraph.cody.chat.SignInWithSourcegraphPanel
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -40,29 +52,53 @@ class CodyToolWindowContent(project: Project) {
...
         (it as CodyOnboardingGuidancePanel).updateDisplayName(displayName)
       }
-      allContentLayout.show(allContentPanel, ONBOARDING_PANEL)
+      cardLayout.show(cardPanel, ONBOARDING_PANEL)
+      showView(cardPanel)
       return
     }
-    allContentLayout.show(allContentPanel, MAIN_PANEL)
+    cardLayout.show(cardPanel, LOADING_PANEL)
+    showView(webviewPanel ?: cardPanel)
```
pkukielka, line 70: That looks like potential source of hidden issues.
If `webviewPanel` will be null when we expect it to be not, and we switch to cardPanel instead, won't it be stuck forever in non-webview mode?

In what cases this can even happen?

dominiccooney: This is just like `MAIN_PANEL` used to be: either it is the webview, or it is the loading indicator.

It won't necessarily be stuck forever in loading mode: When the webview starts up, it sets the webview component and calls `showView` to show it, see `setWebviewComponent`.

pkukielka (approved):
LGTM

PR #2326, opened 2024-09-20
Fix cody status icon in remote environments

pkukielka:
Fixes CODY-3710

## Changes

Looks like rendering status icons is not supported in the remote environments.
Instead, we are forced to display a text replacement.

## Test plan

1. Run Cody in the remote client mode (you can use `:customRunIde -PsplitMode=true`)
2. Make sure "Cody" text is visible instead of the icon

image

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/statusbar/CodyStatusBarWidget.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/statusbar/CodyStatusBarWidget.kt
@@ -11,6 +11,7 @@ import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.wm.StatusBarWidget
 import com.intellij.openapi.wm.WindowManager
 import com.intellij.openapi.wm.impl.status.EditorBasedStatusBarPopup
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/statusbar/CodyStatusBarWidget.kt
```diff
@@ -20,7 +21,9 @@ class CodyStatusBarWidget(project: Project) : EditorBasedStatusBarPopup(project,
     if (currentStatus == CodyStatus.CodyDisabled) {
       return WidgetState.HIDDEN
     }
-    val state = WidgetState(currentStatus.presentableText, "", true)
+    // Remote environment does not support rendering icons, so we use a placeholder text instead
+    val text = if (GraphicsUtil.isRemoteEnvironment()) "Cody" else ""
```
pkukielka, line 25: @dominiccooney Looks like the remote checks I suggested to you before does not work as I expected, but this does.

mkondratek (approved):

PR #2349, opened 2024-09-25
Fix logic for picking-up selected editor when using remote

pkukielka:
Fixes CODY-3831

## Changes

`FileEditorManager.getInstance(it).selectedTextEditor` does not report selected editor properly in the remote mode.
Instead we should always use `FileEditorManager.getInstance(project).selectedTextEditorWithRemotes` or pick proper editor from the list of all open ones.

## Test plan

1. Start IntelliJ in the dev remote mode (by using `:customRunIde -PsplitMode=true`)
2. Run any edit command
3. Do any focus change in the editor - there should be no desync issues reported in the Cody log window

### DIFF
--- a/src/main/java/com/sourcegraph/website/SearchActionBase.java
+++ b/src/main/java/com/sourcegraph/website/SearchActionBase.java
@@ -7,13 +7,13 @@ import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.editor.SelectionModel;
 import com.intellij.openapi.fileEditor.FileDocumentManager;
-import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.sourcegraph.common.BrowserOpener;
 import com.sourcegraph.common.ui.DumbAwareEDTAction;
 import com.sourcegraph.find.SourcegraphVirtualFile;
+import com.sourcegraph.utils.CodyEditorUtil;
 import com.sourcegraph.vcs.RepoInfo;
 import com.sourcegraph.vcs.RepoUtil;
 import com.sourcegraph.vcs.VCSType;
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/sourcegraph/utils/CodyEditorUtil.kt
```diff
@@ -75,13 +75,19 @@ object CodyEditorUtil {
   }
 
   @JvmStatic
-  fun getFocusedEditorForAnActionEvent(e: AnActionEvent): Editor? {
-    return e.project?.let { FileEditorManager.getInstance(it).selectedTextEditor }
+  fun getSelectedEditors(project: Project): Array<out Editor> {
+    return FileEditorManager.getInstance(project).selectedTextEditorWithRemotes
```
mkondratek, line 79: 👍

mkondratek (approved):
lgtm

src/main/java/com/sourcegraph/website/SearchActionBase.java
```diff
@@ -31,10 +32,14 @@ public void actionPerformedMode(@NotNull AnActionEvent event, @NotNull Scope sco
     if (selectedText == null || selectedText.length() == 0) {
       return;
     }
+    Editor editor =
+        Arrays.stream(CodyEditorUtil.getSelectedEditors(project)).findFirst().orElse(null);
+    if (editor == null) {
+      return;
+    }
```
dominiccooney: Is there no `firstOrNull()`? instead of `findFirst().orElse(null)`?

Can't we early return in the expression context with `?:`

```suggestion
    Editor editor =
        Arrays.stream(CodyEditorUtil.getSelectedEditors(project)).findFirst().orElse(null) ?: return
```

src/main/java/com/sourcegraph/website/SearchActionBase.java
```diff
@@ -103,7 +108,8 @@ private String getSelectedText(@Nullable Project project) {
       return null;
     }
 
-    Editor editor = FileEditorManager.getInstance(project).getSelectedTextEditor();
+    Editor editor =
+        Arrays.stream(CodyEditorUtil.getSelectedEditors(project)).findFirst().orElse(null);
```
dominiccooney: This line is happening a *lot*, can we put it into `CodyEditorUtil` and be done with it?

dominiccooney (approved):
LGTM too, thank you for finding and fixing this 👨‍🍳 

Some stylistic suggestions inline at your discretion.

src/main/java/com/sourcegraph/website/SearchActionBase.java
pkukielka: Done :)

PR #2372, opened 2024-10-02
Remove dep on `m.intellij.modules.lang`

mkondratek:
Fixes https://linear.app/sourcegraph/issue/CODY-3895/drop-dependency-on-comintellijmoduleslang-to-support-gateway.

I removed this dep. I tried our features. I see no problems. Verifier does not complain about it missing. I seems that the dependence is redundant. 

## Test plan
1. Full QA, try autocomplete


pkukielka (approved):
LGTM

PR #2373, opened 2024-10-02
Remove deprecated code related to chat export

mkondratek:
The code related to `ContextItem` and `ChatMessage` is redundant once we moved to the webview UI. Let's get rid of it.

## Test plan
Case 1: Export from History tab
Case 2: Cody: Export All Chats as JSON (run action)


dominiccooney (approved):
Splendid. Perfection is achieved when there is nothing left to take away.

src/test/kotlin/com/sourcegraph/cody/agent/protocol/UriUtilsTest.kt
```diff
@@ -3,47 +3,15 @@ package com.sourcegraph.cody.agent.protocol
...
-
-  fun `testGetLinkActionText - remote file`() {
-    val linkActionText = remoteContextItemFile.getLinkActionText(projectPath = null)
-    assertEquals("jetbrains TESTING.md:49-56", linkActionText)
-  }
+class UriUtilsTest : BasePlatformTestCase() {
 
   fun `test getPath`() {
-    fun contextFilePath(path: String) = ContextItemFile(uri = URI.create(path)).getPath().toString()
+    fun contextFilePath(path: String) = VfsUtil.toUri(path)?.toPath().toString()
```
dominiccooney: Looks like this needs work, see the CI test failure.

mkondratek: I removed `test getPath` - previously it was testing `ContextFileItem::getPath`. That class has been removed and so can be this test case.

PR #540, opened 2024-02-08
Remove JetBrains Client compatibility

exigow:
Fixes https://github.com/sourcegraph/jetbrains/issues/366 
Fixes https://github.com/sourcegraph/jetbrains/issues/537 
Fixes https://github.com/sourcegraph/sourcegraph/issues/57569

Motivation: Multiple features are simply not working on this "thin" IDE (see: linked issues) due to missing classes. 

More details here: #537

JetBrains client is prefixed with "JBC" (for example: JBC-232.10227.8).

## Test plan

You shouldn't be able to install Cody on JetBrains Client. It's hard to reproduce (I had issues to download & use this), however this can be simply verified on Cody plugin JetBrains Marketplace page > Versions > Compatible with... (currently JetBrains Client is listed there).

PR #2375, opened 2024-10-02
Fix `com.intellij.jetbrains.client` compatibility

mkondratek:
Reverts https://github.com/sourcegraph/jetbrains/pull/540.

We encountered an issue on when insallig the plugin on windows. Maybe we can remove it... (manual testing in progress)

## Test plan
1. Remote Dev works on windows (install plugin from disk on windows in remote dev)


pkukielka (approved):
LGTM

PR #2382, opened 2024-10-03
Switch to webview-based sign-in flow

pkukielka:
Fixes [CODY-4347](https://linear.app/sourcegraph/issue/CODY-4347)

## Test plan

Full QA, I will add more instructions later.

Depends on:
1. https://github.com/sourcegraph/cody/pull/5794
2. https://github.com/sourcegraph/cody/pull/5918


src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
```diff
@@ -9,17 +9,17 @@ import com.sourcegraph.cody.agent.protocol.CompletionItemParams
...
 import com.sourcegraph.cody.autocomplete.CodyAutocompleteManager
 import com.sourcegraph.cody.autocomplete.action.AcceptCodyAutocompleteAction
-import com.sourcegraph.cody.chat.CodeEditorFactory
 import com.sourcegraph.cody.telemetry.TelemetryV2
 import com.sourcegraph.cody.vscode.InlineCompletionTriggerKind
 import com.sourcegraph.utils.CodyEditorUtil
 
 class CodyDocumentListener(val project: Project) : BulkAwareDocumentListener {
 
+  // TODO: Looks like this functionality is broken after the migration to webview
```
mkondratek: - [x] let us have a linear issue for this one

src/main/kotlin/com/sourcegraph/cody/config/migration/AccountsMigration.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.sourcegraph.cody.auth.CodyAccount
+import com.sourcegraph.cody.auth.depracated.DeprecatedCodyAccountManager
+
+object AccountsMigration {
+  fun migrate() {
+    val codyAccountManager = DeprecatedCodyAccountManager.getInstance()
+    codyAccountManager.getAccounts().forEach { oldAccount ->
+      val token = codyAccountManager.getTokenForAccount(oldAccount)
+      if (token != null) {
+        CodyAccount(oldAccount.server).storeToken(token)
```
mkondratek, line 12: nice

src/main/kotlin/com/sourcegraph/cody/api/SourcegraphApiRequests.kt
```diff
@@ -30,14 +27,16 @@ object SourcegraphApiRequests {
...
-                  }
-                }
-                .apply { operationName = "get profile avatar" })
+    class CodyAccountCodyProEnabled(val codyProEnabled: Boolean?)
+
+    class CodyAccountDetails(
+        val id: String,
+        val username: String,
+        val displayName: String?,
+        val avatarURL: String?
```
mkondratek, line 26: we do not use avatarURL anymore - I think we can remove it

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
```diff
@@ -123,6 +130,24 @@ class CodyAgentClient(private val project: Project, private val webview: NativeW
...
+    CodyAccount(SourcegraphServerPath(params.key)).storeToken(params.value)
+    return CompletableFuture.completedFuture(null)
+  }
+
+  @JsonRequest("secrets/delete")
+  fun secrets_delete(params: Secrets_DeleteParams): CompletableFuture<Null?> {
+    CodyAccount(SourcegraphServerPath(params.key)).storeToken(null)
+    return CompletableFuture.completedFuture(null)
+  }
+
```
mkondratek, line 154: maybe a stupid question but... is it safe? are there any security issues? would the secrets be present in the trace logs?

src/main/java/com/sourcegraph/config/CodyAuthNotificationActivity.java
```diff
@@ -79,15 +64,4 @@ public void actionPerformed(@NotNull AnActionEvent anActionEvent) {
     notification.addAction(neverShowAgainAction);
     Notifications.Bus.notify(notification);
   }
-
-  private void showMissingTokenNotification() {
```
mkondratek, line 83: btw, I wonder... what if the token expires (or is removed)? 🤔 what will happen (with chat, with autocomplete)?

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
mkondratek, line 1: Don't we want to keep this action? I know it is possible to export chats from History tab in the webview. Simply, this action is an actual jetbrains action and can be searched for and run from anywhere

mkondratek (commented):
in general looks good - I will try it in the morning 🌅

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
```diff
@@ -6,74 +6,30 @@ import com.intellij.openapi.project.Project
...
 import javax.swing.JComponent
 import javax.swing.JPanel
 
 @Service(Service.Level.PROJECT)
 class CodyToolWindowContent(project: Project) {
-  private val cardLayout = CardLayout()
-  private val cardPanel = JPanel(cardLayout)
-  val allContentPanel: JComponent = JPanel(GridLayout(1, 1))
+  val allContentPanel: JComponent = JPanel(CardLayout())
+  private val mainPanel = JPanel(GridBagLayout())
```
dominiccooney: We were using the cardlayout to swap between the loading, login and onboarding components.

We can't put the webview in a CardLayout because JetBrains remote doesn't render it, so we are swapping the components.

But now there are only two components to swap: Loading and webview. So we don't need this panel anymore.

 Let's just have `allContentPanel` with whatever trivial layout (GridLayout(1,1) etc.)

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
dominiccooney: We log the secrets in the agent trace, not the log. You have to set an extra environment variable to turn on that level of logging... I think it should be fine.

We already log tokens in that file today... the ExtensionConfiguration message from JetBrains to Agent has the token in it... see build/sourcegraph/cody-agent-trace.json after a debugging session.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
```diff
@@ -123,6 +130,24 @@ class CodyAgentClient(private val project: Project, private val webview: NativeW
...
   }
 
+  @JsonRequest("secrets/get")
+  fun secrets_get(params: Secrets_GetParams): CompletableFuture<String?> {
+    return CompletableFuture.completedFuture(
+        CodyAccount(SourcegraphServerPath(params.key)).getToken())
+  }
+
+  @JsonRequest("secrets/store")
+  fun secrets_store(params: Secrets_StoreParams): CompletableFuture<Null?> {
```
dominiccooney, line 144: A nullable null... Why not a nullable nullable null 🤔

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
```diff
@@ -226,4 +251,16 @@ class CodyAgentClient(private val project: Project, private val webview: NativeW
 
     return saveFileFuture
   }
+
+  @JsonNotification("window/didChangeContext")
+  fun window_didChangeContext(params: Window_DidChangeContextParams) {
+    println(params)
```
dominiccooney: Let's not spam the console. Info logging it might be fine.

Is this, in fact, important for refreshing the status bar icon? `"cody.activated"` will be `true` when the extension thinks it is good to go.

src/main/kotlin/com/sourcegraph/cody/auth/depracated/DeprecatedCodyAccountManager.kt
```diff
@@ -0,0 +1,101 @@
+package com.sourcegraph.cody.auth.depracated
```
dominiccooney: Spelling: deprecated

src/main/kotlin/com/sourcegraph/cody/auth/depracated/DeprecatedCodyAccountManager.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  fun isAccountUnique(name: String, server: SourcegraphServerPath) =
+      getAccounts().none { it.name == name && it.server.url == server.url }
+
+  fun getTokenForAccount(account: DeprecatedCodyAccount): String? =
+      PasswordSafe.instance.get(account.credentialAttributes())?.getPasswordAsString()
+
+  @RequiresEdt
+  internal fun updateAccountToken(account: DeprecatedCodyAccount, newToken: String) {
+    service<DeprecatedCodyPersistentAccounts>().accounts = (getAccounts() - account) + account
```
dominiccooney: This seems tricky, maybe the method should be called `addOrUpdateAccountToken` or something?

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -172,7 +172,7 @@ class CodyAutocompleteManager {
 
     val textDocument: TextDocument = IntelliJTextDocument(editor, project)
 
-    if (isTriggeredExplicitly && CodyAuthenticationManager.getInstance().hasNoActiveAccount()) {
+    if (isTriggeredExplicitly && !CodyAccount.hasActiveAccount()) {
       HintManager.getInstance().showErrorHint(editor, "Cody: Sign in to use autocomplete")
```
dominiccooney: Ugh, we should link to the sign in panel here! Maybe add a TODO?

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
dominiccooney: Yeah, for now we should keep this, and Explain, etc.

It would be a nice cleanup project later to ingest the VSCode package.json and consider wiring up those commands automatically--or at least linting that we have them.

src/main/kotlin/com/sourcegraph/cody/chat/actions/SmellCommand.kt
```diff
@@ -1,8 +0,0 @@
-package com.sourcegraph.cody.chat.actions
```
dominiccooney: This too, I believe we should keep these commands.

src/main/kotlin/com/sourcegraph/cody/chat/actions/NewChatAction.kt
```diff
@@ -12,8 +12,7 @@ class NewChatAction : DumbAwareEDTAction() {
   }
 
   override fun update(event: AnActionEvent) {
-    val hasActiveAccount = CodyAuthenticationManager.getInstance().hasActiveAccount()
-    event.presentation.isEnabled = hasActiveAccount
+    event.presentation.isEnabled = CodyAccount.hasActiveAccount()
```
dominiccooney, line 15: This `CodyAccount.hasActiveAccount`, we should be using the flag we got from `window/didChangeContext "cody.activated"` instead.

src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
```diff
@@ -61,16 +60,22 @@ class SettingsMigration : Activity {
...
+      DeprecatedChatLlmMigration.migrate(project)
+    }
+    RunOnceUtil.runOnceForProject(project, "ChatTagsLlmMigration") {
+      ChatTagsLlmMigration.migrate(project)
+    }
     RunOnceUtil.runOnceForProject(project, "CodyMigrateChatHistory-v2") {
       ChatHistoryMigration.migrate(project)
     }
+    RunOnceUtil.runOnceForProject(project, "AccountsToCodyMigration") {
+      AccountsMigration.migrate()
```
dominiccooney, line 73: Just curious but are these guaranteed to run *in this order*? Is it important?

dominiccooney (approved):
Feedback inline, please take a look.

Approving to unblock since you have multiple sides to land.

We can go further than this, I think, by listening to `window/didChangeContext` and hooking a notification to the AuthStatus observer, but we can work on those in follow up patches if it is easier.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
```diff
@@ -226,4 +251,16 @@ class CodyAgentClient(private val project: Project, private val webview: NativeW
...
+  }
+
+  @JsonNotification("globalStorage/didChange")
```
dominiccooney: Let's not sniff global storage for this. We should be listening to the configuration observable and pushing a semantic, application-level message to the IDE.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
pkukielka: It is a s Dominic said :)

src/main/java/com/sourcegraph/config/CodyAuthNotificationActivity.java
pkukielka: I will test this scenario today. In general I think we should then (if user try to call some action explicitly) show cody panel, and it should be showing login panel.

src/main/kotlin/com/sourcegraph/cody/CodyToolWindowContent.kt
pkukielka: So this is weird, because I tried to simplify it at the start but it was not working without `CardLayout`.
But maybe it was error on my side, I will check again :)

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
pkukielka: Yeah, refreshing status icon bar is part I/ know I missed but I forget to add it at the end.
Thanks, I will add it now!

src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
pkukielka: Actually I think we should just show it. 
I will do the changes, thanks!

src/main/kotlin/com/sourcegraph/cody/chat/actions/NewChatAction.kt
pkukielka: I agree in principle, but it is not as simple.
I need to also know account, in particular to be able to check if it's enterprise/pro/free, which in turn I need to properly render LLM selection in edit window.
But now I think maybe we should add additional flag to the model, and avoid this problem in the first place.
It would allow us to remove quite a bit of code on the jetbrains side.

src/main/kotlin/com/sourcegraph/cody/config/migration/SettingsMigration.kt
pkukielka: AFAIR their are, this code is synchronous.

src/main/java/com/sourcegraph/config/CodyAuthNotificationActivity.java
dominiccooney: By the way, VSCode has various APIs for showing views, we don't implement them. If we need to show the view, we should see if the extension is trying to do it anyway. Maybe we just need to implement a couple of those methods/properties in the VSCode shim.

src/main/kotlin/com/sourcegraph/cody/chat/actions/NewChatAction.kt
dominiccooney: OK. Let's work out a practical split to get this landed, because there's a lot of great cleanup here. And we can smooth out the details later.

I agree for `chat/models` returning the models with the flag sounds good.

I tend to disagree in the specific case of NewChatAction, all we need to know is whether we have an active account. But let's land this first and take a fresh look at account type dependencies after this change is done.

src/main/kotlin/com/sourcegraph/cody/chat/actions/ExportChatsAction.kt
pkukielka: Ok, I reverted that changes.

src/main/java/com/sourcegraph/cody/CodyToolWindowFactory.java
```diff
@@ -30,7 +29,6 @@ public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindo
           content.setPreferredFocusableComponent(null);
           toolWindow.getContentManager().addContent(content);
           DefaultActionGroup customCodySettings = new DefaultActionGroup();
-          customCodySettings.add(new OpenPluginSettingsAction("Cody Settings..."));
```
mkondratek, line 34: image

Probably this is not an item for this PR but our settings seems to be orphaned now 😢 Can we still have them under `Tools > Sourcegraph & Cody`?

src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIHost.kt
```diff
@@ -71,14 +69,7 @@ internal class WebUIHostImpl(
     val id = decodedJson?.get("id")?.asString
     val arg = decodedJson?.get("arg")?.asString
 
-    if ((command == "auth" && decodedJson.get("authKind")?.asString == "signout") ||
-        (isCommand && id == "cody.auth.signout")) {
-      CodyAuthenticationManager.getInstance().setActiveAccount(null)
-    } else if (isCommand && id == "cody.auth.switchAccount") {
```
mkondratek, line 77: `cody.auth.switchAccount` - we do not handle this command now 😕

src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
```diff
@@ -12,17 +12,17 @@ import com.sourcegraph.cody.agent.protocol.CompletionItemParams
...
```
mkondratek: is it still broken or did you manage fix it? if it need more attention can you pls create an issue for that? 🙏

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIHost.kt
pkukielka: Because it shouldn't handled by JetBrains anymore.
JetBrains is now only responsible for storing the tokens in the secure store.
Other than that all account management is done by Cody and we just receive notifications about it.

src/main/kotlin/com/sourcegraph/cody/listeners/CodyDocumentListener.kt
pkukielka: It's still broken, fix will be a bit more involved, I linked an linear issue

src/main/java/com/sourcegraph/cody/CodyToolWindowFactory.java
pkukielka: Thanks for spotting it!
It is definitely for this PR, in fact I removed too much (update channel selection).
I added it back now.

mkondratek:
image

🚀

mkondratek (approved):

PR #2386, opened 2024-10-04
Get model availability form endpoint instead of computing it manually

pkukielka:
## Changes

Follow-up to https://github.com/sourcegraph/cody/pull/5804

Instead of manually fetching account tier and computing model availability we instead receiving it from agent.
This makes code simpler, ensures the same consitions are used by both Cody and JetBrains, and is less error prone.

## Test plan

**Free user**

1. Login as free user
2. Open Edit Code popup
3. Ensure models are visible
4. Choose one without `Pro` label and do the edit -> it should success
5. Choose one with `Pro` label and do the edit -> it should open the account upgrade website

**Pro use**

1. Login as pro user
2. Open Edit Code popup
3. Ensure models are visible, and there are no `Pro` labels anywhere
4. Choose model which was previously marked as pro on the free tier and do the edit -> it should success

**Enterprise user**

1. Login as enterprise user
2. Open Edit Code popup
3. Ensure models are visible, and there are no `Pro` labels anywhere
4. Choose model which was previously marked as pro (if such exists) on the free tier and do the edit -> it should success

Attention: This is different than the current behaviour, but I checked Cody and it allows for model selection for the enterprise accounts so I think JetBrains behaviour drifted there.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
@@ -14,8 +14,7 @@ import com.sourcegraph.cody.agent.protocol_extensions.isCodyProOnly
 import com.sourcegraph.cody.agent.protocol_extensions.isDeprecated
 import com.sourcegraph.cody.agent.protocol_generated.Chat_ModelsParams
 import com.sourcegraph.cody.agent.protocol_generated.Model
-import com.sourcegraph.cody.config.AccountTier
-import com.sourcegraph.cody.config.CodyAuthenticationManager
+import com.sourcegraph.cody.agent.protocol_generated.ModelAvailabilityStatus
 import com.sourcegraph.cody.edit.EditCommandPrompt
 import com.sourcegraph.cody.history.HistoryService
 import com.sourcegraph.cody.history.state.LLMState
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
```diff
@@ -67,59 +66,52 @@ class LlmDropdown(
...
+              it.id == model ||
+                  it.id == selectedFromHistory?.model ||
+                  it.id == selectedFromChatState?.id
+            } ?: models.firstOrNull()
 
     // If the dropdown is already disabled, don't change it. It can happen
     // in the case of the legacy commands (updateAfterFirstMessage happens before this call).
     isEnabled = isEnabled && chatModelFromState == null
-
-    isVisible = !isEnterpriseAccount || hasServerSentModels
```
pkukielka, line 94: @jamesmcnamara After cody commit bump this seems to be not working anymore.

Could you remind me what was the rationale for `hasServerSentModels` introduction?
I think either way we should decide that on the Cody level, and if some conditions are not meet we should simply not sent models through this endpoint. This way we won't have custom, error-prone logic on the client side.

Also, looks like Cody is not allowing to choose between enterprise models so I'm changing this too.

mkondratek: I think that these could be some leftovers related to chats (which are redundant now as we use the webview UI). You can doublecheck that Edit Code dialog works properly (this is the only place where we use the dropdown presently).

pkukielka: I did: see the testplan :)

mkondratek (approved):

src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
jamesmcnamara: Yes the only need for this flag was to ensure that enterprise users saw a dropdown menu for LLM options on the chat view if their enterprise instance supported it. If we are using the web view now, then it won't matter because the webview handles it itself.

PR #2393, opened 2024-10-05
RD: Distinguish INVOKE trigger kind in the completion provider

mkondratek:
Previously, we were assigning AUTOMATIC trigger kind even for a direct call (manually triggered) autocompletions. This PR sets the proper trigger kind for manual triggers.

## Test plan
1. In RD trigger `shift + option + \`


pkukielka (approved):
LGTM

PR #2394, opened 2024-10-05
Ignore compatibility problems (to fix the issues with the completion provider)

mkondratek:
The exception was there before for some other hacks. We need it back again. 

## Test plan
1. The release script runs successfully


pkukielka (approved):
LGTM

PR #2406, opened 2024-10-07
Rewrite InstructionsInputTextArea to use JBTextArea

pkukielka:
## Changes

Some users reported that text in the edit window does not render properly.
I cannot reproduce it but I suspect it might be due to custom rendering we are doing, I hope switching back to default IJ component might solve the issue (and if not, it's still simpler and less code).

## Test plan

1. Run Edit Code command
2. Ghost test saying `"Type what changes you want to make to this file..."` should be visible
3. Ghost text should disappear when user type anything in the window
4. Ghost text should reappear when text is removed


mkondratek (approved):
nice

PR #2432, opened 2024-10-10
Fix new pannel not opening in IJ 242+

pkukielka:
Fixes https://linear.app/sourcegraph/issue/QA-112/

## Test plan

Please test with both lowest and highest supported IJ versions.

**Opening new panel:**

1. Open IJ  with Cody installed 
2. Click "Open in Editor" button (`[ | ]` icon)

**Moving panels around:**

1. Open IJ with Cody installed 
2. Click "Open in Editor" button (`[ | ]` icon) few times to open few tabs
3. Move them around - change tab orders, move them between split windows, etc

Expected result: new panel should be opened


mkondratek (approved):
🟢 2023.2 
🟢 2024.2.3

PR #2434, opened 2024-10-11
feat(uninstall): delete auth credentials on uninstall (CODY-1043)

jamesmcnamara:
This is the JetBrains half of the [CODY-1043](https://linear.app/sourcegraph/issue/CODY-1043/bug-account-credentials-are-cached-between-installs), the VSC half of it is in [this PR](https://github.com/sourcegraph/cody/pull/5819).

For VSCode I was able to write an E2E as there are electron & VSCode test utilities for opening / closing instances of VSC and installing / uninstalling plugins.

Does anyone know if we have similar utilities that we could use to write a test for this behavior in JB?

The good news is that JB actually has proper APIs for this hook so the complexity is much lower here.

## Test plan
I manually tested this by running `CODY_DIR='<my local cody>' ./gradlew -PforceAgentBuild=true buildPlugin` and then opening the plugins menu and installing a plugin from disk.
image

I logged in and then uninstalled from the same menu. After reinstalling, it takes me back to the login screen and all previous endpoints are cleared. When I log back into the same account I see my chat history is removed.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
@@ -170,4 +170,6 @@ interface _LegacyAgentServer {
 
   @JsonRequest("testing/requestErrors")
   fun testingRequestErrors(): CompletableFuture<List<NetworkRequest>>
+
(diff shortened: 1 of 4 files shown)


pkukielka (approved):
LGTM

PR #2439, opened 2024-10-14
Improve autocomplete UX

mkondratek:
Related to https://github.com/sourcegraph/cody/pull/5871.

In this PR we make use of the recent changes ☝️ we remove some redundant calculations for completion text. 
Additionally, we properly derive the insertion text and the position for the inlay. That makes the inlays look more fancy and precise.

Note that Remote Dev does not use inlays. Some completions won't look that nice there yet (but it is no worse).

### Before  
https://github.com/user-attachments/assets/805c9012-7393-45fd-8d9f-22ae4cd5bf30

### After
https://github.com/user-attachments/assets/d5c2e7b6-4bc9-4ee5-ab8c-4f9c427915aa 

## Test plan

Autocomplete works in all cases

| | remote dev | regular |
| --- | --- | --- |
| single line | Case 1 | Case 2 | 
|  multi line | Case 3 | Case 4 |


src/main/kotlin/com/sourcegraph/cody/autocomplete/CodyAutocompleteManager.kt
```diff
@@ -406,11 +369,30 @@ class CodyAutocompleteManager {
...
+      while (startIndex < completion.length &&
+          startIndex < original.length &&
+          completion[startIndex] == original[startIndex]) {
+        startIndex++
+      }
+
+      // Trim common suffix
+      while (endIndex > 0 &&
+          endIndex > startIndex &&
+          original.length - (completion.length - endIndex) > 0 &&
```
pkukielka, line 389: I think it would be simpler this way:

```
var endIndex = completion.length - 1
while (endIndex > 0 &&
          endIndex > startIndex &&
          completion[endIndex] ==
              original[original.length - (completion.length - endIndex)]) {
        endIndex--
}
```

Moving `-1` out will make sure `endIndex > 0` condition is correct.
`original.length - (completion.length - endIndex) > 0` condition should be redundant if ` endIndex > startIndex`, no?

pkukielka (approved):
LGTM

PR #2442, opened 2024-10-14
Fix repo url parsing

pkukielka:
Fixes https://linear.app/sourcegraph/issue/CODY-3132

## Changes

1. `extensionConfiguration_didChange` is not called anymore on every opening of the file. It was used to set `cody.codebase` settings key on the Cody side. But it should not be needed, as[ it should always be obtained from the currently opened file](https://github.com/sourcegraph/cody/blob/main/vscode/src/repository/git-metadata-for-editor.ts#L53) anyway.

## Test plan

1. Create git test project 
2. Edit `.git/config` file and change `[remote "origin"]` to contain url like:
`url = git@github.com:122345/a/b/c/my-Repo.git`
3. Run IntelliJ and switch files few times - there should be no errors (in fact that code path is not executed anymore)
4. Click right mouse button, `Sourcegraph` > `Open Selection In Sourcegraph Web`
5. It should fail, but URL should start with `https://sourcegraph.com/github.com/1234/sourcegraph/a/b/my-Repo/@main/-/blob/`


mkondratek:
To be precise, the url starts with:

```
https://sourcegraph.com/github.com/122345/a/b/c/my-repo@main/-/blob/
```

(I assume there are just typos in the test plan).

src/main/kotlin/com/sourcegraph/config/ConfigUtil.kt
```diff
@@ -81,7 +80,6 @@ object ConfigUtil {
             UserLevelConfig.getAutocompleteProviderType()?.vscodeSettingString(),
         debug = isCodyDebugEnabled(),
         verboseDebug = isCodyVerboseDebugEnabled(),
-        codebase = CodyAgentCodebase.getInstance(project).getUrl().getNow(null),
```
mkondratek, line 84: this is great 🚀

mkondratek (approved):

PR #2451, opened 2024-10-15
Handle runtime missing JCEF

mkondratek:
Resolves https://linear.app/sourcegraph/issue/CODY-3753/make-it-possible-to-automatically-switch-user-runtime-to-one-which.

## Test plan

Tests:

### Dev run (`:customRunIde`)
1. Modify `build.gradle.kts`. Change
  ```
dependencies {
  intellijPlatform {
    jetbrainsRuntime()
...
  ```
  to
  ```
dependencies {
  intellijPlatform {
    jetbrainsRuntimeExplicit("jbr-21.0.3-osx-aarch64-b446.1")
...
  ```
so that the IDE will use no jcef runtime at the start. 
2. `:customRunIde`
3. Verify the notification appears
4. Verify the proper panel appears


mkondratek:
## Demo 

| Notification | Panel |
| --- | --- |
| image |  image |

src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIService.kt
```diff
@@ -122,6 +130,22 @@ class WebUIService(private val project: Project) {
     }
   }
 
+  private fun createWebUIProxy(delegate: WebUIHost): WebUIProxy? =
+      try {
+        missingJcefExceptionOccurred.set(false)
+        WebUIProxy.create(delegate)
+      } catch (e: IllegalStateException) {
+        if (e.message == JCEF_NOT_SUPPORTED_MESSAGE) {
```
pkukielka: This is potentially fragile.
Maybe instead we can save the exception, and then when refreshing panel if jcef runtime is missing then render current notification, and if it is not missing but webview error happened then show the error? Otherwise we will be stuck with the spinner forever anyway, no?

mkondratek: If it's something else than missing jcef exception we simply rethrow it. This behaviour dees not change. We only add a handling for that particular exception.

danielmarquespt:
Copy is approved

src/main/kotlin/com/sourcegraph/cody/chat/ui/ErrorPanel.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.sourcegraph.common.CodyBundle
+import java.awt.GridBagConstraints
+import java.awt.GridBagLayout
+import javax.swing.JLabel
+import javax.swing.JPanel
+import javax.swing.JTextPane
+import javax.swing.text.SimpleAttributeSet
+import javax.swing.text.StyleConstants
+
+class ErrorPanel : JPanel(GridBagLayout()) {
```
mkondratek, line 15: image

src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIService.kt
```diff
@@ -122,6 +127,16 @@ class WebUIService(private val project: Project) {
     }
   }
 
+  private fun createWebUIProxy(delegate: WebUIHost): WebUIProxy? =
+      try {
+        proxyCreationException.getAndSet(null)
+        WebUIProxy.create(delegate)
+      } catch (e: IllegalStateException) {
+        proxyCreationException.getAndSet(e)
+        CodyToolWindowContent.executeOnInstanceIfNotDisposed(project) { refreshPanelsVisibility() }
```
pkukielka, line 136: Is that really needed?

mkondratek: yes

pkukielka (approved):
LGTM

PR #2464, opened 2024-10-16
Fix error in navigation from context files

mkondratek:
Fixes https://linear.app/sourcegraph/issue/QA-122/jetbrains-crash-occurred-while-clicking-on-file-name-from-chat-context.

## Test plan
1. Follow the repro steps from the QA report - that is, run Explain Code and try the context files navigation links

### DIFF
--- a/src/main/kotlin/com/sourcegraph/utils/CodyEditorUtil.kt
+++ b/src/main/kotlin/com/sourcegraph/utils/CodyEditorUtil.kt
@@ -6,6 +6,7 @@ import com.intellij.injected.editor.EditorWindow
 import com.intellij.lang.Language
 import com.intellij.lang.LanguageUtil
 import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.application.invokeLater
 import com.intellij.openapi.command.WriteCommandAction
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.editor.Document
@@ -31,7 +32,6 @@ import com.intellij.util.withScheme
 import com.sourcegraph.cody.agent.protocol_extensions.toOffsetRange
 import com.sourcegraph.cody.agent.protocol_generated.Range
 import com.sourcegraph.config.ConfigUtil
-import com.sourcegraph.utils.ThreadingUtil.runInEdtAndGet
 import java.net.URI
 import java.net.URISyntaxException
 import kotlin.io.path.createDirectories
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/utils/CodyEditorUtil.kt
```diff
@@ -184,7 +184,7 @@ object CodyEditorUtil {
                 selection.start.line.toInt(),
                 /* logicalColumn= */ selection.start.character.toInt())
           }
-      runInEdtAndGet { descriptor.navigate(/* requestFocus= */ preserveFocus != true) }
+      invokeLater { descriptor.navigate(/* requestFocus= */ preserveFocus != true) }
```
pkukielka: I'm wondering if that won't eat the exception, as it will be already executing on the other thread.
Perhaps we need to refactor this method to return `CompletableFuture`

mkondratek: afaik the exceptions are propagated b/w threads. we are `OpenFileDescriptor` (constructor) requires BGT while navigate requires EDT. the problem was that `runInEdtAndGet` uses different modality than `invokeLater`

mkondratek: I verified the links. I tested Open Cody Settings Editor, Generate Unit Tests, Apply (creating new files) - all worked for me without exceptions.

pkukielka: I do not see how it could work.
`invokeLater` is executing in other thread, while current one can even end in the meantime.
There is no way `invokeLater` could throw exception back in the current thread without blocking.

pkukielka: You can programatically throw exception inside that `invokeLater` block and then call `Generate Unit Tests` and see what the behaviour is?

mkondratek: Having this:
image

I got an internal error:
image

mkondratek: Fixed.

I verified the links. 
I tested 
- Open Cody Settings Editor, 
- Generate Unit Tests, 
- Apply (creating new files)
- Save as file 

all worked for me without exceptions 🟢

pkukielka (approved):
LGTM

PR #2484, opened 2024-10-19
Revert "feat(uninstall): delete auth credentials on uninstall (CODY-1043) (#2434)"

jamesmcnamara:
This reverts commit 65f65c5d9695c098d07c7ff01d4b863fd571f188.

The UninstallListener introduced in this commit seems to run for every extension uninstall instead of just when the sourcegraph extension is uninstalled.

## Test plan
This will need to be manually tested.


dominiccooney (approved):
LGTM

believe the test failures are unrelated flakes

PR #2485, opened 2024-10-19
fix(uninstall): Only clear credentials if the Cody app was uninstalled

jamesmcnamara:
Take two of [CODY-1043](https://linear.app/sourcegraph/issue/CODY-1043/bug-account-credentials-are-cached-between-installs) that was reverted in https://github.com/sourcegraph/jetbrains/pull/2484.

The issue is that uninstall events are triggered when any plugin is uninstalled and not just when the plugin that registered the handler is uninstalled. This checks that the plugin being uninstalled is in fact Cody before running the handler.

## Test plan
I have manually tested that when uninstalling other extensions one branch of the `if` runs, and when uninstalling Cody the other half does.

As discussed in https://github.com/sourcegraph/jetbrains/pull/2434 it would be non-trivial to test this behavior because we will need to drive the installation and uninstallation of the plugin from an automated test driver, which will likely require a much more custom test kit.


dominiccooney (changes requested):
Because the previous iteration of this caused a regression, let's add a test for this.

jamesmcnamara:
@dominiccooney Is this something that would be easy to test with the existing integration tests?

dominiccooney:
> Is this something that would be easy to test with the existing integration tests?

@jamesmcnamara I would think about it this way:

1. A unit test. How hard would it be to dispatch that event to the plugin when the plugin does and doesn't match?
2. An integration test. This will be more valuable. To get some plugins installed I'd just look at how the test framework does it. For driving an uninstall, you could see if there are JetBrains APIs for this or failing that use the UI robot to actually go click on uninstall. You will need a second plugin to be the thing that you uninstall.

src/main/kotlin/com/sourcegraph/cody/initialization/UninstallListener.kt
```diff
@@ -12,13 +12,18 @@ import com.sourcegraph.cody.agent.protocol.BillingProduct
 import com.sourcegraph.cody.agent.protocol.TelemetryEventParameters
 import com.sourcegraph.cody.config.CodyAuthenticationManager
 import com.sourcegraph.cody.telemetry.TelemetryV2
+import com.sourcegraph.config.ConfigUtil
 import java.util.concurrent.TimeUnit
 
 class UninstallListener : StartupActivity {
```
pkukielka, line 18: I noticed it looks like it is currently broken:

image

pkukielka:
> > Is this something that would be easy to test with the existing integration tests?
> 
> @jamesmcnamara I would think about it this way:
> 
> 1. A unit test. How hard would it be to dispatch that event to the plugin when the plugin does and doesn't match?
> 2. An integration test. This will be more valuable. To get some plugins installed I'd just look at how the test framework does it. For driving an uninstall, you could see if there are JetBrains APIs for this or failing that use the UI robot to actually go click on uninstall. You will need a second plugin to be the thing that you uninstall.

For unit test we can use `PluginStateManager.fireState(pluginDescriptor, false)` it is what `PluginInstaller::prepareToUninstall` calls under the hood. Actually it is possible that `PluginInstaller::prepareToUninstall` would work out of the box as wel.

jamesmcnamara:
@dominiccooney @pkukielka I've followed your advice and added a unit test using `PluginInstaller::prepareToUninstall` for both cases. I looked at writing an integration test but couldn't see how to do it easily.

jamesmcnamara:
@dominiccooney @pkukielka Friendly Monday morning bump. I've added a test for the uninstaller.

PR #2497, opened 2024-10-21
improve JB reporting flow

kalanchan:
Moving all feedback to community, and making sure only bugs get reported in the JB repo. Also removing the `bug:` prefix as it no longer makes sense since we are only accepting bugs here

## Test plan
N/A


.github/ISSUE_TEMPLATE/jetbrains_feedback.md
```diff
@@ -0,0 +1,12 @@
+---
+name: '❌ Do not submit feedback or feature requests here'
```
dominiccooney: Could we be friendlier here? Someone is trying to give us the gift of feedback or a feature request... can we be less ❌ and *not* and more positive, motivating "to get your feedback handled by the right team" or something like that.

Where does the "about" field get rendered? Can it render a link? Or could we put (link in description) or something like that?

src/main/kotlin/com/sourcegraph/cody/error/CodyErrorSubmitter.kt
```diff
@@ -67,7 +66,7 @@ class CodyErrorSubmitter : ErrorReportSubmitter() {
 
   private fun getTitle(throwableText: String): String {
     val title = trimPostfix(throwableText.lines().first(), 128)
```
dominiccooney: Don't use two lines where one will do.

```suggestion
    return trimPostfix(throwableText.lines().first(), 128)
```

src/main/kotlin/com/sourcegraph/cody/error/CodyErrorSubmitter.kt
```diff
@@ -67,7 +66,7 @@ class CodyErrorSubmitter : ErrorReportSubmitter() {
...
-    return "bug: $title"
+    return "$title"
```
dominiccooney:
```suggestion
```

(Don't know why GitHub doesn't let me do this in a single suggestion...)

dominiccooney (approved):
Some feedback inline, please consider.

dominiccooney:
Please add a test plan.

.github/ISSUE_TEMPLATE/jetbrains_feedback.md
kalanchan: ok!!! thanks for the feedback, I made it more friendly :) 

I kept the ❌ as it's a **bold** eye catcher, but the wording should sound more friendly now!

pkukielka:
@kalanchan are you still working on this? :)

PR #2500, opened 2024-10-22
Move to the new `extensionConfiguration_change` endpoint

mkondratek:
In this PR:
- [x] moved away from the deprecated agent API

## Test plan
1. N/A - I ve just switched the endpoints


pkukielka (approved):
LGTM, although now we also receive `ProtocolAuthStatus`, and I'm wondering if we don't want to handle it somehow...

mkondratek:
> LGTM, although now we also receive `ProtocolAuthStatus`, and I'm wondering if we don't want to handle it somehow...

We can and maybe we want. I cannot think of any example... 🤔 maybe switching accounts, logging in... idk

PR #2531, opened 2024-10-28
Add window/didChangeFocus notification calls 

mkondratek:
Related PR: https://github.com/sourcegraph/cody/pull/5997

## Test plan
I reviewed the `cody-agent-trace.json`. The requests appear correctly in proper time. Also, I checked the agent in a debugger session - the requests come and have the proper param.


CONTRIBUTING.md
mkondratek, line 1: unrel fixes - GitHub does not handle \n correctly

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentServer.kt
```diff
@@ -79,8 +87,15 @@ interface _SubsetGeneratedCodyAgentServer {
...
 
+  @JsonNotification("initialized") fun initialized(params: Null?)
+
+  @JsonNotification("exit") fun exit(params: Null?)
+
   @JsonNotification("extensionConfiguration/didChange")
   fun extensionConfiguration_didChange(params: ExtensionConfiguration)
+
+  @JsonNotification("window/didChangeFocus")
+  fun window_didChangeFocus(params: Window_DidChangeFocusParams)
```
mkondratek, line 98: this is the new notification - the others have been reorganized only

src/main/kotlin/com/sourcegraph/cody/listeners/CodyFocusChangeListener.kt
```diff
@@ -15,8 +16,15 @@ class CodyFocusChangeListener(val project: Project) : FocusChangeListener {
       EditorChangesBus.documentChanged(project, textDocument)
       CodyAgentService.withAgent(project) { agent: CodyAgent ->
         agent.server.textDocumentDidFocus(textDocument)
+        agent.server.window_didChangeFocus(Window_DidChangeFocusParams(true))
```
mkondratek: Here, we perform two calls:
- `textDocument/didFocus`
- `window/didChangeFocus`

I wonder... do we need the two? maybe we can optimize it 🤔

mkondratek: I think `textDocument/didFocus` notification implies `window/didChangeFocus` (true).

src/main/kotlin/com/sourcegraph/cody/listeners/CodyFocusChangeListener.kt
```diff
@@ -15,8 +16,15 @@ class CodyFocusChangeListener(val project: Project) : FocusChangeListener {
...
       }
       IgnoreOracle.getInstance(project).focusedFileDidChange(textDocument.uri)
     }
   }
+
+  override fun focusLost(editor: Editor) {
```
mkondratek: Switching b/w files in the project calls `focusLost` and `focusGained` immediately afterwards.

Is it ok?

src/main/kotlin/com/sourcegraph/cody/listeners/CodyFocusChangeListener.kt
pkukielka: That is true for focus gain, but not for the focus lost.
But if we need notification about focus lost or not - I cannot say :)

src/main/kotlin/com/sourcegraph/cody/listeners/CodyFocusChangeListener.kt
pkukielka: I think that is incorrect implementation:
1. That focus `flickering` is incorrect on its own
2. What if user will close the only one opened document?

What you need is new listener:

```
    val frame = WindowManager.getInstance().getFrame(project)
    val listener =
        object : WindowAdapter() {
          override fun windowActivated(e: WindowEvent?) {
            super.windowActivated(e)
            ApplicationManager.getApplication().executeOnPooledThread { getAuthenticationState() }
          }
        }
    frame?.addWindowListener(listener)
    Disposer.register(this) { frame?.removeWindowListener(listener) }
```

This code is copied, please take a look at `CodyAuthenticationManager::addAuthChangeListener`. You would also need to override `windowDeactivated`

mkondratek: done

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -114,6 +116,16 @@ class CodyAuthenticationManager :
           override fun windowActivated(e: WindowEvent?) {
             super.windowActivated(e)
             ApplicationManager.getApplication().executeOnPooledThread { getAuthenticationState() }
+            CodyAgentService.withAgent(project) { agent: CodyAgent ->
```
pkukielka: Can we extract this listener somewhere an then subscribe to it (e.g. have option to add a callback) if we want in places we need? Or even we can hardcode call to `CodyAccountManager` inside it if you want to make it simpler, structure will just have more sense other way around.
Doing general state notifications inside `CodyAccountManager` looks weird...

mkondratek: I extracted the class

src/main/kotlin/com/sourcegraph/cody/config/CodyWindowManager.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.project.Project
+import com.sourcegraph.cody.agent.CodyAgent
+import com.sourcegraph.cody.agent.CodyAgentService
+import com.sourcegraph.cody.agent.protocol_generated.Window_DidChangeFocusParams
+import java.awt.event.WindowAdapter
+import java.awt.event.WindowEvent
+
+class CodyWindowAdapter(private val project: Project) : WindowAdapter() {
```
pkukielka: Can you also rename file to match class name?

mkondratek: done

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/config/CodyAuthenticationManager.kt
```diff
@@ -109,21 +107,15 @@ class CodyAuthenticationManager :
 
   fun addAuthChangeListener(project: Project) {
```
pkukielka: This method should be now part of `CodyWindowAdapter`, it does not make sense in `CodyAuthenticationManager`

PR #2541, opened 2024-10-29
Make sure  context change is handled properly

pkukielka:
Fixes https://linear.app/sourcegraph/issue/BUGS-621/dumped-back-to-login-screen-in-v7018

## Test plan

I do not have reliable way to reproduce it manually.

Programatically I was testing it by forcing 'sign off' on the cody side (e.g. using `authProvider.setAuthPendingToEndpoint("sourcegraph.com")`) and observing behaviour in JetBrains.

Without my changes it was sometimes causing wrong login panel, or chat panel was not functioning properly.
After my changes it redirects user to the native sign-in panel.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
@@ -10,6 +10,7 @@ import com.intellij.openapi.project.guessProjectDir
 import com.intellij.openapi.vfs.VfsUtil
 import com.intellij.openapi.vfs.VirtualFile
 import com.jetbrains.rd.util.firstOrNull
+import com.sourcegraph.cody.CodyToolWindowContent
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/cody/agent/CodyAgentClient.kt
```diff
@@ -228,4 +231,12 @@ class CodyAgentClient(private val project: Project, private val webview: NativeW
 
     return saveFileFuture
   }
+
+  @JsonNotification("window/didChangeContext")
+  fun window_didChangeContext(params: Window_DidChangeContextParams) {
+    if (params.key == "cody.activated" && params.value?.toBoolean() == false) {
+      CodyAuthenticationManager.getInstance().setActiveAccount(null)
```
pkukielka: This can happen e.g. if cody will decide that token provided by JetBrains is expired.

mkondratek (approved):

PR #2577, opened 2024-11-04
Remove potential infinite init loop

mkondratek:
Accidentally, I found an infinite loop in cody tool window content initialization. It was likely a reason of the recent CI failures. This PR eliminates the risk of the inifinite loop.

## Test plan
1. Run integrationTests locally - should be all green, no sus logs, or errros indicating a loop in call stack


src/main/java/com/sourcegraph/cody/CodyToolWindowFactory.java
```diff
@@ -39,6 +40,10 @@ public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindo
           }
 
           toolWindow.setAdditionalGearActions(customCodySettings);
+
```
pkukielka, line 43: Nice find!

pkukielka (approved):
LGTM!

PR #2584, opened 2024-11-05
Migrate autocomplete related protocol to the generated protocol

mkondratek:
- [x] Update cody commit

## Test plan
1. Autocomplete works no worse


pkukielka (approved):
LGTM

PR #2586, opened 2024-11-06
Fix plugin name in the marketplace

pkukielka:
## Test plan

1. Build and install the plugin
2. Check plugin name in `Settings > Plugins`

image

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 # IntelliJ Platform Artifacts Repositories
 # -> https://plugins.jetbrains.com/docs/intellij/intellij-artifacts.html
 pluginGroup=com.sourcegraph.jetbrains
-pluginName=Sourcegraph
+pluginName=Cody: AI Coding Assistant with Autocomplete & Chat
 # SemVer format -> https://semver.org
 pluginVersion=6.0-localbuild
 # See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html


gradle.properties
```diff
@@ -1,7 +1,7 @@
 # IntelliJ Platform Artifacts Repositories
 # -> https://plugins.jetbrains.com/docs/intellij/intellij-artifacts.html
 pluginGroup=com.sourcegraph.jetbrains
-pluginName=Sourcegraph
+pluginName=Cody: AI Coding Assistant with Autocomplete & Chat
```
mkondratek, line 4: I was thinking about it too 😅 

There is a logical inconsistency with this change. The plugin is not only about Cody but Code Search too.

Our description looks like this (it is defined in [the marketplace settings](https://plugins.jetbrains.com/plugin/9682-sourcegraph/edit)):
image

I think this change should be approved by someone else (marketing team?).

pkukielka: I copied plugin name from `plugin.xml` where it was approved, this is most likely regression from IJ plugin 2.0 migration.

And name was changed intentionally some time ago:
https://github.com/sourcegraph/jetbrains/commit/9761f2597af73fa2db530f0afc3c7258538c9dd2

mkondratek: ok then

mkondratek (approved):

gradle.properties
kalanchan: we'll make a decision to deprecate the code search part, its legacy

PR #2588, opened 2024-11-06
Add options for ConfigFactory to handle customConfig properly

mkondratek:
Fixes https://linear.app/sourcegraph/issue/CODY-4264/a-comment-in-cody-settingsjson-breaks-cody-initialization-stuck-at.

## Test plan
1. Have the cody_settings.json like this:
```
{
  // other providers...
  "openctx.providers": {
    "https://gist.githubusercontent.com/thenamankumar/3708f084fb2abd57adafe7f14620bdf7/raw/e7c7059cb5b04f6feead002e7b3a90c5b1a4bb96/provider.js": true
  }
}
```
2. Cody should initialize

### DIFF
--- a/src/main/kotlin/com/sourcegraph/config/ConfigUtil.kt
+++ b/src/main/kotlin/com/sourcegraph/config/ConfigUtil.kt
@@ -1,7 +1,6 @@
 package com.sourcegraph.config
 
 import com.google.gson.JsonObject
-import com.google.gson.JsonParser
 import com.intellij.ide.plugins.PluginManagerCore
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.editor.Editor
@@ -17,6 +16,9 @@ import com.sourcegraph.cody.config.CodyAuthenticationManager
 import com.sourcegraph.cody.config.ServerAuthLoader
 import com.sourcegraph.cody.config.SourcegraphServerPath
 import com.sourcegraph.cody.config.SourcegraphServerPath.Companion.from
+import com.typesafe.config.ConfigFactory
+import com.typesafe.config.ConfigRenderOptions
+import com.typesafe.config.ConfigValueFactory
 import java.nio.file.Path
 import java.nio.file.Paths
 import kotlin.io.path.readText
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/sourcegraph/config/ConfigUtil.kt
```diff
@@ -141,11 +139,11 @@ object ConfigUtil {
 
     return try {
       val text = customConfigContent ?: getSettingsFile(project).readText()
-      val config = ConfigFactory.parseString(text).resolve()
+      val config = JsonParser.parseString(text).asJsonObject
```
mkondratek: ConfigFactory does not render json. The rendered config contained the comment with # instead of //. Using JsonParser fixes the problem.

pkukielka: Other option would be adding `.setComments(false)` in the current code.
I thought Lightened Config was a bit more forgiving when parsing e.g. additional commas.
Maybe we can write simple test for that and check how those corner cases looks in case of both approaches?

mkondratek: fixed

src/main/kotlin/com/sourcegraph/config/ConfigUtil.kt
```diff
@@ -141,11 +141,13 @@ object ConfigUtil {
 
     return try {
       val text = customConfigContent ?: getSettingsFile(project).readText()
-      val config = ConfigFactory.parseString(text).resolve()
+      var config = ConfigFactory.parseString(text).resolve()
       additionalProperties.forEach { (key, value) ->
-        config.withValue(key, ConfigValueFactory.fromAnyRef(value))
```
mkondratek, line 146: There was a bug 😄 configs are immutable

src/test/kotlin/com/sourcegraph/config/ConfigUtilTest.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    assertEquals(
+        "indentation-based",
+        parsed
+            .get("cody")
+            .asJsonObject
+            .get("experimental")
+            .asJsonObject
+            .get("foldingRanges")
+            .asString)
```
mkondratek, line 33: `withValue` adds the field as 
```
"cody": {
  "experimental": {
    "foldingRanges": "indentation-based"
  }
}
```
instead of 
```
"cody.experimental.foldingRanges": "indentation-based"
```

Is it a problem? 🤔

src/main/kotlin/com/sourcegraph/config/ConfigUtil.kt
mkondratek: Basically, we ignored "foldingRanges": "indentation-based" - how important for us is this option?

src/test/kotlin/com/sourcegraph/config/ConfigUtilTest.kt
pkukielka: IMHO It is not, those are equivalent in VSC config.

pkukielka (approved):
LGTM

PR #2630, opened 2024-11-12
Update dependencies

mkondratek:
This PR updates some deps and removes a few redundant ones.

## Test plan
1. n/a

PR #2632, opened 2024-11-13
Migrate featureFlags/getFeatureFlag and graphql/getCurrentUserCodySubscription

mkondratek:
- [Migrate featureFlags/getFeatureFlag](https://github.com/sourcegraph/jetbrains/pull/2632/commits/ee1d496785c7e76fcc5079afeed0730599ff9129)
- [Migrate graphql/getCurrentUserCodySubscription](https://github.com/sourcegraph/jetbrains/pull/2632/commits/3e5a8888e49418dd7c950c7d2d8fcf049f7b76d3)

# Based on PR #2630

## Full chain of PRs as of 2024-11-13

* PR #2632:
  `mkondratek/chore/migrate-api-part-1` ➔ `mkondratek/chore/update-deps`
* PR #2630:
  `mkondratek/chore/update-deps` ➔ `main`

## Test plan
- Verified with debugger - the server returns proper values.
- Verified with debugger - with the mocked results the notification about the ended trial appears


src/main/kotlin/com/sourcegraph/cody/agent/protocol/GetCurrentUserCodySubscription.kt
```diff
@@ -0,0 +1,18 @@
+package com.sourcegraph.cody.agent.protocol
+
+object GetCurrentUserCodySubscription {
```
mkondratek: We can consider renaming this class.

We should consider options to automatically generate the values of the enums. 

However, I think we can simply remove what is redundant now. It will be no worse. Then, having less code it will be easier to improve the generation.

pkukielka: I think we should put it in `protocol_extensions` package.

pkukielka: Also, if we are only using two values from each enum we could consider using extension methods instead:

```kotlin
fun CurrentUserCodySubscription.isProPlan() {
   return plan == "PRO"
}

fun CurrentUserCodySubscription.isPendingStatus() {
   return status == "PENDING"
}
```

and then on the call site e.g.:

`currentUserCodySubscription.isProPlan()` instead of 
`currentUserCodySubscription.status == GetCurrentUserCodySubscription.Status.PENDING`

Then we won't need custom objects like `GetCurrentUserCodySubscription`. But it is also fine as it is.

mkondratek: makes sense ! thanks

mkondratek: fixed!

pkukielka (approved):
LGTM

PR #2633, opened 2024-11-13
Migrate textDocument/*

mkondratek:
-  [Migrate textDocument/*](https://github.com/sourcegraph/jetbrains/pull/2633/commits/c0a20823f2b7f93d02c3c079e84282140a8cb371) 

# Based on PR #2632

## Full chain of PRs as of 2024-11-13

* PR #2633:
  `mkondratek/chore/migrate-api-part-2` ➔ `mkondratek/chore/migrate-api-part-1`
* PR #2632:
  `mkondratek/chore/migrate-api-part-1` ➔ `mkondratek/chore/update-deps`
* PR #2630:
  `mkondratek/chore/update-deps` ➔ `main`

## Test plan
- Verified with debugger - the proper values are sent
- Tested with current selection in chat, tested with autocompletion, tested with inline edits - works


src/main/kotlin/com/sourcegraph/cody/agent/protocol/ProtocolTextDocumentUtil.kt
```diff
@@ -13,23 +13,17 @@ import com.intellij.util.concurrency.annotations.RequiresEdt
...
-class ProtocolTextDocument
-private constructor(
-    val uri: String,
-    val content: String? = null,
-    val selection: Range? = null,
-    val visibleRange: Range? = null,
-    val contentChanges: List<ProtocolTextDocumentContentChangeEvent>? = null,
-    val testing: TestingParams? = null,
-) {
+class ProtocolTextDocumentUtil {
```
mkondratek: Again, we may want to reconsider the naming. But I'd rather migrate the API quickly and apply the other improvements later.

src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt
```diff
@@ -115,7 +115,7 @@ private constructor(
                       version = ConfigUtil.getPluginVersion(),
                       ideVersion = ApplicationInfo.getInstance().build.toString(),
                       workspaceRootUri =
-                          ProtocolTextDocument.normalizeUriOrPath(
+                          ProtocolTextDocumentUtil.normalizeUriOrPath(
```
pkukielka: Why not put it in the `protocol_extensions` and use the existing extension pattern?
Then you won't need all those renames.

mkondratek: These methods are not extension functions in Kotlin sense. However, I did some refactor and it should look better now.

mkondratek: In particular. I cannot define the other class named `ProtocolTextDocument` since there is one generated already.

src/main/kotlin/com/sourcegraph/cody/agent/protocol_extensions/TestingParamsUtil.kt
```diff
@@ -0,0 +1,6 @@
+package com.sourcegraph.cody.agent.protocol_extensions
+
+object TestingParamsUtil {
```
pkukielka: Could also be `TestingParamsExt` for consistency

pkukielka (approved):
LGTM

src/main/kotlin/com/sourcegraph/cody/agent/protocol_extensions/TestingParamsUtil.kt
mkondratek: 👍

PR #2645, opened 2024-11-14
Restart Cody in case of webview renderer crash

pkukielka:
## Changes

Looks like webview renderer process crashes from time to time for some users - especially in cases like long running IDE or suspending/unsuspending OS.

We might not be able to fully fix the root cause, but we should handle the crash and recover from it.
For now, I'm just restarting the agent which also restarts the webview.
I was trying to only restart the webviews and keep the agent running, but I did not managed to do it on the first try.
Let's merge this as it improves current situation significantly, and I will work on better/quicker restoration of the state in the followup PR(s).

## Test plan

1. Start IntelliJ with Cody panel opened
2. Open process explorer and kill `jcef Helper (Renderer)`
3. Go back to IntelliJ - Cody panel should restart itself

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIProxy.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIProxy.kt
@@ -5,13 +5,16 @@ import com.google.gson.JsonParser
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.ide.CopyPasteManager
+import com.intellij.openapi.project.ProjectManager
 import com.intellij.ui.jcef.JBCefBrowserBase
 import com.intellij.ui.jcef.JBCefBrowserBuilder
 import com.intellij.ui.jcef.JBCefJSQuery
 import com.sourcegraph.cody.agent.CodyAgent
+import com.sourcegraph.cody.agent.CodyAgentService
 import com.sourcegraph.cody.agent.protocol.WebviewOptions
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIProxy.kt
```diff
@@ -393,17 +398,27 @@ private class ExtensionRequestHandler(
...
   ): Boolean {
-    // TODO: Add Telemetry here.
+    proxy.openDevTools()
+    logger.warn(
+        """Certificate error occurred while loading URL: $request_url
+        Error code: ${cert_error?.name}
+        SSL Info: $sslInfo"""
+            .trimIndent())
+    ProjectManager.getInstance().openProjects.forEach { project ->
+      TelemetryV2.sendTelemetryEvent(project, "cody.webview.request", "certError")
```
mkondratek, line 408: 🚀

src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIProxy.kt
```diff
@@ -393,17 +398,27 @@ private class ExtensionRequestHandler(
...
     return false
   }
 
   override fun onRenderProcessTerminated(
       browser: CefBrowser?,
       status: CefRequestHandler.TerminationStatus?
   ) {
-    // TODO: Add Telemetry here.
-    // TODO: Logging.
-    // TODO: Trigger a reload.
```
mkondratek, line 406: 😄

mkondratek (approved):

PR #2650, opened 2024-11-14
feat(rel): add backport GHA to Jetbrains

Chickensoupwithrice:
Having tested backport functionality in the [Cody repo](https://github.com/sourcegraph/cody/pull/6119) I'm now adding it here too.

This PR adds a backport github action based our backport tooling in sourcegraph/backport
It uses the almost the same label as the Cody one, but with a slightly different base branch (backport jetbrains-v<version-number>). Notably, the `jetbrains-v<version-number>` branch doesn't appear to be the standard here, but I think we should adopt the same convention that is used in Cody.

## Test plan
CI

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,27 @@
+name: Backport
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-latest
+    # Only react to merged PRs for security reasons.
+    # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
+    if: >
+      github.event.pull_request.merged
+      && (
+        github.event.action == 'closed'
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'backport')
+        )
+      )
+    steps:
+      - uses: sourcegraph/backport@v2
+        with:
+          github_token: ${{ secrets.BACKPORT_GITHUB_TOKEN }}
+          label_pattern: '^backport (?<base>jetbrains-v\d+\.\d+\.x)$'
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -24,4 +24,4 @@ jobs:
       - uses: sourcegraph/backport@v2
         with:
           github_token: ${{ secrets.BACKPORT_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -0,0 +1,27 @@
...
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'backport')
+        )
+      )
+    steps:
+      - uses: sourcegraph/backport@v2
+        with:
+          github_token: ${{ secrets.BACKPORT_GITHUB_TOKEN }}
+          label_pattern: '^cherrypick (?<base>jetbrains-v\d+\.\d+\.x)$'
```
jdpleiness: Is this the label convention we want?

Chickensoupwithrice: Oh shit, it should be `backport jetbrains-v<version>`

Fixed

Chickensoupwithrice: I spoke to @kalanchan 2 days ago and he said at the time, that for jetbrains the release branch should be `jetbrains-v<version>`, like how in `sourcegraph/cody` it's `vscode-v<version>`

.github/workflows/backport.yml
```diff
@@ -0,0 +1,27 @@
...
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'backport')
+        )
+      )
+    steps:
+      - uses: sourcegraph/backport@v2
+        with:
+          github_token: ${{ secrets.BACKPORT_GITHUB_TOKEN }}
+          label_pattern: '^backport (?<base>jetbrains-v\d+\.\d+\.x)$'
```
kalanchan:
```suggestion
          label_pattern: '^backport (?<base>jb-v\d+\.\d+\.x)$'
```

WOOPs my bad, the branch should be `jb-vX.YZ.x` so let's match this with the label. I need to align with Dominic on what we think the label should look like, can we freely change the pattern once this merged or will it require backport library updates as well?

Chickensoupwithrice: You can freely change the label_pattern here without needing any changes to the backport library :)

kalanchan: woooo ok ty!

jdpleiness (approved):

.github/workflows/backport.yml
Chickensoupwithrice: The only thing is that it needs is the named capture group for `base`

DaedalusG (approved):

PR #2657, opened 2024-11-16
fix: don't add release team as reviewer for backports

Chickensoupwithrice:
Clone of the Cody PR #6136, so that we don't have failing backports

## Test plan
Tested on the [Cody repo](https://github.com/sourcegraph/cody/pull/6136)


kalanchan (approved):
ty!

PR #2661, opened 2024-11-18
Migrate telemetry/recordEvent

mkondratek:
- [Migrate telemetry/recordEvent](https://github.com/sourcegraph/jetbrains/pull/2661/commits/1e91c0b1acb65183d826b0ff60dab3eba5cb6662)

## Test plan
1. All migrated endpoints have been verified with debugger (and reviewed in the trace log).


gradle.properties
```diff
@@ -25,4 +25,4 @@ org.gradle.jvmargs=-Xmx4g -Xms500m
 nodeBinaries.commit=8755ae4c05fd476cd23f2972049111ba436c86d4
 nodeBinaries.version=v20.12.2
 cody.autocomplete.enableFormatting=true
-cody.commit=6ac4a8c1831ad3945fc16f20b8f21947897d3d14
```
mkondratek: - [x] wait for https://github.com/sourcegraph/cody/pull/6144 and and for the release including the change
- [x] update cody commit

src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
```diff
@@ -2,7 +2,5 @@
 
 public class CodeEditorFactory {
 
-  public static final int spaceBetweenButtons = 5;
-
   public static volatile String lastCopiedText = null;
```
mkondratek, line 5: It looks like since we have moved to webview we do not set this value. Hence, we do not log the paste event.

Or at any chance, does the agent handle the paste event? If it does not, we probably need to send the copied text via a notification to the client to track it in here.

pkukielka (approved):
LGTM

src/main/java/com/sourcegraph/cody/chat/CodeEditorFactory.java
pkukielka: https://linear.app/sourcegraph/issue/CODY-4121/fix-keydownpaste-telemetry-events-logging

PR #2662, opened 2024-11-18
Fix incorrect cross-project event propagation

pkukielka:
## Changes

Due to incorrect implementation of event listeners, if more than one project was open, events like caret change were propagated between editors. This PR fixes it.

## Test plan

Test separately with four different options ([opened file | active file | caret position |  selected text]):

1. Open two projects in IntelliJ.
2. Change [opened file | active file | caret position |  selected text] for the first project
3. Change [opened file | active file | caret position |  selected text] for the second project
4. Files set as context in the Cody chat should be different and match files activated on the previous steps.

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/initialization/PostStartupActivity.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/initialization/PostStartupActivity.kt
@@ -47,8 +47,11 @@ class PostStartupActivity : ProjectActivity {
 
     CodyStatusService.resetApplication(project)
 
-    val multicaster = EditorFactory.getInstance().eventMulticaster as EditorEventMulticasterEx
     val disposable = CodyAgentService.getInstance(project)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/sourcegraph/cody/initialization/PostStartupActivity.kt
```diff
@@ -47,8 +47,11 @@ class PostStartupActivity : ProjectActivity {
 
     CodyStatusService.resetApplication(project)
 
-    val multicaster = EditorFactory.getInstance().eventMulticaster as EditorEventMulticasterEx
     val disposable = CodyAgentService.getInstance(project)
+
+    // WARNING: All listeners should check if an event they are receiving is matching project
```
mkondratek, line 52: 😕

mkondratek (approved):

PR #2674, opened 2024-11-21
bump commit to vscode-v1.46.x

kalanchan:
Uses the same commit as `jb-v7.3.x`, targets the [vscode-v1.46.x](https://github.com/sourcegraph/cody/commits/vscode-v1.46.x/) release branch

I probably did this in the wrong order, I should've bumped the commit on main and then cut the release branch 

## Test plan
N/A


pkukielka (approved):
LGTM

PR #2699, opened 2024-11-27
Fix crash on Edit Code 

mkondratek:
Closes https://linear.app/sourcegraph/issue/QA-180/jetbrains-crash-occurred-while-clicking-on-edit-code-after-restarting.

## Test plan
See the linear issue.
You can try removing `AccountData` entry from `workspace.xml` before opening the project.

## Changes

First I tried to simply setActiveAccount for the deprecated manager. This is not the best solution as do not know some values for the account (`name`, `displayName`, `id`). We could hardcode some values but we would be persisting some artificial data for a simple task that is - **all we need is to store the llm that was used for Edit Prompt most recently**.

So instead of hacking & mocking the deprecated logic (HistoryService is used basically only for history reasons, migrations, etc)... instead of that, let's simply use a simply dedicated storage class. 

Also, it could be an actual storage with a corresponding storage file (.xml) but do we really need it? Project session persistence should the enough for most users.


src/main/kotlin/com/sourcegraph/cody/chat/ui/LlmDropdown.kt
```diff
@@ -73,19 +69,14 @@ class LlmDropdown(
...
-              it.model.id == selectedFromChatState?.id
-        } ?: models.firstOrNull()
-
-    // If the dropdown is already disabled, don't change it. It can happen
-    // in the case of the legacy commands (updateAfterFirstMessage happens before this call).
-    isEnabled = isEnabled && chatModelFromState == null
+        availableModels.find { it.model.id == model || it.model.id == defaultLlm?.id }
+            ?: models.firstOrNull()
+
+    isEnabled = true
```
pkukielka: Previously we were disabled model selection after edit and not re-enabling it on retry.
Is that intentional change?

pkukielka (approved):
LGTM

PR #2817, opened 2024-12-05
Store active account settings per project

pkukielka:
Fixes https://linear.app/sourcegraph/issue/QA-188/jetbrains-adding-password-in-keychain-will-logout-user-when-same

## Changes

There is a lot of cosmetic changes due to slightly different interfaces but main changes are there:
https://github.com/sourcegraph/jetbrains/pull/2817/files#diff-56b9b0d56648edfc3a3154ee5357aeaa891fb834b049e5eb0fb1da4a2eff5c2f

This PR changes the way account settings are stored.
Instead of having one settings shared by all cody clients we keep them separated per project.
This simplifies synchronisation between instance (as it is not needed) and is more in-line with model where each agent is separate being. I believe it is also more intuitive to the users.

## Test plan

TBD

### DIFF
--- a/src/main/kotlin/com/sourcegraph/cody/auth/CodyAccountService.kt
+++ b/src/main/kotlin/com/sourcegraph/cody/auth/CodyAccountService.kt
@@ -1,0 +1,56 @@
+package com.sourcegraph.cody.auth
+
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.components.service
+import com.intellij.openapi.project.Project
+import com.sourcegraph.cody.auth.CodySecureStore.getFromSecureStore
+import com.sourcegraph.cody.auth.CodySecureStore.writeToSecureStore
+
+@Service(Service.Level.PROJECT)
+class CodyAccountService(val project: Project) {
+  private val activeAccountMarker = "active_cody_account"
+  private val projectActiveAccountMarker = project.locationHash
+  private var activeAccount: CodyAccount? = null
+
+  @Volatile private var isActivated: Boolean = false
+
+  fun isActivated(): Boolean {
+    return isActivated
+  }
+
+  fun setActivated(isActivated: Boolean) {
+    this.isActivated = isActivated
+  }
+
+  fun loadAccount(serverUrl: String): CodyAccount {
+    val token = serverUrl.let { getFromSecureStore(it) }
+    return CodyAccount(SourcegraphServerPath(serverUrl), token)
+  }
+
+  fun storeAccount(account: CodyAccount) {
+    writeToSecureStore(account.server.url, account.token)
+  }
+
+  fun getActiveAccount(): CodyAccount? {
+    if (activeAccount == null) {
+      val serverUrl =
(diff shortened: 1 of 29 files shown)


src/main/kotlin/com/sourcegraph/cody/auth/CodyAccountService.kt
```diff
@@ -0,0 +1,56 @@
...
+  fun loadAccount(serverUrl: String): CodyAccount {
+    val token = serverUrl.let { getFromSecureStore(it) }
+    return CodyAccount(SourcegraphServerPath(serverUrl), token)
+  }
+
+  fun storeAccount(account: CodyAccount) {
+    writeToSecureStore(account.server.url, account.token)
+  }
+
+  fun getActiveAccount(): CodyAccount? {
```
dominiccooney, line 34: When we do JetBrains-specific state that's not coordinated with TypeScript/agent, we usually end up regretting it.

What happens if the project active account marker points to a signed out account?

src/main/kotlin/com/sourcegraph/cody/config/notification/CodySettingChangeListener.kt
```diff
@@ -20,7 +20,7 @@ class CodySettingChangeListener(project: Project) : ChangeListener(project) {
         object : CodySettingChangeActionNotifier {
           override fun afterAction(context: CodySettingChangeContext) {
             // Notify JCEF about the config changes
-            javaToJSBridge?.callJS("pluginSettingsChanged", ConfigUtil.getConfigAsJson())
+            javaToJSBridge?.callJS("pluginSettingsChanged", ConfigUtil.getConfigAsJson(project))
```
dominiccooney, line 23: Nothing listens to this, so can we simply remove it?

dominiccooney (changes requested):
In general I like the direction of making the settings per-project.
- Let's remove the unused webview notification.
- Please clarify the PR description; it doesn't specify whether you're talking about runtime or on-disk, it makes it sound like you'd have to sign in once for each project. Please clarify the runtime aspect of it.
- Add a manual test plan for removing an account that a project has cached as its preferred account. Does the behavior make sense, when you reopen the project?

dominiccooney:
I assume this was started a while ago, which is why the PR is here. But let's start new work in sourcegraph/cody.

src/main/kotlin/com/sourcegraph/cody/config/notification/CodySettingChangeListener.kt
pkukielka: For code listening on that you would need to look at `sourcegraph/sourcegraph`.
But perhaps it is an overkill, I think popup we are using for code search is recreated every time. I will check if there is no performance hit in refactoring it a bit to not require that.

dominiccooney: As long as it is used, that's fine. Thanks for the pointer.

